PR #52, opened 2020-11-17
DKG

kape1395:
Just for review, for now.


Makefile
```diff
@@ -0,0 +1,16 @@
+all: build
```
dessaya, line 1: I think @lunfardo314 and @BugFreeSoftware are using `.bat` files for commonly used commands. Maybe it would be a good idea to use one single tool for that. Since `.bat` only works in Windows I'm inclined to vote for Makefile :)

packages/dkg/proc.go
```diff
@@ -0,0 +1,622 @@
...
+	chainID                  coretypes.ChainID // Same as dkgID, just parsed.
+	dkShare                  *DKShare          // This will be generated as a result of this procedure.
+	step                     string            // The current step.
+	node                     *node             // DKG node we are running in.
+	nodeIndex                int               // Index of this node.
+	nodeLoc                  string
+	peerLocs                 []string
+	peerPubs                 []kyber.Point
+	coordPub                 kyber.Point
+	treshold                 uint32
```
dessaya, line 36: typo: `threshold`

packages/dkg/proc.go
```diff
@@ -0,0 +1,622 @@
...
+func (p *proc) processLoop(timeout time.Duration) {
+	timeoutCh := time.After(timeout)
+	done := false
+	acceptPeerMsgType := byte(0)
+	acceptPeerMsgCh := make(chan peerMsgCh, len(p.peerPubs))
+	for !done {
+		select {
+		case coordStepCall := <-p.coordStepCh:
+			switch coordStepCall.msg.Step {
+			case "1-R2.1-SendDeals":
```
dessaya, line 193: Could these strings be constants?

packages/dkg/proc.go
```diff
@@ -0,0 +1,622 @@
...
+	"go.dedis.ch/kyber/v3/pairing"
+	rabin_dkg "go.dedis.ch/kyber/v3/share/dkg/rabin"
+	"go.dedis.ch/kyber/v3/sign/bdn"
+	"go.dedis.ch/kyber/v3/sign/schnorr"
+)
+
+//
+// Stands for a DKG procedure instance on a particular node.
+//
+type proc struct {
```
dessaya, line 25: This structure is too large; maybe it could be split into smaller components? And/or remove some non-orthogonal members (like `chainID` which can be calculated from `dkgID`)?

packages/dkg/coordinator.go
```diff
@@ -0,0 +1,153 @@
...
+// procedure on a set of nodes. The client is not required to have an instance
+// of the DkgNode, but may have it (be one of the peers sharing the secret).
+// This function works synchronously, so the user should run it async if needed.
+func GenerateDistributedKey(
+	coordKey kyber.Scalar,
+	coordPub kyber.Point,
+	peerLocs []string,
+	peerPubs []kyber.Point,
+	treshold uint32,
+	version byte, // address.VersionED25519 = 1 | address.VersionBLS = 2
```
dessaya, line 29: Should this be of type `address.Version` instead of `byte`?

Although now I see that `address.VersionED25519` and `address.VersionBLS` are also `byte` and should be `Version`.

My concern is also that I saw this comment repeated a few other times, and I think it may be beneficial to find a way to express the enum explicitly.

dessaya (commented):
@kape1395 This is a really long PR... superficially I think it looks good; if you need a more in-depth review maybe we should do a meeting where you walk us through the code?
Anyway, I commented on a few aesthetic details, I found no serious issues so far (ignoring the commented out code, which I assume will be fixed before merging).

What is `docs/uml/wasp.mdzip`? I'm guessing it's a UML diagram; maybe add a README describing how to open it?

Makefile
lunfardo314: bat is only for convenience if running tests, not important at all

packages/coretypes/chainid.go
```diff
@@ -32,6 +33,16 @@ func NewChainIDFromBytes(data []byte) (ret ChainID, err error) {
 	return
 }
 
+// RandomChainID creates a random chain ID.
+func RandomChainID() ChainID {
+	return (ChainID)(address.Random())
```
lunfardo314, line 38: I think it can be `ChainID(address.Random())`. Just realized recently :)

packages/dkg/coordAPI.go
```diff
@@ -0,0 +1,64 @@
...
+
+// TODO: Should all the messages be signed? Move the signature
+// handling to the WaspClient and WebAPI?
+
+// The overall procedure is the following (from the coordinator's point of view):
+//
+// DkgInit(addrs, InitReq) -*->
+//     <-*- OK
+//
+// DkgStep(addrs, "sendDeals")
```
lunfardo314, line 17: yeah, why don't we do string constants. More idiomatic to Golang imo

packages/dkg/coordAPI.go
```diff
@@ -0,0 +1,64 @@
...
+	PeerPubs  [][]byte `json:"peerPubs"`
+	CoordPub  []byte   `json:"coordPub"`
+	Treshold  uint32   `json:"treshold"`
+	Version   byte     `json:"version"`
+	TimeoutMS uint64   `json:"timeoutMS"`
+}
+
+// StepReq is a message used to coordinate the DKG procedure by
+// ensuring the lock-step, as required by the DKG algorithm
+// assumptions (Rabin as well as Peddersen).
```
lunfardo314, line 51: typo? "Pedersen"?

packages/dkg/coordAPI.go
```diff
@@ -0,0 +1,64 @@
+package dkg
```
lunfardo314, line 1: it is a good style to include in comments links to original documents/articles

packages/dkg/coordinator.go
lunfardo314: Hans was refactoring all these types in Goshimmer, so maybe we wait for the new version and then follow the convention

packages/dkg/coordinator.go
```diff
@@ -0,0 +1,153 @@
...
+	"go.dedis.ch/kyber/v3/pairing"
+	"go.dedis.ch/kyber/v3/sign/bdn"
+	"go.dedis.ch/kyber/v3/sign/schnorr"
+)
+
+// GenerateDistributedKey is called from the client node to initiate the DKG
+// procedure on a set of nodes. The client is not required to have an instance
+// of the DkgNode, but may have it (be one of the peers sharing the secret).
+// This function works synchronously, so the user should run it async if needed.
+func GenerateDistributedKey(
```
lunfardo314, line 23: personally I like more Golang-style structures as named parameters, like `GenerateDistributedKeyParams` .
To avoid long lists of  arguments

packages/dkg/doc.go
```diff
@@ -0,0 +1,11 @@
+// Package dkg is responsible for performing a distributed key
+// generation procedure. The client-side part is implemented
+// as a coodinator, and the nodes sharing a generated secret
```
lunfardo314, line 3: coordinator is not a good name in the context of IOTA. Maybe "dealer" or "initiator"?

lunfardo314 (approved):

packages/dkg/proc.go
kape1395: Fixed.

packages/dkg/proc.go
kape1395: Replaced by integer constants.

packages/dkg/coordinator.go
kape1395: I missed `address.Version` somehow.
Replaced to use it.

packages/coretypes/chainid.go
kape1395: Good point, changed that. The same for (address.Address)(child).

packages/dkg/coordAPI.go
kape1395: Updated to use integer constants to be more aligned with the rest of the code.

packages/dkg/coordAPI.go
kape1395: Yes, thanks.

packages/dkg/coordAPI.go
kape1395: The coordination is not based on any particular paper.
Use of the Rabin algorithm is shown by the dedis/kyber imports.
Not sure, if we need to duplicate that info here. Or should I?

packages/dkg/coordinator.go
kape1395: Updated to use a structure.

packages/dkg/doc.go
kape1395: Renamed it to "Initiator". Dealer has another meaning in DKG.

packages/dkg/proc.go
kape1395: Made it a bit more compact.

PR #59, opened 2020-12-14
DKG and UDP peering.

kape1395:
- Rabin-based DKG implemented and integrated.
  - Peering interface extracted, UDP based implementation added.


client/dkshares.go
```diff
@@ -0,0 +1,56 @@
+package client
+
+// Copyright 2020 IOTA Stiftung
```
lunfardo314, line 1: I think recommended practice is to put licence statement before the `package` , see for example https://github.com/google/go-licenses/blob/master/licenses/classifier_test.go

main.go
```diff
@@ -12,18 +11,23 @@ import (
...
 	"github.com/iotaledger/wasp/plugins/publisher"
+	"github.com/iotaledger/wasp/plugins/registry"
 	"github.com/iotaledger/wasp/plugins/testplugins/nodeping"
 	"github.com/iotaledger/wasp/plugins/wasmtimevm"
 	"github.com/iotaledger/wasp/plugins/webapi"
+	"go.dedis.ch/kyber/v3/pairing"
 )
 
 func main() {
+	suite := pairing.NewSuiteBn256() // TODO: [KP] Single suite should be used in all the places.
```
lunfardo314, line 29: Is it really necessary to use the same copy of suite everywhere?

packages/chain/chainimpl/chainobj.go
```diff
@@ -59,61 +73,52 @@ func newCommitteeObj(chr *registry.ChainRecord, log *logger.Logger, onActivation
 			addr.String(), chr.CommitteeNodes)
 		return nil
 	}
-	a := (address.Address)(chr.ChainID)
-	dkshare, keyExists, err := registry.GetDKShare(&a)
+	dkshare, err := dksProvider.LoadDKShare(&chr.ChainID)
```
lunfardo314: I see chainID is now used as a key for DKShare  in the registry. **This is not correct**. DKShare must be accessed with the address (hash of the public key, corresponding to the distributed secret). ChainID is an alias of Address just temporary, until next refactoring of Goshimmer, with color locks. Then ChainID will become alias of the Color and ChainID will become impossible to use as a key for DKShare.

packages/chain/chainimpl/interface.go
```diff
@@ -166,32 +166,31 @@ func (c *chainObj) ReceiveMessage(msg interface{}) {
 	}
 }
 
-// sends message to peer by index. It can be both committee peer or access peer
+// SendMsg sends message to peer by index. It can be both committee peer or access peer.
+// TODO: [KP] Maybe we can use a broadcast instead of this?
```
lunfardo314, line 170: essentially it is a broadcast, it also can be done in parallel

packages/coretypes/chainid.go
```diff
@@ -74,3 +74,8 @@ func (chid *ChainID) Read(r io.Reader) error {
 	}
 	return nil
 }
+
+// Equal does what it should.
+func (chid *ChainID) Equal(other *ChainID) bool {
```
lunfardo314, line 79: why we need this? ChainID is an byte array, can be compared by ==

packages/peering/group/group.go
```diff
@@ -0,0 +1,210 @@
...
+	return uint16(NotInGroup), errors.New("peer_not_found_by_pub_key")
+}
+
+// SendMsgByIndex implements peering.GroupProvider.
+func (g *groupImpl) SendMsgByIndex(peerIdx uint16, msg *peering.PeerMessage) {
+	g.nodes[peerIdx].SendMsg(msg)
+}
+
+// Broadcast implements peering.GroupProvider.
+func (g *groupImpl) Broadcast(msg *peering.PeerMessage, includingSelf bool) {
```
lunfardo314, line 79: may be spawned go routines to send in parallel. If that makes sense at all

packages/registry/dkshare.go
```diff
@@ -1,68 +1,45 @@
...
-	if err != nil {
-		return nil, false, err
-	}
-	return ret, true, nil
 }
 
-func LoadDKShare(addr *address.Address, maskPrivate bool) (*tcrypto.DKShare, error) {
-	data, err := database.GetRegistryPartition().Get(dbkey(addr))
+// LoadDKShare implements dkg.RegistryProvider.
+func (r *Impl) LoadDKShare(chainID *coretypes.ChainID) (*tcrypto.DKShare, error) {
```
lunfardo314: already commented above: address must be a key for DKShare in the registry, not ChainID

lunfardo314 (changes requested):

client/dkshares.go
kape1395: Updated.

main.go
kape1395: I think it will be easier to ensure uniformity of the suite in use.

packages/chain/chainimpl/chainobj.go
kape1395: Fixed. Removed ChaniID from DKShare and related places.

packages/chain/chainimpl/interface.go
kape1395: I think there is no need to make that in parallel now, because the sending part is not a bootleeneck.

packages/coretypes/chainid.go
kape1395: It is unclear from the calling code, if the == comparison works for this kind of objects.
That's why I added this function.

packages/peering/group/group.go
kape1395: That is not needed for now.

packages/registry/dkshare.go
kape1395: Fixed.

lunfardo314 (approved):

PR #60, opened 2020-12-16
Blobs

dessaya:
* Added `listBlobs` view to `blob` SC
* `wasp-cli` now can manipulate blobs
* Dashboard now shows blobs

### DIFF
--- a/tools/cluster/demo/tests/blob-test.sh
+++ b/tools/cluster/demo/tests/blob-test.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+DIR="${BASH_SOURCE%/*}"
+if [[ ! -d "$DIR" ]]; then DIR="$PWD"; fi
+. "$DIR/common.sh"
+
+wasp-cli init
+wasp-cli request-funds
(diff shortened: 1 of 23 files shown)


tools/cluster/demo/tests/blob-test.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
kape1395, line 1: Wouldn't is be better to write these tests in go? They would be run by `go test ./...`, thus easier to maintain them working?

kape1395 (commented):
Just a question on tests.

tools/cluster/demo/tests/blob-test.sh
dessaya: I don't know, maybe? These tests require lots of string parsing in order to interpret the command output and make assertions. If they were written in Go, I estimate they would take 10x - 20x LOC and be way more cumbersome to write.

kape1395: OK

lunfardo314 (approved):

PR #61, opened 2020-12-17
Fee+owner

dessaya:
Dashboard and `wasp-cli` now show fee & chain owner information.

### DIFF
--- a/packages/kv/codec/color.go
+++ b/packages/kv/codec/color.go
@@ -4,14 +4,14 @@ import (
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/balance"
 )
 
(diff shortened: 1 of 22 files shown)


packages/kv/codec/color.go
```diff
@@ -4,14 +4,14 @@ import (
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/balance"
 )
 
-func DecodeColor(b []byte) (*balance.Color, bool, error) {
+func DecodeColor(b []byte) (balance.Color, bool, error) {
```
lunfardo314, line 7: that is good refactor. I think we need also to refactor all returns of `*hashing.HashValue` and similar pointers

lunfardo314 (approved):

PR #65, opened 2020-12-22
Dashboard cli log

dessaya:
* Dashboard now shows the log of each contract
* Added the `wasp-cli chain log <contract-name>` command
* Improved styles and navigation in dashboard

### DIFF
--- a/packages/dashboard/chainaccount.go
+++ b/packages/dashboard/chainaccount.go
@@ -63,24 +63,20 @@ const tplChainAccount = `
 {{define "title"}}On-chain account details{{end}}
 
 {{define "body"}}
-<div class="container">
-<div class="row">
-<div class="col-sm">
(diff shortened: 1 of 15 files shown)


packages/dashboard/chainaccount.go
```diff
@@ -11,7 +12,7 @@ import (
 	"github.com/labstack/echo"
 )
 
-const chainAccountRoute = "/chains/:chainid/account/:agentid"
+const chainAccountRoute = "/chains/:chainid/account/:type/:id"
```
kape1395: `chains` is plural and `account` is singular.
Maybe we should take uniform approach for that?
I would prefer the singular form.

kape1395 (commented):
A minor comment regarding the URL paths.

PR #67, opened 2020-12-30
Add `/doc` & `/doc/swagger.json` endpoints for API specification

dessaya:
This PR adds a specification for the wasp webapi in Swagger format.

Screenshot from 2020-12-30 19-56-39

This uses https://github.com/pangpanglabs/echoswagger which allows to dynamically generate `swagger.json` from the echo routes, so node code generation step is necessary.


client/chainrecord.go
```diff
@@ -10,7 +10,7 @@ import (
 
 const (
 	PutChainRecordRoute     = "chainrecord"
-	GetChainRecordListRoute = "chainrecord"
+	GetChainRecordListRoute = "chainrecords"
```
kape1395, line 13: As I remember, we have moved to the singular form in the URLs?

client/jsonable/address.go
```diff
@@ -2,59 +2,35 @@ package jsonable
...
 
-func NewAddress(address *address.Address) *Address {
-	return &Address{address: *address}
+func NewAddress(address *address.Address) Address {
+	return Address(address.String())
 }
 
 func (a Address) MarshalJSON() ([]byte, error) {
-	return json.Marshal(a.Address().String())
+	return json.Marshal(string(a))
```
kape1395, line 17: Does it encode []byte to base64?
I thought it just interprets all the bytes as string characters, which is incorrect in this case.

plugins/webapi/admapi/dkshares.go
```diff
@@ -90,22 +117,25 @@ func handleDKSharesGet(c echo.Context) error {
 func makeDKSharesInfo(dkShare *tcrypto.DKShare) (*client.DKSharesInfo, error) {
 	var err error
 
-	var sharedPubKeyBin []byte
-	if sharedPubKeyBin, err = dkShare.SharedPublic.MarshalBinary(); err != nil {
+	b, err := dkShare.SharedPublic.MarshalBinary()
```
kape1395, line 120: Don't you mess with the previously declared err variable in this way?

kape1395 (commented):
Cool PR! Several questions inline.

plugins/webapi/admapi/dkshares.go
lunfardo314: this pretty common in my practice. Sometimes you need separate err and then you double-declare in some situations. Not big deal imo

client/chainclient/status.go
```diff
@@ -17,7 +17,7 @@ type SCStatus struct {
 	StateIndex uint32
 	Timestamp  time.Time
 	StateHash  *hashing.HashValue
-	StateTxId  *valuetransaction.ID
+	StateTxId  valuetransaction.ID
```
lunfardo314, line 20: I think it is a good practice here to use object instead of pointers, to make GC less involved. We probably should do it for other packages, like hashing

client/jsonable/address.go
```diff
@@ -2,59 +2,35 @@ package jsonable
...
 
-type Address struct {
-	address address.Address
-}
-
-type ChainID struct {
-	chainID coretypes.ChainID
-}
+// Address is the base58-encoded representation of address.Address
+type Address string
```
lunfardo314, line 10: I'd better avoid using same type names, especially ubiquitous publics like Address, if they are only used for some function. In my count there are at least 3 different types named Address in wasp/goshimmer, different packages of course.  Probably matter of style and taste but maybe there's better naming convention for this?

client/jsonable/address.go
lunfardo314: this may be confusing indeed but it is correct. The `Address(address.String())` first converts `address` to base58, then it interprets that string as in `type Address string`, the alias of string name `Address`. The Marshal method just outputs that base58 string as a byte array.

client/chainrecord.go
dessaya: IIRC in the dashboard (which is served in a different port) we changed `/chain` (which displays the list of chains) to `/chains`, so I felt this change was consistent. I'm not very opinionated about this anyway. I'm fine with reverting or change it to something else.

client/jsonable/address.go
dessaya: Right, what I changed here is where the base58 conversion is made, but the behavior is the same.

plugins/webapi/admapi/dkshares.go
dessaya: I'm not 100% sure but think that `:=` only creates new variables if they were previously undeclared. Eg:

```golang
var a int
var c int
a, b, c := f() // this only creates b
```

client/chainclient/status.go
dessaya: Agree, I'm changing this organically as I refactor other things.

client/jsonable/address.go
dessaya: I don't see it as a serious problem because it is prefixed by the namespace; ie `address.Address` is easily distinguishable from `jsonable.Address`.

But I'm also happy to change the name if we find a good one.  `jsonable.JSONAddress?` `jsonable.Base58Address?`

PR #68, opened 2020-12-31
refactor fee handling and panic rollback logic 

lunfardo314:
Now VMContext for each request first does the request token and fee token handling and only after that takes snapshot of the state update and transaction builder. Then the VM is wrapped into the panic catcher and invoked. 
If VM call panics or the top request call returns an error, the state and transaction builder is recovered to the point where request and fee tokens are already settled in the state. 
The panics during fee processing is a internal error and should never happen
TODO: fallback policy for the rest of the transfer
All tests pass

### DIFF
--- a/packages/coretypes/hname.go
+++ b/packages/coretypes/hname.go
@@ -6,7 +6,6 @@ package coretypes
 import (
 	"bytes"
 	"encoding/binary"
-	"fmt"
 	"io"
 	"strconv"
 
@@ -26,12 +25,7 @@ const FuncInit = "init"
 // EntryPointInit is a hashed name of the init function
 var EntryPointInit = Hn(FuncInit)
 
-func init() {
-	fmt.Printf("----------------\nreserved entry pont '%s': %s\n----------------\n",
-		FuncInit, EntryPointInit.String())
-}
-
-// NewHnameFromBytes constructur, unmarshalling
+// NewHnameFromBytes constructor, unmarshalling
 func NewHnameFromBytes(data []byte) (ret Hname, err error) {
 	err = ret.Read(bytes.NewReader(data))
 	return
--- a/packages/coretypes/hname.go
+++ b/packages/coretypes/hname.go
@@ -6,6 +6,7 @@ package coretypes
 import (
 	"bytes"
 	"encoding/binary"
+	"fmt"
 	"io"
 	"strconv"
 
@@ -24,6 +25,11 @@ const FuncInit = "init"
 
 // EntryPointInit is a hashed name of the init function
 var EntryPointInit = Hn(FuncInit)
+
+func init() {
+	fmt.Printf("----------------\nreserved entry pont '%s': %s\n----------------\n",
+		FuncInit, EntryPointInit.String())
+}
 
 // NewHnameFromBytes constructur, unmarshalling
 func NewHnameFromBytes(data []byte) (ret Hname, err error) {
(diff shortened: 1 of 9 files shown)


packages/coretypes/hname.go
```diff
@@ -25,6 +26,11 @@ const FuncInit = "init"
 // EntryPointInit is a hashed name of the init function
 var EntryPointInit = Hn(FuncInit)
 
+func init() {
+	fmt.Printf("----------------\nreserved entry pont '%s': %s\n----------------\n",
+		FuncInit, EntryPointInit.String())
+}
+
```
dessaya: Is this change intended to be merged? Or was it used to debug something and no longer needed?

packages/vm/runvm/runtask.go
```diff
@@ -90,12 +93,11 @@ func runTask(task *vm.VMTask, txb *statetxbuilder.Builder) {
 	}
 	// Note: can't take tx ID!!
 	task.Log.Debugw("runTask OUT",
-		"result batch size", task.ResultBlock.Size(),
-		"result batch state index", task.ResultBlock.StateIndex(),
-		"result variable state hash", stateHash.String(),
-		"result essence hash", hashing.HashData(task.ResultTransaction.EssenceBytes()).String(),
-		"result tx finalTimestamp", time.Unix(0, task.ResultTransaction.MustState().Timestamp()),
+		"batch size", task.ResultBlock.Size(),
```
dessaya, line 96: Should this be "block size"?

packages/vm/vmcontext/call.go
```diff
@@ -52,15 +50,15 @@ func (vmctx *VMContext) Call(contract coretypes.Hname, epCode coretypes.Hname, p
...
 	}
-	if err := vmctx.pushCallContextWithTransfer(contract, params, transfer); err != nil {
+	if err := vmctx.pushCallContextWithTransfer(targetContract, params, transfer); err != nil {
 		return nil, err
 	}
 	defer vmctx.popCallContext()
 
-	// prevent calling 'init' not from root contract or not while initializing root
-	if epCode == coretypes.EntryPointInit && contract != root.Interface.Hname() {
+	// prevent calling 'init' not from root targetContract or not while initializing root
```
dessaya: This rename was unintentional I think?

packages/vm/vmcontext/call.go
```diff
@@ -52,15 +50,15 @@ func (vmctx *VMContext) Call(contract coretypes.Hname, epCode coretypes.Hname, p
...
+	if epCode == coretypes.EntryPointInit && targetContract != root.Interface.Hname() {
 		if !vmctx.callerIsRoot() {
-			return nil, fmt.Errorf("attempt to call init not from root contract")
+			return nil, fmt.Errorf("attempt to call init not from root targetContract")
```
dessaya: Also this

packages/vm/vmcontext/runreq.go
```diff
@@ -3,59 +3,80 @@ package vmcontext
...
+	vmctx.lastError = nil
 	func() {
 		// panic catcher for the whole call from request to the VM
 		defer func() {
 			if r := recover(); r != nil {
 				vmctx.lastResult = nil
 				vmctx.lastError = fmt.Errorf("recovered from panic in VM: %v", r)
-				vmctx.log.Error(vmctx.lastError)
-				debug.PrintStack()
+				//debug.PrintStack()
```
dessaya: If this is no longer needed can it be removed instead of commented out?

dessaya (approved):

packages/vm/runvm/runtask.go
lunfardo314: it might be but I use 'batch' for 'batch of requests' and batch size is number of requests. While "block size" is often understood as size in bytes in the block

PR #69, opened 2021-01-01
free token handling

lunfardo314:
Now all tokens in the request transaction which are in excess to the tokens specified in the request metadata are 
considered free and are accrued to the chain owner (after fees).


packages/vm/core/root/internal.go
```diff
@@ -86,22 +86,25 @@ func GetChainInfo(state kv.KVStore) (*ChainInfo, error) {
 // GetFeeInfo is an internal utility function which returns fee info for the contract
 // It is called from within the 'root' contract as well as VMContext and viewcontext objects
 // It is not exposed to the sandbox
-func GetFeeInfo(state kv.KVStore, hname coretypes.Hname) (balance.Color, int64, int64, error) {
-	rec, err := FindContract(state, hname)
-	if err != nil {
-		// contract not found
-		return balance.Color{}, 0, 0, err
+func GetFeeInfo(state kv.KVStore, hname coretypes.Hname) (balance.Color, int64, int64) {
+	rec, _ := FindContract(state, hname)
```
dessaya, line 90: Not sure why the error is ignored? Errors should never pass silently. At least `if err != nil { panic(err) }`?

packages/vm/core/root/internal.go
```diff
@@ -86,22 +86,25 @@ func GetChainInfo(state kv.KVStore) (*ChainInfo, error) {
...
+
+func GetFeeInfoByContractRecord(state kv.KVStore, rec *ContractRecord) (balance.Color, int64, int64) {
+	var ownerFee, validatorFee int64
+	if rec != nil {
+		ownerFee = rec.OwnerFee
+		validatorFee = rec.ValidatorFee
 	}
-	feeColor, defaultOwnerFee, defaultValidatorFee, err := GetDefaultFeeInfo(state)
-	ownerFee := rec.OwnerFee
+	feeColor, defaultOwnerFee, defaultValidatorFee, _ := GetDefaultFeeInfo(state)
```
dessaya, line 100: Same here

packages/vm/vmcontext/runreq.go
```diff
@@ -2,36 +2,42 @@ package vmcontext
...
+		return
 	}
 	// snapshot state baseline for rollback in case of panic
 	snapshotTxBuilder := vmctx.txBuilder.Clone()
 	snapshotStateUpdate := vmctx.stateUpdate.Clone()
 
 	vmctx.lastError = nil
 	func() {
-		// panic catcher for the whole call from request to the VM
+		// panic catcher for the whole callByProgramHash from request to the VM
```
dessaya, line 40: unintended rename?

packages/vm/core/root/internal.go
lunfardo314: returns nil if contract not found. The rest is default processing. No need to panic.

packages/vm/core/root/internal.go
lunfardo314: returns default values

packages/vm/core/root/internal.go
dessaya: But the error could be a DB error? Schema inconsistency? Etc.

PR #71, opened 2021-01-06
Support for single node chains.

kape1395:
DKG and Chain adjusted to work with N=1 as well.


packages/dkg/proc.go
```diff
@@ -178,6 +195,9 @@ func (p *proc) processLoop(timeout time.Duration, doneCh chan map[uint16]*peerin
 //
 func (p *proc) rabinStep1R21SendDealsMakeSent(step byte, initRecv *peering.RecvEvent, prevMsgs map[uint16]*peering.PeerMessage) (map[uint16]*peering.PeerMessage, error) {
 	var err error
+	if p.dkgImpl == nil {
+		return nil, errors.New("unexpected_step_for_n=1")
```
dessaya: Just curious, why use _ instead of spaces here?

dessaya (approved):

packages/dkg/proc.go
kape1395: I'm not sure either. The style guide says it should begin with lower case and end without a period. Intuitively I assume, if that's not a sentence, then maybe a mnemonic. I can change it to spaces, if that's more common.

lunfardo314: I'd use spaces but it is a matter of taste I think

lunfardo314 (approved):

packages/dkg/proc.go
kape1395: I will fix that.

PR #79, opened 2021-01-27
Blobref refactor

lunfardo314:
- refactored database access (dbprovider package)
- implemented Put/Get interface for binary blobs into the registry


packages/dbprovider/db.go
```diff
@@ -0,0 +1,87 @@
...
+	"sync"
+	"time"
+)
+
+type DBProvider struct {
+	log             *logger.Logger
+	db              database.DB
+	store           kvstore.KVStore
+	partitions      map[coretypes.ChainID]kvstore.KVStore
+	partitionsMutex *sync.RWMutex
```
dessaya, line 18: I know this is not new code, but I'm curious: why is this `partitions` map needed? Is it a cache to reduce allocations?

packages/registry/blobcache.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/iotaledger/wasp/packages/dbprovider"
+	"github.com/iotaledger/wasp/packages/hashing"
+)
+
+func dbKeyForBlob(h hashing.HashValue) []byte {
+	return dbprovider.MakeKey(dbprovider.ObjectTypeBlob, h[:])
+}
+
+// Writes data into the registry with the key of its hash
+func (r *Impl) PutBlob(data []byte) error {
```
dessaya: Maybe `PutBlob` can return the hash?

packages/sctransaction/reqargs.go
```diff
@@ -0,0 +1,89 @@
...
+// RequestArguments encodes request parameters taking into account hashes of data blobs
+type RequestArguments dict.Dict
+
+// NewRequestArguments constructor
+func NewRequestArguments() RequestArguments {
+	return RequestArguments(dict.New())
+}
+
+// AddArgument add new ordinary argument. Encodes the key as "normal"
+func (a RequestArguments) AddArgument(name string, data []byte) {
```
dessaya: suggest renaming to just `Add`

packages/sctransaction/reqargs.go
```diff
@@ -0,0 +1,89 @@
...
+	a[kv.Key("-"+name)] = data
+}
+
+// AddArgumentAsBlobHash adds argument with the data hash instead of data itself.
+// Encodes key as "blob reference"
+func (a RequestArguments) AddArgumentAsBlobHash(name string, data []byte) {
+	h := hashing.HashData(data)
+	a[kv.Key("*"+name)] = h[:]
```
dessaya, line 28: Isn't this susceptible to bugs? e.g. what if one of the regular keys starts with `*`?

Would something like this be simpler?:
```
type RequestArguments struct {
  Args dict.Dict
  BlobHashes dict.Dict
}
```

packages/sctransaction/reqargs.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// HasBlogRef return if request arguments contain at least one blog reference
```
dessaya: typo: Blog != Blob (there are many instances of this)

plugins/database/plugin.go
```diff
@@ -44,35 +40,46 @@ func configure(_ *node.Plugin) {
 	}
 
 	// we open the database in the configure, so we must also make sure it's closed here
-	err = daemon.BackgroundWorker(pluginName, closeDB, parameters.PriorityDatabase)
+	err = daemon.BackgroundWorker(pluginName, func(shutdownSignal <-chan struct{}) {
+		<-shutdownSignal
+		log.Infof("Syncing database to disk...")
+		dbProvider.Close(shutdownSignal)
```
dessaya: Why does Close receive the `shutdownSignal` if it has already been triggered?

plugins/database/plugin.go
```diff
@@ -44,35 +40,46 @@ func configure(_ *node.Plugin) {
...
-		log.Panicf("Failed to start as daemon: %s", err)
+		log.Panicf("Failed to start a daemon: %s", err)
 	}
 }
 
 func run(_ *node.Plugin) {
-	if err := daemon.BackgroundWorker(pluginName+"[GC]", runGC, parameters.PriorityBadgerGarbageCollection); err != nil {
+	if err := daemon.BackgroundWorker(pluginName+"[GC]", func(shutdownSignal <-chan struct{}) {
+		dbProvider.RunGC(shutdownSignal)
+	}, parameters.PriorityBadgerGarbageCollection); err != nil {
```
dessaya: can this be simplified to this?

`daemon.BackgroundWorker(pluginName+"[GC]", dbProvider.RunGC, parameters.PriorityBadgerGarbageCollection)`

packages/dbprovider/db.go
lunfardo314: essentially yes. I made it because in the past partitions weren't cached by kvstore, so this is precaution. It may be removed, after deeper analysis

packages/registry/blobcache.go
lunfardo314: yeah, why not.

packages/sctransaction/reqargs.go
lunfardo314: ok

packages/sctransaction/reqargs.go
lunfardo314: what you proposed that was my first implementation :)  However I decided for prefixes. The collisions with names starting with '*' won't happen because currently all keys are prefixed by exactly one byte, all of them are removed automatically when decoded

packages/sctransaction/reqargs.go
lunfardo314: ok

plugins/database/plugin.go
lunfardo314: good point

plugins/database/plugin.go
lunfardo314: good catch

PR #81, opened 2021-01-28
add endpoints for blob registry

dessaya:


client/blob.go
```diff
@@ -0,0 +1,34 @@
...
+// PutBlob uploads a blob to the registry
+func (c *WaspClient) PutBlob(data []byte) (hashing.HashValue, error) {
+	req := model.NewBlobData(data)
+	res := &model.BlobInfo{}
+	err := c.do(http.MethodGet, routes.PutBlob(), req, res)
+	return res.Hash.HashValue(), err
+}
+
+// GetBlob fetches a blob by its hash
+func (c *WaspClient) GetBlob(hash hashing.HashValue) ([]byte, error) {
```
lunfardo314, line 20: interestingly, `GetBlob` is not even needed in API, only `PutBlob` and `HasBlob`

lunfardo314 (approved):

PR #84, opened 2021-02-04
Added extension solo.NewSignatureScheme

brunoamancio:
It is now possible to create wallets with zero balance, since it is very useful to be able to create addresses with zero balance for unit tests with extensions directly from solo.

### DIFF
--- a/packages/solo/solo.go
+++ b/packages/solo/solo.go
@@ -4,6 +4,10 @@
 package solo
 
 import (
+	"sync"
+	"testing"
+	"time"
+
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/address"
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/address/signaturescheme"
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/balance"
@@ -27,9 +31,6 @@ import (
 	"github.com/iotaledger/wasp/plugins/wasmtimevm"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zapcore"
-	"sync"
-	"testing"
-	"time"
 )
 
 // DefaultTimeStep is a default step for the logical clock for each PostRequest call.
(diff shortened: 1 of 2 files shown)


packages/solo/solo.go
```diff
@@ -320,11 +321,26 @@ func (env *Solo) NewSignatureSchemeWithFunds() signaturescheme.SignatureScheme {
...
-	keypair := ed25519.GenerateKeyPair()
-	ret := signaturescheme.ED25519(keypair)
+	ret, pubKeyBytes := env.NewSignatureSchemeAndPubKey()
 	_, err := env.utxoDB.RequestFunds(ret.Address())
 	require.NoError(env.T, err)
-	env.AssertAddressBalance(ret.Address(), balance.ColorIOTA, testutil.RequestFundsAmount)
+	return ret, pubKeyBytes
+}
+
+// NewSignatureScheme generates new ed25519 signature scheme
```
lunfardo314, line 330: The godoc comment is not correct. The empty signature scheme is created independendently from the UTXODB

packages/solo/solo.go
```diff
@@ -320,11 +321,26 @@ func (env *Solo) NewSignatureSchemeWithFunds() signaturescheme.SignatureScheme {
...
+// from the UTXODB faucet.
+func (env *Solo) NewSignatureScheme() signaturescheme.SignatureScheme {
+	ret, _ := env.NewSignatureSchemeAndPubKey()
+	return ret
+}
+
+// NewSignatureSchemeAndPubKey generates new ed25519 signature scheme
```
lunfardo314, line 336: see above godoc comment

lunfardo314:
thanks for PR. So, you added new functions for empty sig scheme.
Actually, it was intentionally implemented 'withFunds' to reduce boilerplating and also to keep number of Solo functions minimal. Normally you need some funds in the address and if you don't, you just do not use it.
I will merge after minor corrections I requested.

lunfardo314 (changes requested):

packages/solo/solo.go
brunoamancio: > thanks for PR. So, you added new functions for empty sig scheme.
> Actually, it was intentionally implemented 'withFunds' to reduce boilerplating and also to keep number of Solo functions minimal. Normally you need some funds in the address and if you don't, you just do not use it.

I agree that often you need funds. But in cases where you are testing sending and receiving funds from an address to a chain and back to the value tangle, it becomes useful to verify who has funds without having to offset the balances with the default amount of funds when verifying them. It is very useful when writing more complex tests.

> I will merge after minor corrections I requested.

I will make the changes asap.

brunoamancio:
Committed the requested changes and fixed a pre-existing typo.

lunfardo314 (approved):

PR #85, opened 2021-02-11
wasp-cli: autodetect chain nodes to upload blob

dessaya:


tools/wasp-cli/chain/committee.go
```diff
@@ -0,0 +1,17 @@
+package chain
+
+import (
+	"github.com/iotaledger/wasp/tools/wasp-cli/config"
+	"github.com/iotaledger/wasp/tools/wasp-cli/log"
+)
+
+func chainCommittee() []int {
+	chain, err := config.WaspClient().GetChainRecord(GetCurrentChainID())
```
lunfardo314, line 9: in the future we should not use this function of getting all committee nodes without proper authetication

lunfardo314 (approved):

PR #89, opened 2021-02-11
Vmtest

BugFreeSoftware:
Merged develop
Added initial documentation
Cleaned up some leftover files
Used proper build tags instead of all the t.SkipNow() calls
  tests can now be ran (after building wasm code first) by adding -tags=wasmtest
  all other tests are unaffected and will run the same


lunfardo314 (changes requested):
We need to:
- move all TinyGo stuff (tests etc) to a separate directory named  `experimental`
- move all `schema` related stuff (tests, generated code etc) to `experimental`
- get rid of `testenv` package
- leave the `experimental` stuff outside the `develop` branch. Let it reside in separate `experimental` branch and not present in `develop`

BugFreeSoftware:
Closing this one and will redo them as separate commits

PR #94, opened 2021-02-15
Experimental

BugFreeSoftware:
Removal of everything classes as experimental from develop branch


packages/solo/solo.go
```diff
@@ -36,6 +36,10 @@ import (
 // DefaultTimeStep is a default step for the logical clock for each PostRequest call.
 const DefaultTimeStep = 1 * time.Millisecond
 
+// default supply of tokens returned by the UTXODB faucet
+// which is therefore the amount returned by NewSignatureSchemeWithFunds() and such
+const Supply = 1337
```
lunfardo314, line 41: actually, the 1337 constant is defined in utxodb. 
But I think it is useful to have it in Solo too, only would be good it to be defined through utxodb constant.
Also, the name `Supply` is not good, it is not a supply in the ledger. Something `WalletFunds`, `MaxWalletFunds` or `FaucetFunds` may be better

contracts/common/deploy.go
```diff
@@ -0,0 +1,40 @@
...
+	TraceHost  = true
+)
+
+var (
+	ContractAccount coretypes.AgentID
+	ContractId      coretypes.ContractID
+	CreatorWallet   signaturescheme.SignatureScheme
+)
+
+func DeployContract(t *testing.T, scName string) *solo.Chain {
```
lunfardo314, line 25: Not to be confused with more general function, rename it to `NewChainWithWasmContract` or something.
Note that environment and other attributed will be available via from chain

lunfardo314 (approved):

PR #97, opened 2021-02-16
ver 0.1.0

lunfardo314:
PR before merging into `master` and making `master` default branch

### DIFF
--- a/contracts/getproc.go
+++ b/contracts/getproc.go
@@ -3,8 +3,8 @@ package contracts
 import (
 	"fmt"
 	"github.com/iotaledger/wasp/packages/coretypes"
+	"github.com/iotaledger/wasp/packages/coretypes/coreutil"
 	"github.com/iotaledger/wasp/packages/hashing"
-	"github.com/iotaledger/wasp/packages/vm/contract"
 	"sync"
 )
 
@@ -15,18 +15,18 @@ var (
 	allExamplesMutex = &sync.Mutex{}
 )
 
-// AddProcessor adds new processor to the runtime registry of example processors.
+// AddExampleProcessor adds new processor to the runtime registry of example processors.
 // The 'proc' represents executable of the specific smart contract.
-// It must implement vmtypes.Processor
-func AddProcessor(c *contract.ContractInterface) {
+// It must implement coretypes.Processor
+func AddExampleProcessor(c *coreutil.ContractInterface) {
 	allExamplesMutex.Lock()
 	defer allExamplesMutex.Unlock()
 	allExamples[c.ProgramHash] = c
-	fmt.Printf("added example processor name: '%s', program hash: %s, dscr: '%s'\n",
+	fmt.Printf("----- AddExampleProcessor: name: '%s', program hash: %s, description: '%s'\n",
 		c.Name, c.ProgramHash.String(), c.Description)
 }
 
-// GetExampleProcessor retrieves smart contract processor (VM) by the hash (whith existence flag)
+// GetExampleProcessor retrieves smart contract processor (VM) by the hash (with existence flag)
 func GetExampleProcessor(progHash hashing.HashValue) (coretypes.Processor, bool) {
 	ret, ok := allExamples[progHash]
 	if !ok {
--- a/contracts/getproc.go
+++ b/contracts/getproc.go
@@ -1,3 +1,6 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
 package contracts
 
 import (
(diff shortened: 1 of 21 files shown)


contracts/getproc.go
```diff
@@ -15,18 +15,18 @@ var (
 	allExamplesMutex = &sync.Mutex{}
 )
 
-// AddProcessor adds new processor to the runtime registry of example processors.
+// AddExampleProcessor adds new processor to the runtime registry of example processors.
 // The 'proc' represents executable of the specific smart contract.
-// It must implement vmtypes.Processor
-func AddProcessor(c *contract.ContractInterface) {
+// It must implement coretypes.Processor
+func AddExampleProcessor(c *coreutil.ContractInterface) {
```
kape1395, line 24: Not sure, if it is OK to gave ExampleProcessor in a package that is not an example.

packages/coretypes/coreutil/contract.go
```diff
@@ -1,4 +1,6 @@
-package contract
+// package coreutil provides functions to describe interface of the core contract
```
kape1395, line 4: License headers missing.

packages/vm/core/accounts/impl.go
```diff
@@ -3,8 +3,8 @@ package accounts
 import (
 	"fmt"
 	"github.com/iotaledger/wasp/packages/coretypes"
+	assert2 "github.com/iotaledger/wasp/packages/coretypes/assert"
```
kape1395: Why do we need an alias here?

packages/vm/core/blob/impl.go
```diff
@@ -3,7 +3,7 @@ package blob
 import (
 	"fmt"
 	"github.com/iotaledger/wasp/packages/coretypes"
-	"github.com/iotaledger/wasp/packages/coretypes/coreutil"
+	assert2 "github.com/iotaledger/wasp/packages/coretypes/assert"
```
kape1395: Why the alias is needed here?

packages/vm/core/root/impl.go
```diff
@@ -10,7 +10,7 @@ import (
 	"fmt"
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/balance"
 	"github.com/iotaledger/wasp/packages/coretypes"
-	"github.com/iotaledger/wasp/packages/coretypes/coreutil"
+	assert2 "github.com/iotaledger/wasp/packages/coretypes/assert"
```
kape1395, line 13: The same here.

packages/vm/core/testcore/sandbox_tests/test_sandbox_sc/impl_check_ctx.go
```diff
@@ -2,7 +2,7 @@ package test_sandbox_sc
 
 import (
 	"github.com/iotaledger/wasp/packages/coretypes"
-	"github.com/iotaledger/wasp/packages/coretypes/coreutil"
+	assert2 "github.com/iotaledger/wasp/packages/coretypes/assert"
```
kape1395: and here

plugins/banner/plugin.go
```diff
@@ -11,7 +11,7 @@ const PluginName = "Banner"
 
 const (
 	// AppVersion version number
-	AppVersion = "v0.0.1"
+	AppVersion = "v0.1.0"
```
kape1395, line 14: Maybe it should be taken from `git describe --tags` automatically?

kape1395: One way to do that: https://stackoverflow.com/questions/11354518/application-auto-build-versioning

kape1395 (commented):
Comments are mainly on naming.

packages/coretypes/coreutil/contract.go
lunfardo314: fixed

contracts/getproc.go
lunfardo314: essentially all those contracts are examples. Cannot come to better categorization atm

packages/vm/core/accounts/impl.go
lunfardo314: no need. Just artifact of refactoring. Fixed

packages/vm/core/blob/impl.go
lunfardo314: fixed

plugins/banner/plugin.go
lunfardo314: I know, looked at it, seems complicated, postponed for the future as TODO

kape1395 (approved):

PR #103, opened 2021-02-18
cleanup package structure

lunfardo314:
- minor adjustments to examples


contracts/examples_core/getproc.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-package contracts
+package examples_core
```
dessaya, line 4: Using underscores in package names is not idiomatic Go: https://blog.golang.org/package-names

Is `_core` really important? Can this package be just `contracts/examples`?

dessaya (approved):

PR #108, opened 2021-02-21
Support for "MintSupply" function in apilib, sandbox and Solo

lunfardo314:
- support for `MintSupply` function in apilib, sandbox and Solo
- revisited tx builder + tests
- concurrency: adjustments in Solo and tests


packages/apilib/request.go
```diff
@@ -54,6 +57,10 @@ func CreateRequestTransaction(par CreateRequestTransactionParams) (*sctransactio
 			return nil, err
 		}
 	}
+	txb.AddMinting(par.Mint)
+	//for addr, amount := range par.Mint{
+	//	txb.MintColoredTokens
+	//}
```
dessaya: Is the commented out code necessary?

packages/coretypes/sandbox.go
```diff
@@ -56,7 +63,7 @@ type Sandbox interface {
 	Utils() Utils
 }
 
-// PostRequestParams is parameters of the PostRequest call
+// PostRequestParams is parameters of the PostRequestSync call
```
dessaya: This rename was probably unintended?

packages/sctransaction/properties/properties.go
```diff
@@ -39,20 +45,24 @@ type Properties struct {
...
 	// free tokens: tokens with output to chain address - tokens transferred by requests - request tokens - chain token
 	// In most cases it is empty, because all tokens should be transferred with requests
 	// Free tokens normally should be returned to the sender
 	freeTokensByAddress map[address.Address]coretypes.ColoredBalances
 }
 
-func (tx *Transaction) calcProperties() (*Properties, error) {
-	ret := &Properties{
+func init() {
+	sctransaction.NewProperties = calcProperties
```
dessaya: Not a fan of this pattern :sweat_smile: -- it makes code harder to understand and follow, and it requires the awkward `_` imports.

I'm guessing it is due to a cyclic dependency. Can `calcProperties` be public and just call this to create a new instance? (ie, get rid of `sctransaction.NewProperties`).

Or have `sctransaction.Properties()` explicitly receive the constructor as an argument.

Or just have everything in the same package, as it was before. IMO either way would be better.

packages/sctransaction/transaction.go
```diff
@@ -17,11 +21,14 @@ import (
 // the stateSection and requestSection are parsed from the dataPayload of the value transaction
 type Transaction struct {
 	*valuetransaction.Transaction
-	stateSection   *StateSection
-	requestSection []*RequestSection
-	properties     *Properties // cached properties. If nil, transaction is semantically validated and properties are calculated
+	stateSection     *StateSection
+	requestSection   []*RequestSection
+	cachedProperties coretypes.SCTransactionProperties // cached cachedProperties. If nil, transaction is semantically validated and cachedProperties are calculated
```
dessaya: unintended rename? `cached cachedProperties`

packages/sctransaction/transaction.go
```diff
@@ -55,17 +62,17 @@ func ParseValueTransaction(vtx *valuetransaction.Transaction) (*Transaction, err
 	return ret, nil
 }
 
-// Properties returns valid properties if sc transaction is semantically correct
-func (tx *Transaction) Properties() (*Properties, error) {
-	if tx.properties != nil {
-		return tx.properties, nil
+// Properties returns valid cachedProperties if sc transaction is semantically correct
```
dessaya: unintended rename?

packages/solo/req.go
```diff
@@ -45,7 +47,7 @@ func NewCallParamsFromDic(scName, funName string, par dict.Dict) *CallParams {
 	return ret
 }
 
-// NewCallParams creates structure which wraps in one object call parameters, used in PostRequest and callViewFull
+// NewCallParams creates structure which wraps in one object call parameters, used in PostRequestSync and callViewFull
```
dessaya, line 50: Was this rename intentional? (this one may be good, I'm pointing it out just in case :sweat_smile: )

BugFreeSoftware:
I only see a MintedSupply() int64. How can I determine the color of the minted supply?
Am I missing something?
In the interface for ScBalances in Rust you can already call minted(), which returns the minted color.
In that case you can simply ask the balance for that color.
let minted_color = ctx.balances().minted();
let minted_amount = ctx.balances().balance(&minted_color);

That functionality does not have a counterpart on the host side any more. It used to check if there was a balance for color equal to tx hash and return that color if it was present, and otherwise it returns ScColor::MINT to indicate no minted tokens were found

packages/apilib/request.go
lunfardo314: no its not :)

packages/coretypes/sandbox.go
lunfardo314: it wasn't

packages/sctransaction/properties/properties.go
lunfardo314: hm, I don't disagree with you. There 2 such places now I think. I wanted to decouple interface of properties from implementation. Because the transaction structure will change significantly in the future, I want to sing about semantical properties separately from transaction.
 Let me think a bit more, we could discuss it separately

lunfardo314:
> I only see a MintedSupply() int64. How can I determine the color of the minted supply?
> Am I missing something?
> In the interface for ScBalances in Rust you can already call minted(), which returns the minted color.
> In that case you can simply ask the balance for that color.
> let minted_color = ctx.balances().minted();
> let minted_amount = ctx.balances().balance(&minted_color);
> 
> That functionality does not have a counterpart on the host side any more. It used to check if there was a balance for color equal to tx hash and return that color if it was present, and otherwise it returns ScColor::MINT to indicate no minted tokens were found

No this functionality wasn't present in the current version of sandbox. 
It has nothing to with the transfer, the transfer cannot contain Color_New (MINT). So, it means your current implementation is useless, i.e. not needed.
The `MintedSupply()` return new minted supply **not to the chain account** (it is not possible to mint new tokens to the chain because of many reasons. Such transaction are semantically incorrect and filtered out well before it reaches  the VM (at least it must be like that).
The `MintedSupply()` is a proof of how many tokens were minted to outside. This proof of new supply should be used in the `TokenRegistry`, there's no other way. 
The color of the minted supply (i.e. hash/ID of the request transaction) can be taken from the `RequestID()`

dessaya (approved):

PR #109, opened 2021-02-21
Changed usage of *hashing.HashValue to using hashing.HashValue

BugFreeSoftware:


packages/vm/processors/processors.go
```diff
@@ -67,8 +67,8 @@ func (cps *ProcessorCache) newProcessor(programHash hashing.HashValue, programCo
 	return nil
 }
 
-func (cps *ProcessorCache) ExistsProcessor(h *hashing.HashValue) bool {
-	_, ok := cps.processors[*h]
+func (cps *ProcessorCache) ExistsProcessor(h hashing.HashValue) bool {
```
lunfardo314, line 70: note that return HashValue saves GC effort, that is why we are refactoring it. 
However the passing HashValue as a parameter to immutable object instead of pointer wastes processor resource. Pointer is 8 bytes while value is 32 bytes.   
This applies to other changes too

lunfardo314 (approved):

PR #111, opened 2021-02-22
Implemented MustGet extensions to solo.

brunoamancio:
- Implemented MustGet extensions to solo. Sample in helloworld_test.go. This provides the usage:
`env.MustGetString(resultDict[Param_key])`

- Added IsEmpty() to hash.go (minor change).


contracts/rust/helloworld/test/helloworld_test.go
```diff
@@ -36,7 +36,7 @@ func TestViewGetHelloWorld(t *testing.T) {
 		ScName, ViewGetHelloWorld,
 	)
 	require.NoError(t, err)
-	hw, _, err := codec.DecodeString(res[VarHelloWorld])
+	hw := chain.Env.MustGetString(res[VarHelloWorld])
```
lunfardo314, line 39: no need for this. We have `kvdecoder` package for it, if you need.
``` 
d := kvdecoder.New(res)
s := d.MustGetString(VarHelloWorld)
// or s, err := kvdecoder.GetString(VarHelloWorld)
```

packages/hashing/hash.go
```diff
@@ -154,3 +156,8 @@ func (h *HashValue) Read(r io.Reader) error {
 	}
 	return nil
 }
+
+// IsEmpty checks if HashValue is equals to an empty HashValue struct.
+func (h HashValue) IsEmpty() bool {
```
lunfardo314, line 161: why `h.IsEmpty()` is better than `h == hashing.NilHash` ?
Golang ideology is **minimalism**, very different from Java, C#, Rust and the likes

Do not think we need it tbh

packages/kv/codec/hashvalue.go
```diff
@@ -13,5 +13,5 @@ func DecodeHashValue(b []byte) (hashing.HashValue, bool, error) {
 }
 
 func EncodeHashValue(value hashing.HashValue) []byte {
-	return value[:]
+	return value.Bytes()
```
lunfardo314, line 16: `Bytes` was implemented in previous versions but not used much. Time to deprecate. 
The reason: why use `value.Bytes()` is better than  `value[:]` ?

lunfardo314 (changes requested):

packages/solo/results.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	CouldNotConvertDataInto = "Could not convert data into "
+	DataDoesNotExist        = "Data does not exist."
+)
+
+// MustGetInt64 converts input data into int64. Panics when either no data is provided or cannot be converted.
+func (env *Solo) MustGetInt64(data []byte) int64 {
```
lunfardo314, line 18: Not sure we need this package, seems redundant.
Anyway, use is `kvdecoder` for results (`dict.Dict` type) would make more sense, even in this implementation

contracts/rust/helloworld/test/helloworld_test.go
brunoamancio: In the context of writing unit tests, it is useful to have unit tests written as cleanly as possible. Nobody wants to write boilerplate code in every separate unit test.

packages/hashing/hash.go
brunoamancio: I agree it is minimal. The issue I'm trying to solve is the need to search for `hashing.NilHash`. People writing unit tests will most likely instantiate lots of empty structs like `hashing.HashValue{}` and not use hashing.NilHash. is it is easy to check for empty in an extension, people will use it.

packages/kv/codec/hashvalue.go
brunoamancio: Because that is duplicated code. `hashing.HashValue` already has that https://github.com/iotaledger/wasp/blob/master/packages/hashing/hash.go#L37. There is no need to have that twice. I would remove one of the duplicates and this was chosen.

packages/solo/results.go
brunoamancio: Same argument as before: In the context of writing unit tests, it is useful to have unit tests written as cleanly as possible. Nobody wants to write boilerplate code in every separate unit test.

brunoamancio:
The MustGet extensions are a way to keep unit tests cleaner. Every extra boilerplate line makes code less readable and takes longer to maintain existing tests. They also ensure that, in the context of Solo, variables are loaded, or panic. Of course, we could do without them, but that generates more need for boilerplate utils classes and functions in SC projects with tests using solo.

packages/kv/codec/hashvalue.go
brunoamancio: I will insist on this change, please let me know if I should really remove it.

contracts/rust/helloworld/test/helloworld_test.go
dessaya: If I understand correctly, `MustGetString` will panic if `VarHelloWorld` is not present in `res`. This changes the behavior of the original code, which allows that with `_`, getting an empty string in `hw`. I don't know If the behavior change is intended or even right. Maybe it is in this particular case, but in some cases it will not. The `codec.Decode*` functions are a bit awkward, I know, but in turn they are versatile. One problem with these new `MustGet*` IMO is that they are useful only in one of many possible scenarios: you want to get a value that is required to be present. That is not always the case.

Also, I agree with @lunfardo314: we already have `kvdecoder` for a wrapper around `codec`. You could argue that it is not practical and the proposed solution is better. That may well be true, but in that case it would be nice to see some kind or proof. I mean, using hard numbers. The easiest one is lines of code. In this commit we have +245, -8 LOC. Barring the behavior change, in this specific use case the refactor results in +1, -1 LOC (ie, no gains). A refactor that adds ~200 LOC is only good if it allows to remove at least ~200 LOC elsewhere. This refactor in particular maybe allows to simplify code in several tests; if that is the case it would be nice to see that in the same PR.

Thanks!

brunoamancio: `This changes the behavior of the original code`, there was no original code in the context of solo, no behavior has been changed. The context in which Solo runs is unit testing and it should provide the necessary tools to write tests. 
The extension is expected to be used with Solo and the usage is exactly what you have said, it will panic if the expected result missing. In a unit test, `solo.MustGetString` means that the string must exist, or the unit test must fail because the expected string is not present. 

> One problem with these new MustGet* IMO is that they are useful only in one of many possible scenarios: you want to get a value that is required to be present. That is not always the case.

When unit testing a SCfunction, one must know what the expected response is to make the necessary assertions. In this case, the presence of a string is required. A "solo.GetString" would not require the presence of that string, and is also a usecase of unit tests, although it has not yet been implemented.

> A refactor that adds ~200 LOC is only good if it allows to remove at least ~200 LOC elsewhere. This refactor in particular maybe allows to simplify code in several tests; 

This is not a refactoring. This is an addition of functionality which makes writing unit tests more developer-friendly, imho, and is therefore fine to add lines. Keep in mind the 200 new lines include unit tests.

> if that is the case it would be nice to see that in the same PR.

It would be nice indeed. It just takes a bit of time. I can do that later. For example, in this file alone 15 lines would be gone: [micropay_test.go](https://github.com/iotaledger/wasp/blob/a4ab1a4e9be8e8bf26ac4f0dee25f998bdd85ce7/contracts/examples_core/micropay/micropay_test.go)

brunoamancio: I have refactored code to use the new extension. Since lines of code seem to matter, even though the changes are related to unit testing: I have added a few empty lines which are supposed to make tests more readable. Also, integrated a few functions which were extracted into helper functions, so the unit tests are easier to read without the need of those helper functions. With those measures taken out, the amount of lines have decreased.

brunoamancio:
As suggested by @dessaya, I have refactored code where the usage of the new functions is helpful.

dessaya:
That's better @brunoamancio ! This actually shows that the test functions can be simplified.

But I still think that `packages/solo/results.go` is not necessary, because `kvdecoder` solves the same problem, and is more generic.

Eg, instead of: `sup := chain.Env.MustGetInt64(res[ParamSupply])`

You can do: `sup := kvdecoder.New(res).MustGetInt64(ParamSupply)`

Actually right now this is not possible because the `kvdecoder` methods receive a pointer, so you are forced to do `d := kvdecoder.New(...)` and then `d.MustGetInt64(...)`. A simple solution would be to change the `kvdecoder` methods to receive a value instead of a pointer.

Thanks!

packages/hashing/hash.go
dessaya, line 161: Suggest renaming to `IsZero` or `IsNil`.

I think `IsZero` is better, athough `IsNil` is more consistent with `NilHash`.

brunoamancio: I agree. I'm going to rename it.

brunoamancio:
> That's better @brunoamancio ! This actually shows that the test functions can be simplified.
> 
> But I still think that `packages/solo/results.go` is not necessary, because `kvdecoder` solves the same problem, and is more generic.
> 
> Eg, instead of: `sup := chain.Env.MustGetInt64(res[ParamSupply])`
> 
> You can do: `sup := kvdecoder.New(res).MustGetInt64(ParamSupply)`
> 
> Actually right now this is not possible because the `kvdecoder` methods receive a pointer, so you are forced to do `d := kvdecoder.New(...)` and then `d.MustGetInt64(...)`. A simple solution would be to change the `kvdecoder` methods to receive a value instead of a pointer.
> 
> Thanks!

It does not solve the same problem, but it has similarities. `kvdecoder` does not care about the testing context, while these new functions do. Each unit test has a `t` which is passed into solo, so it can require and fail tests when necessary. So, for each unit test there must be an explicit call to the decoder if we have something like `kvdecoder.New(env, res).MustGetInt64(ParamSupply)` which is cumbersome and generates boilerplate code. Solo (or ´t´) is required because it cares about the test context and has `t` in it. 

The sample you provided would generate unit tests being written like this:

```
func Test_of_mine_1(t *testing.T) {
	const string contractName = "contractName"
	const string contractWasmFilePath = "filepath.wasm"
	const string functionName = "my_function"

	env := solo.New(t, true, true)
	const string chainName = "myChain"
	chain := env.NewChain(nil, chainName)

	err := chain.DeployWasmContract(nil, contractName, contractWasmFilePath)
	require.NoError(t, err)

	req := solo.NewCallParams(contractName, functionName)

	responseBytes, err = chain.PostRequest(req, nil)
	require.NoError(t, err)

	// Important for us
	responseInt64 := kvdecoder.New(env, responseBytes).MustGetInt64(ParamSupply) // we do not want to have the env argument (nor t) here, since the kvdecoder is used in other contexts.
	const int64 expectedResponseInt64 = int64(42)
	require.Equals(t, expectedResponseInt64, responseInt64)
}
```

I see 2 options to improve that. The first is as I've implemented it, directly under solo, so we have:
```
       env := solo.New(t, true, true)
       // ...
       responseInt64 := env.MustGetInt64(responseBytes[ParamSupply])
       const int64 expectedResponseInt64 = int64(42)
       require.Equals(t, expectedResponseInt64, responseInt64)
```
or  move the logic I've implemented, which cares about what the codec cares about PLUS the test context, to a substructure of solo.
```
       env := solo.New(t, true, true)
       // ...
       responseInt64 := env.responseHandler.MustGetInt64(responseBytes[ParamSupply]) // "responseHandler" can be changed.
       const int64 expectedResponseInt64 = int64(42)
       require.Equals(t, expectedResponseInt64, responseInt64)
```

It is important to keep in mind the difference between what each of the following contexts cares about:
- **kvdecoder**: MustGet extensions panic on error (we do not want to panic, we want to fail tests cleanly)
- **codec**: Decode functions do not panic, but rather return a value, an `exists` flag, and an error. <------ This is what we need to fail tests cleanly, as I have implemented it.
- **solo**: MustGet extensions use codec to determine if data could be decoded. When it is not decoded, it fails the test cleanly.

dessaya:
> we do not want to panic, we want to fail tests cleanly

Don't agree here. Panicking is a perfectly good way to fail tests cleanly. That's one of the reasons `kvdecoder` is more generic: you can use it in both normal code and in test code.

brunoamancio:
I understand your disagreement. Although, I prefer not to repeat boilerplate in every unit test I write, and I like writing them. So I'll close the PR, write a wrapper around solo, and extend it as I feels better to me. Thanks for the time reviewing it, though.

PR #112, opened 2021-02-23
Implemented MintedSupply and RequestId

BugFreeSoftware:
Also updated repo with latest Wasm


packages/coretypes/requestid.go
```diff
@@ -45,7 +45,7 @@ func NewRequestIDFromBytes(data []byte) (ret RequestID, err error) {
 }
 
 // Bytes returns requestID as a byte slice
-func (rid *RequestID) Bytes() []byte {
+func (rid RequestID) Bytes() []byte {
```
lunfardo314: don't like it. It means it returns slice of allocated (in heap)  new copy if request id. For the immutable datatype it is the best to have Actually, I am thinking about deprecating all `Bytes()` methods for data types based on byte arrays

packages/vm/sandbox/sandbox.go
```diff
@@ -75,6 +75,7 @@ func (s *sandbox) RequestID() coretypes.RequestID {
 	return s.vmctx.RequestID()
 }
 
+// note: MintedColor() is RequestID().Bytes()[:32]
```
lunfardo314: to be precise,  minted color is equal to `balance.Color(RequestID().TransactionID())`

lunfardo314 (changes requested):
Only one problem, not even related to the MintedSupply

packages/vm/sandbox/sandbox.go
BugFreeSoftware: fixed comment to reflect this

packages/coretypes/requestid.go
BugFreeSoftware: I did that to get it in line with the other coretypes Bytes() functions, because I noticed I could not simply chain 2 calls.

lunfardo314: after discussion with Diego removed all `Bytes()`. Sorry. Please use [:] for hash types. Can you pull latest `develop`? Because now it is a conflict

lunfardo314 (approved):
NVM, I'll resolve th conflict

PR #130, opened 2021-03-12
Ipfs

Juliusan:
i've implemented a downloader, which supports downloading from http, https and from a configurable ipfs http gateway.


packages/downloader/downloader.go
```diff
@@ -0,0 +1,135 @@
...
+
+		return nil
+	}
+}
+
+func (this *Downloader) contains(uri string) bool {
+	var ok bool
+	this.downloadsMutex.Lock()
+	_, ok = this.downloads[uri]
+	this.downloadsMutex.Unlock()
```
lunfardo314: more idiomatic pattern is:
``` go
mutex.Lock()
defer mutex.Unlock()

// the rest
```

packages/downloader/downloader.go
```diff
@@ -0,0 +1,135 @@
...
+	return ok
+}
+
+func (this *Downloader) markStarted(uri string) {
```
lunfardo314: see above

packages/downloader/downloader.go
```diff
@@ -0,0 +1,135 @@
...
+	this.downloadsMutex.Lock()
+	this.downloads[uri] = true
+	this.downloadsMutex.Unlock()
+}
+
+func (this *Downloader) markCompleted(uri string) {
+	this.downloadsMutex.Lock()
```
lunfardo314: see above

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/iotaledger/wasp/packages/registry"
+	"github.com/iotaledger/wasp/packages/testutil"
+	"github.com/labstack/echo/v4"
+	"github.com/stretchr/testify/require"
+)
+
+const MOCK_SERVER_PORT string = ":9999"
+const FILE_CID string = "someunrealistichash12345"
+
+var FILE []byte = []byte{'s', 'o', 'm', 'e', ' ', 'f', 'i', 'l', 'e', ' ', 'f', 'o', 'r', ' ', 't', 'e', 's', 't', 'i', 'n', 'g'} // should be constant, but...
```
lunfardo314: ?

Juliusan: Just some constants for tests. I don't know what's unclear here :-) The comment is, because I wasn't able to define byte array typed slice constant. I had to use `var` instead of `const`. See https://stackoverflow.com/questions/13137463/declare-a-constant-array.

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/wasp/packages/coretypes"
+	"github.com/iotaledger/wasp/packages/hashing"
+)
+
+type Downloader struct {
+	log            *logger.Logger
+	ipfsGateway    string
+	downloads      map[string]bool // Just a HashSet. The value of the element is not important. The existence of key in the map is what counts.
+	downloadsMutex *sync.Mutex
```
dessaya: No need to make it a pointer, it can be just `sync.Mutex`; its zero value is already a properly initialized instance.

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Accepted URIs:
+//  * http://<url of the contents>
+//          e.g. http://some.place.lt/some/contents.txt
+//  * https://<url of the contents>
+//          e.g. https://some.place.lt/some/contents.txt
+//  * ipfs://<cid of the contents>
+//          e.g. ipfs://QmeyMc1i9KLqqyqYCksDZiwntxwuiz5Z1hbLBrHvAXyjMZ
+func (this *Downloader) DownloadAndStore(hash hashing.HashValue, uri string, cache coretypes.BlobCache) error {
```
dessaya: `this` is not idiomatic Go. Suggest renaming to `d`.

https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/wasp/packages/coretypes"
+	"github.com/iotaledger/wasp/packages/hashing"
+)
+
+type Downloader struct {
+	log            *logger.Logger
+	ipfsGateway    string
+	downloads      map[string]bool // Just a HashSet. The value of the element is not important. The existence of key in the map is what counts.
```
dessaya: The comment could be just a GoDoc compliant sentence:

```
// downloads is the set of urls currently being downloaded
downloads map[string]bool
```

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+	return defaultDownloader
+}
+
+// Accepted URIs:
+//  * http://<url of the contents>
+//          e.g. http://some.place.lt/some/contents.txt
+//  * https://<url of the contents>
+//          e.g. https://some.place.lt/some/contents.txt
+//  * ipfs://<cid of the contents>
+//          e.g. ipfs://QmeyMc1i9KLqqyqYCksDZiwntxwuiz5Z1hbLBrHvAXyjMZ
```
dessaya: This maybe should also be in GoDoc format.

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+func (this *Downloader) DownloadAndStore(hash hashing.HashValue, uri string, cache coretypes.BlobCache) error {
+	if this.contains(uri) {
+		this.log.Warnf("File %s is already being downloaded. Skipping it.", uri)
+		return nil
+	} else {
```
dessaya: No need for `else`: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+		this.markStarted(uri)
```
dessaya: I think there is a race condition: another thread could have started the download between checking `this.contains()` and here.

Options to fix this:
* Add a `loadOrStore()` method that does both things while keeping the mutex lock.
* Replace the map+mutex with a [sync.Map](https://golang.org/pkg/sync/#Map) and use its `LoadOrStore` method.
* Replace the map+mutex+multiple goroutines with a channel and a single goroutine with a main loop.

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+		go func() {
```
dessaya: This function is maybe too long, can some part of it be extracted into smaller functions?

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+			case "ipfs":
+				download, err = this.DonwloadFromHttp(this.ipfsGateway + "/ipfs/" + path)
+			case "http":
+				download, err = this.DonwloadFromHttp(uri)
+			case "https":
+				download, err = this.DonwloadFromHttp(uri)
+			default:
+				this.log.Errorf("Unknown protocol %s of uri %s.", protocol, uri)
+				return
+			}
```
dessaya: For example this block may be a `download(uri string) ([]byte, error)` function

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+
+		}()
+
+		return nil
+	}
+}
+
+func (this *Downloader) contains(uri string) bool {
+	var ok bool
```
dessaya: You can replace

```
var ok bool
_, ok = this.downloads[uri]
```

with

```
_, ok := this.downloads[uri]
```

packages/downloader/downloader.go
```diff
@@ -0,0 +1,138 @@
...
+	this.downloadsMutex.Lock()
+	defer this.downloadsMutex.Unlock()
+
+	delete(this.downloads, uri)
+}
+
+func (this *Downloader) DonwloadFromHttp(url string) ([]byte, error) {
+	var response *http.Response
+	var err error
+	response, err = http.Get(url)
```
dessaya: Same here: 

```
response, err := http.Get(url)
```

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/iotaledger/wasp/packages/dbprovider"
+	"github.com/iotaledger/wasp/packages/hashing"
+	"github.com/iotaledger/wasp/packages/registry"
+	"github.com/iotaledger/wasp/packages/testutil"
+	"github.com/labstack/echo/v4"
+	"github.com/stretchr/testify/require"
+)
+
+const MOCK_SERVER_PORT string = ":9999"
+const FILE_CID string = "someunrealistichash12345"
```
dessaya: ALL_CAPS is not idiomatic, even for constants. Use `mockServerPort`, `fileCID`, etc.

https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

packages/downloader/downloader_test.go
dessaya:
```
var file = []byte("some file for testing")
```

Also, no need for the comment

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+var FILE []byte = []byte{'s', 'o', 'm', 'e', ' ', 'f', 'i', 'l', 'e', ' ', 'f', 'o', 'r', ' ', 't', 'e', 's', 't', 'i', 'n', 'g'} // should be constant, but...
+
+func startMockServer() *echo.Echo {
+	var e *echo.Echo
+	e = echo.New()
+	e.GET("/ipfs/:cid", func(c echo.Context) error {
+		var cid string
+		var response []byte
+		cid = c.Param("cid")
```
dessaya: cid := c.Param("cid")

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+		switch cid {
+		case FILE_CID:
+			response = FILE
+		default:
+			return c.NoContent(http.StatusNotFound)
+		}
+		return c.Blob(http.StatusOK, "text/plain", response)
+	})
+	go func() {
+		var err error = e.Start(MOCK_SERVER_PORT)
```
dessaya: `err := e.Start(MOCK_SERVER_PORT)`

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+const MOCK_SERVER_PORT string = ":9999"
+const FILE_CID string = "someunrealistichash12345"
+
+var FILE []byte = []byte{'s', 'o', 'm', 'e', ' ', 'f', 'i', 'l', 'e', ' ', 'f', 'o', 'r', ' ', 't', 'e', 's', 't', 'i', 'n', 'g'} // should be constant, but...
+
+func startMockServer() *echo.Echo {
+	var e *echo.Echo
+	e = echo.New()
```
dessaya: `e := echo.New()`

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}()
+	return e
+}
+
+func stopMockServer(e *echo.Echo) {
+	var ctx context.Context
+	var cancel context.CancelFunc
+	var err error
+
+	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
```
dessaya:
```
ctx, cancel := ...
err := ...
```

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+	var downloader *Downloader = New(log, "http://localhost"+MOCK_SERVER_PORT)
+	var server *echo.Echo = startMockServer()
+	time.Sleep(100 * time.Millisecond) // Time to wait for server start
+
+	var hash hashing.HashValue = hashing.HashData(FILE)
+
+	var db *dbprovider.DBProvider = dbprovider.NewInMemoryDBProvider(log)
+	var reg *registry.Impl = registry.NewRegistry(nil, log, db)
+	var result bool
+	var err error
```
dessaya: Same thing, replace with `:=` :)

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+	require.NoError(t, err)
+	require.False(t, result, "The file should not be part of the registry before the download")
+	err = downloader.DownloadAndStore(hash, "ipfs://"+FILE_CID, reg)
+	require.NoError(t, err)
+	time.Sleep(100 * time.Millisecond) // Time to wait for download completion
+	result, err = reg.HasBlob(hash)
+	require.True(t, result, "The file must be part of the registry after the download")
+	require.NoError(t, err)
+
+	stopMockServer(server)
```
dessaya: Use defer for this:

```
server := startMockServer()
defer stopMockServer()
```

packages/downloader/downloader_test.go
```diff
@@ -0,0 +1,85 @@
...
+	var reg *registry.Impl = registry.NewRegistry(nil, log, db)
+	var result bool
+	var err error
+
+	result, err = reg.HasBlob(hash)
+	require.NoError(t, err)
+	require.False(t, result, "The file should not be part of the registry before the download")
+	err = downloader.DownloadAndStore(hash, "ipfs://"+FILE_CID, reg)
+	require.NoError(t, err)
+	time.Sleep(100 * time.Millisecond) // Time to wait for download completion
```
dessaya: Suggest replacing `time.Sleep` with events, so that the test can finish instantly.

Also, the other `time.Sleep` above should be possible to remove by starting echo with a custom listener, as discussed.

(This can be done in another PR later).

plugins/downloader/plugin.go
```diff
@@ -0,0 +1,25 @@
+package downloader
+
+import (
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/hive.go/node"
+	"github.com/iotaledger/wasp/packages/downloader"
+	"github.com/iotaledger/wasp/packages/parameters"
+)
+
+// PluginName is the name of the web API plugin.
```
dessaya: ~web API~ downloader

plugins/downloader/plugin.go
```diff
@@ -0,0 +1,25 @@
...
+
+func Init() *node.Plugin {
+	var configure, run func(*node.Plugin)
+	configure = func(*node.Plugin) {
+		var log *logger.Logger = logger.NewLogger(PluginName)
+		downloader.Init(log, parameters.GetString(parameters.IpfsGatewayAddress))
+	}
+	run = func(*node.Plugin) {
+		// Nothing to run here
+	}
```
dessaya:
```
configure := ...
run := ...
log := ...
```

dessaya (changes requested):
Good PR!

Don't be intimidated by the amount of comments; it's mostly just me being pedantic about idiomatic Go. None of my observations are really critical, except maybe the one about the race condition.

packages/downloader/downloader.go
Juliusan: I hope, I understood well, what a godoc format is.

packages/downloader/downloader.go
Juliusan: I don't see advantage in turning a one time used block into separate function.

packages/downloader/downloader.go
Juliusan: Ok, I can do that.

packages/downloader/downloader.go
Juliusan: Is it necessary? I like seeing declared types of the variables... :-(

packages/downloader/downloader_test.go
Juliusan: I used `constMockServerPort`, `constFileCID` for that. So, it can be seen that these names represent constants.

packages/downloader/downloader_test.go
Juliusan: To replace this `timer.Sleep` with events, I'll have to include some mechanism in application code, which will really be used for tests only. Is this a normal practice in Go - program mechanisms for testing right into production code?

packages/downloader/downloader.go
Juliusan: I've implemented the first option.

packages/downloader/downloader_test.go
Juliusan: I've removed this sleep from the test. I'll look into another sleep later.

dessaya (approved):

packages/downloader/downloader_test.go
Juliusan: I've removed the second sleep as well.

PR #132, opened 2021-03-15
Implemented std::ops::BitXor for ScHname to make xor operations seemless. 

brunoamancio:
Add support for `my_schname1 ^ my_schname2`.

### DIFF
--- a/contracts/rust/wasmlib/src/hashtypes.rs
+++ b/contracts/rust/wasmlib/src/hashtypes.rs
@@ -291,6 +291,60 @@ impl MapKey for ScHname {
     }
 }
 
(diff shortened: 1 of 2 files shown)


BugFreeSoftware (changes requested):
For the moment we would like tests to be separate from the library code to facilitate easier porting of the library code to other languages. Not many languages support tests within the same source code.

brunoamancio:
Moved tests to their own test file, as per https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory

BugFreeSoftware (approved):

PR #137, opened 2021-04-02
Solo refactor

Juliusan:
Solo now uses Mempool. Also had to move dummy blob cache to different package to avoid circular dependencies.

### DIFF
--- a/packages/solo/solo.go
+++ b/packages/solo/solo.go
@@ -7,8 +7,11 @@ import (
 	"github.com/iotaledger/goshimmer/packages/ledgerstate"
 	"github.com/iotaledger/goshimmer/packages/ledgerstate/utxodb"
 	"github.com/iotaledger/goshimmer/packages/ledgerstate/utxoutil"
+	"github.com/iotaledger/wasp/packages/chain"
+	"github.com/iotaledger/wasp/packages/chain/mempool"
 	"github.com/iotaledger/wasp/packages/sctransaction"
 	"github.com/iotaledger/wasp/packages/testutil/testlogger"
+	"github.com/iotaledger/wasp/packages/util"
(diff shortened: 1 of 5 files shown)


packages/solo/solo.go
```diff
@@ -99,11 +102,9 @@ type Chain struct {
...
 
 	// related to asynchronous backlog processing
-	runVMMutex   *sync.Mutex
-	reqCounter   atomic.Int32
-	chInRequest  chan coretypes.Request
-	backlog      []coretypes.Request
-	backlogMutex *sync.RWMutex
+	runVMMutex *sync.Mutex
+	reqCounter atomic.Int32
+	backlog    chain.Mempool
```
lunfardo314: let's rename `backlog` to `mempool`

packages/solo/solo.go
```diff
@@ -340,19 +330,13 @@ func (ch *Chain) addToBacklog(r coretypes.Request) {
...
 
 	ret := make([]coretypes.Request, 0)
-	remain := ch.backlog[:0]
-	for _, req := range ch.backlog {
-		if len(ret) >= maxBatch {
-			remain = append(remain, req)
-			continue
-		}
+	ready := ch.backlog.GetReadyList(0)
+	ready = ready[:util.MinInt(len(ready), maxBatch)]
```
lunfardo314: better just use something like 
``` 
m := a
if b < m{
   m = b
}
```
instead of yet another function in utils

lunfardo314 (changes requested):

packages/solo/solo.go
Juliusan: I thought, such function might be useful in other places too to avoid code duplication and shorten it.

PR #139, opened 2021-04-06
statemgr: do not use global db instance

dessaya:


sascha1337 (commented):
people had issues with chain ID references, while trying to develop. 
seems like this PR looks promising, we felt there was something "off" - example:

```json
{"Code":400,"Message":"Invalid contract ID: xxx"}
```
_is this the great bugfix PR ?_

lunfardo314 (approved):

PR #158, opened 2021-05-05
Statemgr refactor

Juliusan:
I've left logging messages, which start with XXX. I feel that they might be useful for further development of State manager, however I'll remove them for the final version (that's what XXX is used for). I'll also try to look through all the logging in State manager, tidy it up and make it more clear.

The state manager diagram is also still missing. I'll update it for current version of the code and upload it later.


Juliusan:
Changed XXX to WWW to avoid missinterpretation :-)

packages/chain/chain.go
```diff
@@ -65,10 +65,11 @@ type Committee interface {
 // TODO temporary wrapper for Committee need replacement for all peers, not only committee.
 //  Must be close to GroupProvider but less functions
 type PeerGroupProvider interface {
+	Lock()
+	Unlock()
```
lunfardo314, line 69: Lock/Unlock methods in interface is not good. The part with group provider will need to be refactored anyways to get rid state managed dependency from Committee

lunfardo314 (approved):

PR #162, opened 2021-05-07
Peering improvements

lunfardo314:
- implemented `peering.PeerDomainProvider` interface used by the state manager. It implements unordered set of nodes (neighbors) which are indentified by their netIDs. 
- implemented `coretypes.PeerNetworkConfigProvider` interface to abstract configuration of peering environment from the rest of code


packages/peering/peering.go
```diff
@@ -111,9 +118,18 @@ type GroupProvider interface {
...
+	PeerCollection
+}
+
+// PeerDomainProvider implements unordered set of peers which can dynamically change
+// All peers in the domain shares same peeringID. Each peer within domain is identified via its netID
+type PeerDomainProvider interface {
+	SendMsgByNetID(netID string, msg *PeerMessage)
+	SendMsgToRandomPeers(upToNumPeers uint16, msg *PeerMessage)
+	SendSimple(netID string, msgType byte, msgData []byte)
+	SendMsgToRandomPeersSimple(upToNumPeers uint16, msgType byte, msgData []byte)
```
Juliusan, line 130: I think this method and some others (`SendMsgToRandomPeers`, `SendSimple`) are general methods, which should be part of some more general interface. They use `SendMsgByNetID` to achieve their goals and their implementation should not depend on implementation of `PeerDomainProvider`. Thus, I would suggest implementing this more general interface only once and making it backed by some implementation of `SendMsgByNetID` and `PeerCollection` methods, which might differ in different cases.

Moreover, this more general interface implementation might include permutation functionality (sending messages to random nodes), which should also be common to all the implementations of `PeerDomainProvider`. This is to avoid code duplication and to lessen the set of functions which must be mocked in tests. Not mocked functions are tested contrary to the ones that are mocked.

I hope, my idea is clear :-)

Juliusan: I was thinking about something like this. Of course this is just a sketch:
```
type PeerDomainProvider interface {
	SendMsgByNetID(netID string, msg *PeerMessage)
	PeerCollection
}
type PeerDomainProviderGeneral struct{
	PeerDomainProvider
}
func (*PeerDomainProviderGeneral) SendMsgToRandomPeers(...){...}
func (*PeerDomainProviderGeneral) SendSimple(...){...}
func (*PeerDomainProviderGeneral) SendMsgToRandomPeersSimple(...){...}
func (*PeerDomainProviderGeneral) ReshufflePeers(...){...}```

PR #168, opened 2021-05-20
Feat: evm gas fees

jorgemmsilva:
Gas fees used by the EVM can now be payed in iotas

implemented the concept of:
 - EVM chain owner
 - GasPerIota 

added the functions:
 - setGasPerIota
 - setOwner
 - withdrawGasFees - which can be called by the evm owner to transfer the fees collected
 
 
 If iotas sent are not enough to cover the gasLimit specified, the transaction won't be accepted


contracts/native/evmchain/testutils.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+
+			gasFee, _, err := codec.DecodeUint64(result.MustGet(FieldGasFee))
+			require.NoError(t, err)
+
+			var receipt *types.Receipt
+			err = rlp.DecodeBytes(result.MustGet(FieldResult), &receipt)
+			// TODO update to new Receipt decoding
+			// receipt, err := DecodeReceipt(result.MustGet(FieldResult))
+			require.NoError(t, err)
```
jorgemmsilva: @dessaya this was breaking, so I reverted to the previous iteration for now

contracts/native/evmchain/testutils.go
```diff
@@ -0,0 +1,156 @@
...
+			return receipt, gasFee, nil
+		}
+
+	}
+
+	return contractAddress, callFn
+}
+
+// helper to reuse code to call the `retrieve` view in the storage contract
+func GetCallRetrieveView(t *testing.T, chain *solo.Chain, contractAddress common.Address, contractABI abi.ABI) func() uint32 {
```
dessaya: probably this should just call the view instead of returning a func?

```go
func callStorageRetrieve(t *testing.T, chain *solo.Chain, contractAddress common.Address) uint32 {
  ...
}
```

`contractABI` in theory is not needed as a parameter since it's always the Storage contract.

packages/solo/solo.go
```diff
@@ -41,7 +41,7 @@ const DefaultTimeStep = 1 * time.Millisecond
 // Saldo is the default amount of tokens returned by the UTXODB faucet
 // which is therefore the amount returned by NewKeyPairWithFunds() and such
 const (
-	Saldo              = uint64(1337)
+	Saldo              = uint64(1000000) // 1Mi
```
dessaya: Maybe this should use the constant defined in goshimmer side?

contracts/native/evmchain/interface.go
```diff
@@ -47,16 +56,31 @@ const (
...
+	FieldAgentId                 = "i"
+	FieldTransactionHash         = "h"
+	FieldTransactionData         = "t"
+	FieldTransactionDataBlobHash = "th"
+	FieldBalance                 = "b"
+	FieldCallArguments           = "c"
+	FieldResult                  = "r"
+	FieldBlockNumber             = "bn"
+	FieldBlockHash               = "bh"
+	FieldEvmOwner                = "evmOwner"
```
dessaya: `FieldEVMOwner` to conform with idiomatic Go naming

dessaya (approved):

contracts/native/evmchain/testutils.go
jorgemmsilva: returning a function was to help with reusability, without having to pass the parameters every time:
```
	retrieve := GetCallRetrieveView(t, chain, contractAddress, contractABI)
	...
	value := retrieve()
	...
	value2 := retrieve()
	...etc
```

in hindsight it doesn't get reused too many times, so its not that useful, also good point about the abi.
will refactor 👍

dessaya (approved):

lunfardo314 (approved):

PR #169, opened 2021-05-24
test new users are able to use the EVM without ether

jorgemmsilva:


contracts/native/evmchain/testutils.go
```diff
@@ -107,6 +107,9 @@ func createCallFnWithGasLimit(t *testing.T, chain *solo.Chain, env *solo.Solo, c
 	senderAddress := crypto.PubkeyToAddress(sender.PublicKey)
 
 	nonce := getNonceFor(t, chain, senderAddress)
+	if nonce == 0 {
+		nonce = 1
+	}
```
dessaya: What is this for?

jorgemmsilva: This was wrong.  Thanks for pointing it out 🤦‍♂️ .
The signer was wrong, it was still using the faucet key, hence it wanted nounce 1 instead of 0.
fixed here: https://github.com/iotaledger/wasp/pull/169/commits/3902cb2788706442c19d0c9816b1f895c9356d82

dessaya (approved):

PR #173, opened 2021-05-27
set faucet pow to 0 in tests

dessaya:


client/goshimmer/goshimmer.go
```diff
@@ -10,23 +10,29 @@ import (
 	"github.com/iotaledger/wasp/packages/transaction"
 )
 
+const DefaultPoWTarget = -1
```
jorgemmsilva: this seems to not be used

lunfardo314 (approved):

PR #177, opened 2021-05-31
refactor: Database access

jorgemmsilva:
There is a singleton `dbmanager` that controls the database access

In general, code accessing the database can have the partition logic abstracted (no need to have access to the entire dbProvider), a simple kvStore works.

the database plugin takes care of the lifecycle of the dbmanager singleton (init, close, GC), and usage - it exposes 3 functions (that use the dbmanager internally):
```go
GetRegistryKVStore() kvstore.KVStore

GetOrCreateKVStore(chainID *coretypes.ChainID) kvstore.KVStore

GetKVStore(chainID *coretypes.ChainID) kvstore.KVStore
```

chainRegistry has a dedicated `registryDB` field on chainrecord, each chain has its own dedicated db instance.


main.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/iotaledger/wasp/plugins/cli"
 	"github.com/iotaledger/wasp/plugins/config"
 	"github.com/iotaledger/wasp/plugins/dashboard"
-	"github.com/iotaledger/wasp/plugins/database"
+	databaseplugin "github.com/iotaledger/wasp/plugins/database"
```
dessaya: Why the package alias?

packages/coretypes/chain_record_test.go
```diff
@@ -0,0 +1,28 @@
+package coretypes
+
+import (
+	"testing"
+
+	"github.com/iotaledger/wasp/packages/registry_pkg/chain_record"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChainRecord(t *testing.T) {
```
dessaya: Shouldn't this test be in the chainrecord package?

packages/dashboard/dashboard_test.go
```diff
@@ -44,7 +44,7 @@ func TestDashboardChainView(t *testing.T) {
 
 	// make sure we are using .Base58()
 	require.NotContains(t, html.Text(), "OutputID {")
-	require.NotContains(t, html.Text(), "Address {")
+	// require.NotContains(t, html.Text(), "Address {")  // TODO check with Diego, not sure of the reason for this check
```
dessaya: Because historically `Address.String()` was the same as `Address.Base58()`. So the dashboard used `Address.String()` everywhere to show addresses.

Recently it was changed so that `Address.String()` returns something like `"Address { ... }"` and it broke the dashboard. This check ensures that we are using `.base58()` instead of `.String()` to show addresses.

Why did you need to comment it? Was it failing?

packages/database/dbmanager/dbmanager.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/iotaledger/hive.go/kvstore"
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/wasp/packages/database/dbprovider"
+	"github.com/iotaledger/wasp/packages/parameters"
+)
+
+type DBManager struct {
+	log         *logger.Logger
+	dbInstances map[[ledgerstate.AddressLength]byte]*dbprovider.DBProvider
+	mutex       *sync.RWMutex
```
dessaya: This could be a `sync.RWMutex`, no need for pointer. The zero value is ready to be used.

packages/database/dbmanager/dbmanager.go
```diff
@@ -0,0 +1,120 @@
...
+	inMemory    bool
+}
+
+var Instance *DBManager
+
+func CreateInstance(logger *logger.Logger, inMemory bool) {
+	Instance = NewDBManager(logger, inMemory)
```
dessaya: Maybe do something like this first?
```go
if Instance != nil {
  panic("CreateInstance called twice")
}

File: packages/database/dbmanager/dbmanager.go [hunk 6]
```diff
@@ -0,0 +1,120 @@
...
+func (m *DBManager) Close() {
+	for _, instance := range m.dbInstances {
+		instance.Close()
+	}
+}
+
+func (m *DBManager) createDedicatedDbInstance(chainID *ledgerstate.AliasAddress) *dbprovider.DBProvider {
+	// create new db instance
+	if m.inMemory {
+		m.log.Infof("IN MEMORY DATABASE")
```
dessaya: Maybe add the chainID to the log message?

packages/database/dbmanager/dbmanager.go
```diff
@@ -0,0 +1,120 @@
...
+		instance = m.createDedicatedDbInstance(chainID)
+	} else {
+		instance = m.GetRegistryDbInstance()
+	}
+
+	m.dbInstances[chainID.Array()] = instance
+	return instance
+}
+
+func (m *DBManager) GetOrCreateDBInstance(chainID *ledgerstate.AliasAddress, dedicatedDbInstance bool) *dbprovider.DBProvider {
```
dessaya: Unless I'm missing the point, IMHO it would be slightly better (less confusing) to split this into 2 methods:

```
GetOrCreateDBInstance(chainID *ledgerstate.AliasAddress) *dbprovider.DBProvider
GetOrCreateRegistryDBInstance() *dbprovider.DBProvider
```

Or:

```
const Registry = &ledgerstate.AliasAddress{}
```
and just a single method:
```
GetOrCreateDBInstance(chainID *ledgerstate.AliasAddress) *dbprovider.DBProvider
```

Then, to create the registry instance: `GetOrCreateDBInstance(Registry)`.

Personally I like this last option better, so that the registry partition is not a special case at all. The same for `GetKVStore...` methods.

packages/database/dbmanager/dbmanager.go
```diff
@@ -0,0 +1,120 @@
...
+		instance = m.createInstance(chainID, dedicatedDbInstance)
+	}
+	return instance
+}
+
+func (m *DBManager) GetDBInstance(chainID *ledgerstate.AliasAddress) *dbprovider.DBProvider {
+	return m.dbInstances[chainID.Array()]
+}
+
+func (m *DBManager) GetRegistryDbInstance() *dbprovider.DBProvider {
```
dessaya: Nit: `Db` -> `DB` for consistency

packages/registry_pkg/blobcache_test.go
```diff
@@ -3,16 +3,16 @@ package registry_pkg
...
 	"github.com/iotaledger/wasp/packages/hashing"
 	"github.com/iotaledger/wasp/packages/testutil/testlogger"
 	"github.com/stretchr/testify/require"
 )
 
 func TestBlobPutGet(t *testing.T) {
 	log := testlogger.NewLogger(t)
-	db := dbprovider.NewInMemoryDBProvider(log)
-	reg := NewRegistry(nil, log, db)
+	dbmanager.CreateInstance(log, true)
```
dessaya: Can this test be refactored so that it does not use a singleton?

packages/registry_pkg/chain_record/chain_record.go
```diff
@@ -0,0 +1,64 @@
+package chain_record
```
dessaya: https://golang.org/doc/effective_go#package-names

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps. Err on the side of brevity, since everyone using your package will be typing that name.

Given that, I would remove the underscore: `chainrecord`. `registry_pkg` could also be named `registry`?

Also, why was `ChainRecord` moved to its own package? Because of cyclic dependencies?

packages/registry_pkg/chain_record/chain_record.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+
+	"github.com/iotaledger/goshimmer/packages/ledgerstate"
+	"github.com/iotaledger/hive.go/marshalutil"
+)
+
+// ChainRecord represents chain the node is participating in
+// TODO optimize, no need for a persistent structure, simple activity tag is enough
+type ChainRecord struct {
+	ChainIdAliasAddress *ledgerstate.AliasAddress
```
dessaya: Why did this change from type `*coretypes.ChainID`? Also, consider `ChainAddress` or just `Address` for less verbose names?

packages/registry_pkg/chain_record/chain_record.go
```diff
@@ -0,0 +1,64 @@
...
+	Active              bool
+	DedicatedDbInstance bool // whether the chain data is stored as a separate db instance/file
```
dessaya: I'm not sure I understand why this attribute has to be here. Could there exist 2 different ChainRecords with the same address and different values for this bool?

main.go
jorgemmsilva: We have a `database` package, so I renamed the plugin to `databaseplugin` to avoid confusion

packages/coretypes/chain_record_test.go
jorgemmsilva: It should, I moved the chainrecord package around multiple times and forgot to move the test file. Thanks, fixing.

packages/dashboard/dashboard_test.go
jorgemmsilva: It failed because I changed the chainrecord to have an alias address, rather than a "coretype.chainID" (to avoid dependency cycles).
Thanks will uncomment the check and make sure to use `.base58()`

packages/database/dbmanager/dbmanager.go
jorgemmsilva: my thought process here was to have a special call for registry (that should always exist), and to abstract the logic of the Registry being an empty address.

From the PoV of the code that is using this API, I think its more intuitive to call:
```go
GetRegistryKVStore()
```
than: 
```go
GetOrCreateDBInstance(nil)
```

Happy to change to a single `GetOrCreateDBInstance`, I initially implemented it like that but figured having a separated registry function would be nice

packages/registry_pkg/chain_record/chain_record.go
jorgemmsilva: thanks, will fix the naming 👍 .
yes, ended up moving to a separate package because of cyclic dependencies

packages/registry_pkg/chain_record/chain_record.go
jorgemmsilva: I kinda like the verbose name, from `ChainIDAliasAddress` I can immediately understand that this is an aliasAddress that represents a chainID 😄 , but I guess the `AliasAddress ` type is enough to represent that. Will refactor 👍

packages/registry_pkg/chain_record/chain_record.go
jorgemmsilva: I think it shouldn't be possible have multiple `ChainRecords` with the same address. (not sure if we enforce that, tho)
This attribute is here because this was the requirements I was given 😄 . From what I understand we want to have chains that are stored in the main instance (same db file as the registry), and others that are stored as standalone instances (separate files)

main.go
dessaya: But... if the package was renamed, then the directory should be renamed as well, so that the alias is not needed here.
Not a big deal since it only affects here, but it is still somewhat confusing IMO.

packages/registry_pkg/chain_record/chain_record.go
dessaya: Ok, then maybe I disagree with the requirement :stuck_out_tongue: 
@lunfardo314 Why have chains stored in different places?

packages/database/dbmanager/dbmanager.go
dessaya: I agree, it's fine to have a separate `GetRegistryKVStore()` for convenience. My issue is if the caller has to provide 2 arguments to create/get a chain partition. But that's probably because I didn't understand the requirements. In my head there is the registry DB (used to store whatever) and one DB for each chain (used to store stuff related to that chain). So, from the outside there should be only `GetRegistryKVStore()` and `GetChainKVStore(chainID)`. Right?

dessaya: Also, I think the trick of using the zero address for the registry DB is probably confusing and no longer needed. The registry DB could be stored in a DB called "registry".

dessaya (approved):

PR #184, opened 2021-06-06
Implemented optimistic reading of the chain's state

lunfardo314:
- the chain state (the database) is updated by the state manager
- other asynchronous parts may be reading the state meanwhile
- the optimistic reading ensures that any party which is reading the state during update will get notified on the updated state via the special error code. It then can retry or abandon its operation


tools/cluster/tests/inccounter_solo_test.go
```diff
@@ -43,6 +45,7 @@ func TestIncSoloRepeatMany(t *testing.T) {
 		WithIotas(1)
 	_, err = chain.PostRequestSync(req, nil)
 	require.NoError(t, err)
+	time.Sleep(1 * time.Second)
```
dessaya, line 48: I see there are lots of calls to `time.Sleep` in tests, which probably make the tests take longer then they could, and also is not infallible.
Could there be a better alternative? For example listening to publisher events. Maybe for another PR, but I think it would pay off to make the effort to remove all time.Sleep() in tests.

packages/chain/mempool/mempool.go
```diff
@@ -46,135 +58,259 @@ func New(stateReader state.StateReader, blobCache coretypes.BlobCache, log *logg
...
+func (m *mempool) removeFromInBuffer(req coretypes.Request) {
+	m.inMutex.Lock()
+	defer m.inMutex.Unlock()
+	if _, ok := m.inBuffer[req.ID()]; ok {
+		delete(m.inBuffer, req.ID())
+		m.outBufCounter++
+	}
+}
+
+// fills up the buffer with requests in the in buffer
```
jorgemmsilva:
```suggestion
// fills up the buffer with requests from the in-buffer
```

packages/coretypes/coreutil/globalindex.go
```diff
@@ -0,0 +1,82 @@
...
+// GlobalSync implements optimistic read baselines and global locks
+type GlobalSync interface {
+	GetSolidIndexBaseline() StateBaseline
+	SetSolidIndex(idx uint32) GlobalSync // for use in state manager
+	InvalidateSolidIndex() GlobalSync    // only for state manager
+	Mutex() *sync.RWMutex
+}
+
+type globalSync struct {
+	solidIndex  atomic.Uint64
```
jorgemmsilva: could be `atomic.Uint32`, since we're using it to store uint32 indexes

packages/solo/run.go
```diff
@@ -20,20 +20,30 @@ import (
...
 	ch.runVMMutex.Lock()
 	defer ch.runVMMutex.Unlock()
 
-	for _, r := range batch {
+	ch.reqCounter.Add(int32(len(reqs)))
+	ch.mempool.ReceiveRequests(reqs...)
+	ch.mempool.WaitInBufferEmpty()
+
+	//ready := ch.mempool.ReadyNow(ch.Env.LogicalTime())
+	//require.EqualValues(ch.Env.T, len(reqs), len(ready))
```
jorgemmsilva: commented code can probably be removed

packages/webapi/state/callview.go
```diff
@@ -44,24 +45,21 @@ func handleCallView(c echo.Context) error {
...
-
-	vctx, err := viewcontext.NewFromDB(database.GetKVStore(theChain.ID()), *theChain.ID(), theChain.Processors())
-	if err != nil {
-		return fmt.Errorf(fmt.Sprintf("Failed to create context: %v", err))
-	}
-
-	ret, err := vctx.CallView(contractHname, coretypes.Hn(fname), params)
+	vctx := viewcontext.NewFromChain(theChain)
+	var ret dict.Dict
+	_ = optimism.RepeatOnceIfUnlucky(func() error {
```
jorgemmsilva, line 59: should we handle the error here in case `CallView` errors twice?
```suggestion
	err = optimism.RepeatOnceIfUnlucky(func() error {
```

jorgemmsilva (approved):

packages/coretypes/coreutil/globalindex.go
lunfardo314: this is on purpose. The higher 4 bytes are used as 1 bit of information to invalidate the state completely for the period when it is updating. After it is updated, new index is set

packages/webapi/state/callview.go
lunfardo314: the error is assigned to the `err` in the closure. So, it is handled

lunfardo314: the return error from `RepeatOnceIfUnlucky` may be useful when we use closure without inline context

packages/coretypes/coreutil/globalindex.go
```diff
@@ -0,0 +1,97 @@
...
+	SetSolidIndex(idx uint32) GlobalSync // for use in state manager
+	InvalidateSolidIndex() GlobalSync    // only for state manager
+	Mutex() *sync.RWMutex
+}
+
+type globalSync struct {
+	// the lower 32 bits stores index.
+	// If the higher 32 bits are not 0, the stateBaseline is invalid until we set 0
+	// into higher 32 bits. The one uint64 variable is used for atomicity
+	solidIndex atomic.Uint64
```
dessaya: Why not use a struct with `atomic.Value` if this object stores more than one logical value?

packages/coretypes/coreutil/globalindex.go
```diff
@@ -0,0 +1,97 @@
...
+// GlobalSync and StateBaseline interfaces implements optimistic (non-blocking) access to the
+// global state (database) of the chain
+
+// region GlobalSync  ////////////////////////////////
+
+type GlobalSync interface {
+	GetSolidIndexBaseline() StateBaseline
+	SetSolidIndex(idx uint32) GlobalSync // for use in state manager
+	InvalidateSolidIndex() GlobalSync    // only for state manager
+	Mutex() *sync.RWMutex
```
dessaya: I think these three functions provide a leaky abstraction, and make it easy for the caller to break things. E.g. the caller may forget to call `InvalidateSolidIndex` before `Mutex().Lock()`.

Would it make sense to have something like this instead?

```go
type GlobalSync interface {
	GetSolidIndexBaseline() StateBaseline
        WithWriteLock(func () uint32)
}
```

Then, instead of:

```go
g.InvalidateSolidIndex()
g.Mutex().Lock()
defer g.Mutex().Unlock()

// do stuff

g.SetSolidIndex(newBlockIndex)
```

the caller would do this:

```go
g.WithWriteLock(func () uint32 {
    // do stuff
    return newBlockIndex
})
```

packages/coretypes/coreutil/globalindex.go
```diff
@@ -0,0 +1,97 @@
...
+func (g *globalSync) Mutex() *sync.RWMutex {
+	return g.globalMutex
+}
+
+// endregion  ///////////////////////////////////////////////////
+
+// region StateBaseline //////////////////////////////////////////////
+
+type StateBaseline interface {
+	Set()
```
dessaya, line 70: Unclear why is `Set()` needed? Is it for when the reader needs to retry? Why not just get a new baseline instead?

Same for all `SetBaseline()` methods: instead of "resetting" a KVReader, just create a new one with the latest baseline.

packages/vm/viewcontext/viewcontext.go
```diff
@@ -56,22 +44,28 @@ func (v *viewcontext) CallView(contractHname coretypes.Hname, epCode coretypes.H
...
 		}()
-		ret, err = v.mustCallView(contractHname, epCode, params)
+		ret, err = v.callView(contractHname, epCode, params)
 	}()
 	return ret, err
 }
 
-func (v *viewcontext) mustCallView(contractHname coretypes.Hname, epCode coretypes.Hname, params dict.Dict) (dict.Dict, error) {
+func (v *viewcontext) callView(contractHname coretypes.Hname, epCode coretypes.Hname, params dict.Dict) (dict.Dict, error) {
+	v.stateReader.SetBaseline()
```
dessaya: Not sure if setting the baseline here is correct? If one view calls another, this will end up setting the baseline in the middle of the outer view call?

tools/cluster/tests/inccounter_solo_test.go
lunfardo314: yes, I know. I don't know easy solution for this. I had to add these sleeps because now when mempool is asynchronous it all needs more times to read from internal request queues, especially in tests with smart contract posting requests cross-chain.
We will discuss it and maybe we indeed have invest effort to find better solution

packages/coretypes/coreutil/globalindex.go
lunfardo314: that is an alternative. yes. Is it less ugly? :) Not sure

packages/coretypes/coreutil/globalindex.go
lunfardo314: I agree. But mutex is not used and probably I will remove it anyway (I had and idea it is needed but it seems it is not)

packages/coretypes/coreutil/globalindex.go
lunfardo314: well, that would be a functional equivalent.  But in mempool you do not get new reader each time, but you are privided a reader when created and later you just set baselines. It can be both ways.

packages/vm/viewcontext/viewcontext.go
lunfardo314: that is a bug. Nice catch!

packages/coretypes/coreutil/globalindex.go
lunfardo314: oh, now I remember. I wanted to avoid dependency of `OptimisticKVStore` reader from global stuff like `GlobalSync`, so I introduced local abstraction `StateBaseline`. The `OptimisticStateReader` of course depends on global stuff.

dessaya: If the problem is mempool, instead of receiveing the reader it could receive a function that provides a new reader: something like `getReader func() OptimisticKVStore`. It would call that function each time it wants to "reset" the baseline.

And `OptimisticKVStore` only needs to receive a predicate function: `isValid func() bool`.

The `GlobalSync` could be something like:

```go
type SyncronizedKVStore interface {
	GetReader() OptimisticKVStore // you call this each time you need a new "baseline"
	WithWriter(f func (kv KVStore) bool) // invalidates all current readers if f returns true
}
```

Would that be a simpler approach? No need to pass around the state index, no need for a baseline object.

PR #192, opened 2021-06-15
Feat: offledger request webapi and gossip

jorgemmsilva:
## webapi endpoint

New endpoint which adds the ability to send an offledger requests to a WASP network.

To issue an offledger request we need to send a http `POST` request to:
```
/request/:chainID
```
where `:chainID` is the base58 representation of the target chanID.

for the body of the request there are 2 options:

- JSON payload containing the (signed) offledger request as a base64 string:
```json
{ "request": "...base64 string..." }
```

- binary representation of the (signed) offledger request

http headers  must be set accordingly: `application/json` or `application/octet-stream`

if successful, the endpoint will return an empty response with: HTTP status: 202 - Accepted.

---

## gossip logic
 - we keep track of requests by using the mempool.
 - when a new request appears the node will check if it is part of the committee for the target chain
 - in case the node is part of the chain committee, it will process the request and share it with all the other committee nodes
 - in case the node is *not* part of the committee, it will forward the request to a pre-defined number of random peers (currently 10)

--- 
## further work
 - should add a clust test using access nodes (currently not easy to create such configuration)
 - stress test (?)


client/chainclient/chainclient.go
```diff
@@ -1,17 +1,22 @@
...
 	"github.com/iotaledger/wasp/packages/coretypes"
 	"github.com/iotaledger/wasp/packages/coretypes/chainid"
+	"github.com/iotaledger/wasp/packages/coretypes/request"
 	"github.com/iotaledger/wasp/packages/coretypes/requestargs"
 	"github.com/iotaledger/wasp/packages/transaction"
+	"github.com/iotaledger/wasp/packages/webapi/routes"
 )
 
-// Client allows to send webapi requests to a specific chain in the node
+// Client allows to send on-ledger or off-ledger requests to a specific chain in the node
```
dessaya:
```suggestion
// Client allows to interact with a specific chain in the node, for example to send on-ledger or off-ledger requests
```

Because using the client you can also call a view, query its record, etc

client/chainclient/chainclient.go
```diff
@@ -61,3 +66,28 @@ func (c *Client) Post1Request(
...
+	if err != nil {
+		return nil, err
+	}
+	httpReq.Header.Add("Content-type", `application/json"`)
+	resp, err := httpclient.Do(httpReq)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	return offledgerReq, nil
```
dessaya: `WaspClient` should provide this as a method. `chain.Client` acts mostly as a dumb wrapper for `WaspClient` (and the goshimmer client).

Also, I think `WaspClient.do()` already does most of this logic.

packages/chain/chainimpl/interface.go
```diff
@@ -103,6 +103,21 @@ func (c *chainObj) ReceiveMessage(msg interface{}) {
 	}
 }
 
+const gossipUpToNPeers = 10
```
dessaya: Should this be configurable?

packages/chain/mempool/mempool.go
```diff
@@ -154,6 +155,16 @@ func (m *mempool) ReceiveRequests(reqs ...coretypes.Request) {
 	}
 }
 
+// ReceiveRequest used to receive off-ledger request
+func (m *mempool) ReceiveRequest(req coretypes.Request) bool {
+	// could be worth it to check if the request was already processed in the blocklog.
+	// Not adding this check now to avoid overhead, but should be looked into in case re-gossiping happens a lot
+	if _, exists := m.inBuffer[req.ID()]; exists {
```
dessaya, line 163: Doesn't this need to lock the `inMutex`?

packages/chain/peermsg.go
```diff
@@ -19,9 +19,10 @@ import (
 
 // Message types for the committee communications.
 const (
-	MsgGetBlock     = 1 + peering.FirstUserMsgCode
-	MsgBlock        = 2 + peering.FirstUserMsgCode
-	MsgSignedResult = 3 + peering.FirstUserMsgCode
+	MsgGetBlock         = 1 + peering.FirstUserMsgCode
+	MsgBlock            = 2 + peering.FirstUserMsgCode
+	MsgSignedResult     = 3 + peering.FirstUserMsgCode
+	MsgOffLedgerRequest = 4 + peering.FirstUserMsgCode
```
dessaya:
```suggestion
	MsgGetBlock         = 1 + peering.FirstUserMsgCode + iota
	MsgBlock
	MsgSignedResult
	MsgOffLedgerRequest
```

packages/coretypes/request/requestimpl.go
```diff
@@ -473,4 +474,16 @@ func (req *RequestOffLedger) Tokens() *ledgerstate.ColoredBalances {
...
+}
+
+func NewRequestOffLedgerFromBase64(str string) (*RequestOffLedger, error) {
+	data, err := base64.StdEncoding.DecodeString(str)
+	if err != nil {
+		return nil, err
+	}
+	return NewRequestOffLedgerFromBytes(data)
+}
+
```
dessaya: If the purpose of this is solely for json encoding/decoding, it should not be necessary since a `[]byte` is automatically encoded as a base64 string by the `json` package.

You could also create a new object in `webapi/model` if necessary.

packages/webapi/request/request.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (o *offLedgerReqAPI) handleNewRequest(c echo.Context) error {
+	chainID, offLedgerReq, err := parseParams(c)
+	if err != nil {
+		return err
+	}
+
+	ch := o.getChain(chainID)
+	if ch == nil {
+		return xerrors.Errorf("Unknown chain: %s", chainID.Base58())
```
dessaya: This should probably be an `httperrors.NotFound` or `BadRequest`. Otherwise, it returns status 500 by default.

packages/webapi/request/request.go
```diff
@@ -0,0 +1,68 @@
...
+
+type getChainFn func(chainID *chainid.ChainID) chain.ChainCore
+
+func AddEndpoints(server echoswagger.ApiRouter, getChain getChainFn) {
+	instance := &offLedgerReqAPI{
+		getChain: getChain,
+	}
+	server.POST(routes.NewRequest(":chainID"), instance.handleNewRequest).
+		SetSummary("New off-ledger request").
+		AddParamPath("", "request", "request in base64").
```
dessaya: Is the parameter in the path not a base58 chainID?

packages/chain/chainimpl/interface.go
jorgemmsilva: for sure, I'll add a param for this 👍

packages/chain/mempool/mempool.go
lunfardo314: yes it does need to lock `inBuffer`

jorgemmsilva: thanks, fixing

packages/webapi/request/request.go
jorgemmsilva: it is, fixing this - thanks

client/chainclient/chainclient.go
```diff
@@ -61,3 +62,18 @@ func (c *Client) Post1Request(
...
+	entrypoint coretypes.Hname,
+	args requestargs.RequestArgs,
+) (*request.RequestOffLedger, error) {
+	offledgerReq := request.NewRequestOffLedger(contractHname, entrypoint, args)
+	offledgerReq.Sign(c.KeyPair)
+	err := c.WaspClient.PostOffLedgerRequest(&c.ChainID, offledgerReq)
+	if err != nil {
+		return nil, err
+	}
+	return offledgerReq, nil
```
dessaya:
```suggestion
	return c.WaspClient.PostOffLedgerRequest(&c.ChainID, offledgerReq)
```

client/offledger.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/iotaledger/wasp/packages/webapi/routes"
+)
+
+func (c *WaspClient) PostOffLedgerRequest(chainID *chainid.ChainID, req *request.RequestOffLedger) error {
+	data := model.OffLedgerRequestBody{Request: req.Bytes()}
+	err := c.do("POST", routes.NewRequest(chainID.Base58()), data, nil)
+	if err != nil {
+		return err
+	}
+	return nil
```
dessaya:
```suggestion
	return c.do("POST", routes.NewRequest(chainID.Base58()), data, nil)
```

dessaya (approved):

PR #194, opened 2021-06-16
Wasmvm

BugFreeSoftware:


packages/kv/codec/encodego.go
```diff
@@ -12,22 +12,22 @@ import (
 
 func Encode(v interface{}) []byte {
```
lunfardo314, line 13: Here and below. We have to think twice if we really want to support full range of integer types. The problem is, you don’t really know what SC expects. Remember, we do not provide type information with data (and we do not intend to). It is not a high level programming environment. We only pass binary data in core contracts, which is part if the VM but can be called from outside.
If SC expects `uint32` and you encode `uint16(42)` in the call, it will panic, while semantically it is perfectly ok. So, with full spectrum of integer types like in the PR we will create problems without solving any.

lunfardo314 (approved):

PR #200, opened 2021-06-18
Access node and committee rotation

lunfardo314:
- implemented and tested access node functionality, i.e. when nodes don't have keys and do not participate in consensus but always keep synced state
- implemented and tested committee rotation function, i.e. when validators of the chain are moved to another state controller address
- some smaller improvements


client/chainclient/chainclient.go
```diff
@@ -76,3 +78,80 @@ func (c *Client) PostOffLedgerRequest(
...
+	}
+	par.Transfer = ledgerstate.NewColoredBalances(ret)
+	return par
+}
+
+func (par *PostRequestParams) WithIotas(i uint64) *PostRequestParams {
+	return par.WithTransferEncoded(ledgerstate.ColorIOTA, i)
+}
+
+func encodeIntToUint54(i interface{}) uint64 {
```
jorgemmsilva: uint54? I think we have enough integer types already 😨

client/chainclient/chainclient.go
```diff
@@ -76,3 +78,80 @@ func (c *Client) PostOffLedgerRequest(
...
+func (par *PostRequestParams) WithTransferEncoded(colval ...interface{}) *PostRequestParams {
+	ret := make(map[ledgerstate.Color]uint64)
+	if len(colval) == 0 {
+		return par
+	}
+	if len(colval)%2 != 0 {
+		panic("WithTransferEncode: len(params) % 2 != 0")
+	}
+	for i := 0; i < len(colval)/2; i++ {
+		key, ok := colval[2*i].(ledgerstate.Color)
```
jorgemmsilva, line 103: why do we have `len(colval)/2` and then `colval[2*i]`? 
is the usage like this?
```
par.WithTransferEncoded(color1,color2,color3, 100, 200, 300)
```
?
isn't it easier to just provide a map of colors to balances?

client/chainclient/chainclient.go
lunfardo314: Hahaha

client/chainclient/chainclient.go
lunfardo314: It is `color1, value1, color2, value2, …`. This is a pattern used for example by `Debugw`

client/chainclient/chainclient.go
```diff
@@ -76,3 +78,80 @@ func (c *Client) PostOffLedgerRequest(
 	offledgerReq.Sign(c.KeyPair)
 	return offledgerReq, c.WaspClient.PostOffLedgerRequest(&c.ChainID, offledgerReq)
 }
+
+// NewPostRequestParams simplifies encoding of request parameters
+func NewPostRequestParams(p ...interface{}) *PostRequestParams {
+	return &PostRequestParams{
+		Args: requestargs.New(nil).AddEncodeSimpleMany(parseParams(p)),
+	}
+}
```
dessaya, line 87: Propose to centralize all this utility stuff in `requstargs` or `codec` packages (or a mix of both). This would also remove complexity in `solo` since it has almost the same logic duplicated.

Here (and in Solo) we should only receive a `RequestArgs` or `Dict` instance. How to construct that instance is not a responsibility of `chainclient` or `solo`.

dessaya (approved):

client/chainclient/chainclient.go
jorgemmsilva: shouldn't it be like this?
```go
for i := 0; i < len(colval); i+=2 {
		key, ok := colval[i].(ledgerstate.Color)
		if !ok {
			panic("toMap: ledgerstate.Color expected")
		}
		ret[key] = encodeIntToUint64(colval[i+1])
	}
```

client/chainclient/chainclient.go
lunfardo314: agree

PR #204, opened 2021-06-22
Consensus test

Juliusan:
As I am moving to cluster tests, thought that this might be a good thing to merge, what I've done. Unfortunately consensus `TestCruelWorld` fails some times (I would say 30% failure rate), but the problem is probably in ACS algorithm and Karolis is investigating it.


packages/chain/statemgr/timers.go
```diff
@@ -5,54 +5,20 @@ package statemgr
...
+const TimerGetBlockRetryConstNameConst = "GetBlockRetry"
+
+func NewStateManagerTimers() util.TimerParams {
+	return util.NewTimerParams(
+		// period of state pull retry
+		util.NewTimerParam(TimerPullStateRetryNameConst, 1*time.Second),
+		// how long delay state pull after state candidate received
+		util.NewTimerParam(TimerPullStateAfterStateCandidateDelayNameConst, 1*time.Second),
+		util.NewTimerParam(TimerGetBlockRetryConstNameConst, 3*time.Second),
+	)
```
dessaya: What is the benefit of changing from the static struct to a dynamic map?

Does something like this not work for some reason?

```go
type TimeParams struct {
	PullStateRetry         time.Duration
	PullStateNewBlockDelay time.Duration
	GetBlockRetry          time.Duration
}

func NewStateManagerTimers() TimeParams {
  // default values here
  return Timers{
	PullStateRetry:         1 * time.Second
	PullStateNewBlockDelay: 1 * time.Second
	GetBlockRetry:          3 * time.Second
  }
}

func (t TimeParams) WithPullStateRetry(v time.Duration) TimeParams {
  t.PullStateRetry = v
  return t
}

func (t TimeParams) WithPullStateNewBlockDelay(v time.Duration) TimeParams {
  t.PullStateNewBlockDelay = v
  return t
}

func (t TimeParams) WithGetBlockRetry(v time.Duration) TimeParams {
  t.GetBlockRetry = v
  return t
}
```

lunfardo314: I agree with the comment. Statical structure is more appropriate

lunfardo314 (approved):

packages/chain/statemgr/timers.go
Juliusan: This is to avoid annoying code repetition. For each retry you need to write a `with` method. An analogous class is used in state manager and in consensus. It's annoying to copy paste half of the code from one place to another. Moreover, it is not safe to use class fields. That's why I use `get` methods, that must be copied too. Of course, you can avoid writing methods completely and use only fields. In that case there would be no code duplication, however the code wouldn't look so nice IMO.

So to sum up, I think this solution is nice, because all the common code is in one place (`packages/util/timers.go`) and for the application you only need to do what is really needed: describe the timers, their defaults and return a collection of them.

dessaya: Adding getters for no reason is so Java :rofl: 

Seriously though, accessing the fields from outside is just fine. As far as I can see, this is just a data object, no need to encapsulate anything. The only logic it needs is setting the default values, which can be handled by the constructor.

jorgemmsilva (approved):

dessaya (approved):

PR #207, opened 2021-06-23
Add cluster tests for big offledger request and access nodes

jorgemmsilva:


jorgemmsilva:
tests are expected to fail until this is merged: https://github.com/iotaledger/goshimmer/pull/1503

tools/cluster/tests/offledger_requests_test.go
```diff
@@ -1,66 +1,166 @@
...
+	"github.com/iotaledger/wasp/packages/vm/core/blob"
+	"github.com/iotaledger/wasp/packages/vm/core/root"
+	"github.com/iotaledger/wasp/tools/cluster"
+	clutest "github.com/iotaledger/wasp/tools/cluster/testutil"
 	"github.com/stretchr/testify/require"
 )
 
-func TestOffledgerRequests(t *testing.T) {
+//nolint:unparam
+func newWalletWithFunds(t *testing.T, clust *cluster.Cluster, ch *cluster.Chain, wapnode int, seedN, iotas uint64) (*chainclient.Client, *ed25519.KeyPair) {
```
dessaya: This is new code? Why is nolint needed? :thinking:

dessaya: Also there is a typo: `wapnode`

dessaya (approved):

tools/cluster/tests/offledger_requests_test.go
jorgemmsilva: thanks, fixed the typo.
`//nolint:unparam` if because it's returning the user wallet `*ed25519.KeyPair` and its not used anywhere. It made sense for me to keep it as it can be needed

dessaya: This being a private function, I think it's better to just remove the unused return value, and add it again whenever it is needed. Using `nolint` should be for really exceptional cases IMO.

jorgemmsilva: sure, updated 👍

lunfardo314 (approved):

PR #208, opened 2021-06-25
updated common account adjustment logic

lunfardo314:
- updated logic does not automatically adjust owner's `agentID` to the common account. The chain owner's `agentID` is treated just as usual account. It only has rights when controlling common account
- implemented `harvest` method in `accounts`. It allows owner of the chain transfer funds from common account to its own account


dessaya (approved):

packages/vm/core/accounts/impl.go
```diff
@@ -18,85 +18,130 @@ func initialize(ctx coretypes.Sandbox) (dict.Dict, error) {
...
+	// otherwise assume harvest all
+	amount, err := par.GetUint64(ParamWithdrawAmount)
+	harvestAll := true
+	if err == nil && amount > 0 {
+		harvestAll = false
+	}
+	// if color not specified and amount is specified, default is harvest specified amount of iotas
+	color := par.MustGetColor(ParamWithdrawColor, ledgerstate.ColorIOTA)
+
+	a.RequireChainOwner(ctx, "harvest")
```
jorgemmsilva: this could be called at the top of the function:
```go
	state := ctx.State()
	a := assert.NewAssert(ctx.Log())
	a.RequireChainOwner(ctx, "harvest")
```

jorgemmsilva (approved):

PR #214, opened 2021-06-28
Manual peering

kape1395:
- Node identity crypto updated: kyber -> hive.
  - Only trusted nodes can participate in the peering network.
  - Signatures added to the peering messages, CRC removed.


packages/registry/registry_impl.go
```diff
@@ -234,71 +309,44 @@ func (r *Impl) HasBlob(h hashing.HashValue) (bool, error) {
 // region NodeIdentity //////////////////////////////////////////
 
 // GetNodeIdentity implements NodeIdentityProvider.
-func (r *Impl) GetNodeIdentity() (*key.Pair, error) {
+func (r *Impl) GetNodeIdentity() (*ed25519.KeyPair, error) {
```
lunfardo314, line 312: wouldn't it be nicer to have something like `node.ID` or similar instead of `*ed25519.KeyPair` everywhere?

tools/cluster/cluster.go
```diff
@@ -47,6 +48,26 @@ func (clu *Cluster) GoshimmerClient() *goshimmer.Client {
 	return goshimmer.NewClient(clu.Config.goshimmerAPIHost(), clu.Config.FaucetPoWTarget)
 }
 
+func (clu *Cluster) TrustAll() error {
```
lunfardo314, line 51: I like the name of the function 😄

lunfardo314 (approved):

jorgemmsilva (approved):

packages/registry/registry_impl.go
kape1395: Lets discuss and do that in a separate PR just to avoid having this not merged for to long?

PR #215, opened 2021-06-29
remove global registries of native contracts + vm types

dessaya:
This PR removes some global / automatic initialization code that may be unwanted in some cases (e.g. from a CLI tool you import some contract's interface, you don't want it to run the node initialization code).

* Removed global variable `allExamples`, now example/native contracts have to be registered explicitly before deployment
* Removed global variable `vmconstructors`
* Created new `processors.Config` object containing both registered vm types and native contracts. A `Config` instance is initialized in the `processors` plugin and later passed to the `Chains` object as a parameter.
* Renamed VM type names (`builtinvm` -> `core`, `examplevm` -> `native`, `wasmtimevm` -> `wasmtime`) and centralized in new `vmtypes` package
* When initializing Solo, if you need to use a native contract you must do something like `solo.New(t, false, false).WithNativeContract(mycontract.Interface)`


packages/vm/readme.md
```diff
@@ -135,9 +135,9 @@ The call to `processors.RegisterVMType` takes name of the new _VM type_ and the
 new `coretypes.Processor` object from the binary data of the program.
 
 The following _VM types_ are pre-defined in the current release of the Wasp:
-* `builtinvm` represents core contracts
-* `examplevm` represents example contracts which conforms to the native interface and are hardcoded before run
-* `wasmtimevm` represents Wasmtime WebAssembly interpreter and native `Rust/Wasm` environment to create smart contracts.
+* `core` represents core contracts
+* `native` represents example contracts which conforms to the native interface and are hardcoded before run
```
lunfardo314: suggestion: ```represents example and other contracts (such as EVM contract) which conforms....``` 
The EVM contract is native but not an example

packages/vm/readme.md
```diff
@@ -135,9 +135,9 @@ The call to `processors.RegisterVMType` takes name of the new _VM type_ and the
...
+* `wasmtime` represents Wasmtime WebAssembly interpreter and native `Rust/Wasm` environment to create smart contracts.
 
 To implement new types of interpreters, for example EVM, other languages or interpreters, a new _VM Type_
```
lunfardo314: this is not needed here ```, for example EVM``` 
We implement EVM as just native contract

lunfardo314 (approved):

packages/vm/readme.md
jorgemmsilva: EVM could be considered a "core" contract in this case
Related: will EVMchain be optional or always built-in on every wasp node deployment?

dessaya: We haven't decided that about the evmchain contract.

PR #221, opened 2021-07-01
Fix: improve cluster test stability

jorgemmsilva:
some measures to make the cluster tests more stable:
- on startup wait for API to be ready by polling the /info endpoint, instead of relying on stdout
- re-broadcast offledger requests in an interval (configurable) // <- this will be improved in a following PR with an ACK message to prevent network spam
- more solid wait condition for cluster tests: reusable`waitUntil` function

### DIFF
--- a/plugins/webapi/plugin.go
+++ b/plugins/webapi/plugin.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"net"
 	"net/http"
-	"sync"
 	"time"
 
 	"github.com/iotaledger/hive.go/daemon"
@@ -28,18 +27,14 @@ var (
 	Server echoswagger.ApiRoot
 
 	log *logger.Logger
-
-	initWG sync.WaitGroup
 )
 
 func Init() *node.Plugin {
 	Plugin := node.NewPlugin(PluginName, node.Enabled, configure, run)
-	initWG.Add(1)
 	return Plugin
 }
 
 func WaitUntilIsUp() { // TODO: Not used?
-	initWG.Wait()
 }
 
 func configure(*node.Plugin) {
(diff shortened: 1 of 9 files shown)


plugins/webapi/plugin.go
```diff
@@ -28,18 +27,14 @@ var (
...
-	initWG sync.WaitGroup
 )
 
 func Init() *node.Plugin {
 	Plugin := node.NewPlugin(PluginName, node.Enabled, configure, run)
-	initWG.Add(1)
 	return Plugin
 }
 
 func WaitUntilIsUp() { // TODO: Not used?
```
dessaya: If this function is not used, let's remove it?

tools/cluster/cluster.go
```diff
@@ -310,12 +313,34 @@ func (clu *Cluster) startServer(command, cwd, name string, initOk chan<- bool, i
...
 		stdoutPipe,
 		func(line string) { fmt.Printf("[ %s] %s\n", name, line) },
-		waitFor(initOkMsg, initOk),
 	)
+	go clu.waitForAPIReady(initOk, nodeIndex)
 
 	return cmd, nil
 }
 
+const poolAPIInterval = 500 * time.Millisecond
```
dessaya: typo: `pool` -> `poll`

tools/cluster/tests/util.go
```diff
@@ -275,3 +280,64 @@ func findContract(chain *cluster.Chain, name string) (*root.ContractRecord, erro
...
+			chain.ChainID, incCounterSCHname, inccounter.FuncGetCounter,
+		)
+		require.NoError(t, err)
+		counter, _, err := codec.DecodeInt64(ret.MustGet(inccounter.VarCounter))
+		require.NoError(t, err)
+		return counter == expected
+	}
+}
+
+func createCheckContractDeployedFn(chain *cluster.Chain, contractName string) conditionFn {
```
dessaya: Suggest renaming these to `counterEquals` and `contractIsDeployed`

So this becomes slightly more readable:

```diff
-waitUntil(t, createCheckContractDeployedFn(chain1, incCounterSCName), makeRange(0, 9), 30*time.Second)
+waitUntil(t, counterEquals(chain1, incCounterSCName), makeRange(0, 9), 30*time.Second)
```

dessaya (approved):

PR #225, opened 2021-07-01
EVM support

dessaya:


contracts/native/evmchain/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+The `--alloc` parameter specifies the genesis allocation for the EVM chain,
+with syntax `address:wei [address:wei ...]`.
+
+## JSON-RPC
+
+Once your EVM chain is deployed, you can use the `wasp-cli chain evm jsonrpc`
+command to start a JSON-RPC server. This will allow you to connect any standard
+Ethereum tool, like Metamask.
```
jorgemmsilva, line 37: in theory we could have multiple evmchains on a single wasp node (as we can have multiple ISCP chains), could this be a problem wrt the json rpc server?

packages/evm/evmtest/keys.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+package evmtest
+
+import (
+	"crypto/ecdsa"
+	_ "embed"
```
jorgemmsilva: I dont think this import is needed here

packages/evm/jsonrpc/jsonrpc_test.go
```diff
@@ -0,0 +1,633 @@
...
+func (e *testEnv) transactionByBlockHashAndIndex(blockHash common.Hash, index uint) *types.Transaction {
+	tx, err := e.client.TransactionInBlock(context.Background(), blockHash, index)
+	if errors.Is(err, ethereum.NotFound) {
+		return nil
+	}
+	require.NoError(e.t, err)
+	return tx
+}
+
+func (e *testEnv) uncleByBlockHashAndIndex(blockHash common.Hash, index uint) map[string]interface{} {
```
jorgemmsilva, line 193: I don't think we can have uncle blocks 🤷‍♂️

jorgemmsilva (approved):
Awesome stuff 🚀

packages/evm/jsonrpc/jsonrpc_test.go
dessaya: That's right, and actually the service implementation always returns nil. The test in line 399 asserts that it always returns nil, for completeness.

contracts/native/evmchain/README.md
dessaya: Yeah, nice catch. For now the cli tool expects the contract to be named `evmchain`. I'll add that as a configuration parameter.

jorgemmsilva: even then, it could potentially be a problem with overlapping ports, right? if we have to run multiple rpc servers in the same wasp node. we probably also need a way to specify the port on that command

dessaya: The port is already configurable with `--listen` flag

PR #226, opened 2021-07-02
Limitting number of requests in batch

Juliusan:
In addition to limiting number of requests in single transaction (I've discussed with Evaldas why his solution is not working), one new test is included, which fails without this fix. Another step is to increase number of iterations in the test to 4 and see why it fails there.

And a minor change: while grepping through the logs "FAIL: state does not exist" messages make it harder to find the failed test cases. Changed "FAIL" to "ERROR".


packages/chain/consensus/action.go
```diff
@@ -126,6 +118,39 @@ func (c *Consensus) prepareVMTask(reqs []coretypes.Request) *vm.VMTask {
...
+	for _, req := range reqs {
+		_, isOnLedgerReq := req.(*request.RequestOnLedger)
+		if isOnLedgerReq {
+			onLedgerCount++
+			if onLedgerCount >= ledgerstate.MaxInputCount {
+				// do not include more on-ledger requests that number of tx inputs allowed-1 ("-1" for chain input)
+				include = false
+			} else {
+				include = true
+			}
```
jorgemmsilva, line 134:
```suggestion
			// do not include more on-ledger requests that number of tx inputs allowed-1 ("-1" for chain input)
			include =  onLedgerCount < ledgerstate.MaxInputCount
```

packages/chain/consensus/action.go
```diff
@@ -126,6 +118,39 @@ func (c *Consensus) prepareVMTask(reqs []coretypes.Request) *vm.VMTask {
...
+		} else {
+			include = true
+		}
+		if include {
+			reqsFiltered = append(reqsFiltered, req)
+		}
+	}
```
jorgemmsilva: imo logic could be simplified a bit, maybe extracted to a separate function:
```suggestion
	for _, req := range reqs {
		_, isOnLedgerReq := req.(*request.RequestOnLedger)
		if isOnLedgerReq {
			onLedgerCount++
			if onLedgerCount > ledgerstate.MaxInputCount {
				// do not include more on-ledger requests that number of tx inputs allowed-1 ("-1" for chain input)
				continue
			}
		}
		reqsFiltered = append(reqsFiltered, req)
	}
```

jorgemmsilva: maybe above should be `>=` (according to the current implementation), but I think `>` should be fine ☝️

tools/cluster/tests/advanced_inccounter_test.go
```diff
@@ -60,6 +60,60 @@ func TestAccessNode(t *testing.T) {
...
+	t.Logf("addr1: %s", addr1.Base58())
+
+	chain1, err := clu1.DeployChain("chain", clu1.Config.AllNodes(), committee, uint16(consensusSize), addr1)
+	require.NoError(t, err)
+	t.Logf("chainID: %s", chain1.ChainID.Base58())
+
+	description := "testing with inccounter"
+	programHash = inccounter.Interface.ProgramHash
+
+	_, err = chain1.DeployContract(contractName, programHash.String(), description, nil)
```
jorgemmsilva: `contractName` was renamed to: `incCounterSCName`, to be more explicit

dessaya (approved):

tools/cluster/tests/advanced_inccounter_test.go
Juliusan: Sorry... Missed that after I merged `develop`. Fixed.

packages/chain/consensus/action.go
Juliusan: I don't like `continue`. It reminds me of structural programming too much. I also do not like functions, which are called only once in the code :-) But I will fix it.

jorgemmsilva: okay, feel free to keep as-is. This is more of a style preference than anything 👍

jorgemmsilva (approved):

lunfardo314 (approved):

PR #228, opened 2021-07-05
Feat: broadcast missing reqs

jorgemmsilva:
### consensus:

- implements push/pull of missing requests: committee nodes that see an ACS batch being agreed upon and are lacking some of the requests contained on said batch can query the other committee nodes by sending a `MissingRequestIDsMsg` containing the IDs of the requests missing. This functionality can be toggled on/off by specifying `peering.pullMissingRequests` in the node configuration.
- in order to protect against malformed on-ledger requests being push/pulled (because we don't store the entire tagle TX, only the relevant data form the output), this PR adds the request HASH to the batch proposal payload and uses this hash to calculate the batches intersection

--- 

### offledger requests:
- implements ACK messages to notify nodes that offledger requests has been received and they can stop sending that request, thus reducing network spam

--- 

### misc: 
- renamed `gossipOffledgerRequests` to `broadcastOffledgerRequests` as it is a better representation of what is being done.
- created `messages` package inside `chain` so we have a central place where these messages are defined.


packages/chain/mempool/mempool.go
```diff
@@ -271,26 +271,22 @@ func (m *Mempool) ReadyNow(now ...time.Time) []coretypes.Request {
...
-	m.poolMutex.RLock()
-	defer m.poolMutex.RUnlock()
-
-	ret := make([]coretypes.Request, 0, len(reqids))
-	for _, reqid := range reqids {
+func (m *Mempool) ReadyFromIDs(nowis time.Time, reqids ...coretypes.RequestID) ([]coretypes.Request, []int, bool) {
+	requests := make([]coretypes.Request, 0, len(reqids))
+	missingRequestIndexes := []int{}
+	for i, reqid := range reqids {
 		reqref, ok := m.pool[reqid]
```
Juliusan: Unprotected read. Shouldn't it be a call to `GetRequest`?

packages/chain/chainimpl/interface.go
```diff
@@ -99,43 +100,96 @@ func (c *chainObj) ReceiveMessage(msg interface{}) {
...
-			(*c.peers).SendMsgToRandomPeersSimple(uint16(gossipUpToNPeers), chain.MsgOffLedgerRequest, msgData)
+		broadcastUpToNPeers := parameters.GetInt(parameters.OffledgerBroadcastUpToNPeers)
+		getPeerIDs = func() []string {
+			return (*c.peers).GetRandomPeers(broadcastUpToNPeers)
 		}
 	}
-	broadcastInterval := time.Duration(parameters.GetInt(parameters.OffledgerBroadcastInterVal)) * time.Millisecond
 
-	ticker := time.NewTicker(broadcastInterval)
+	sendMessage := func(shouldSendToPeer shouldSendToPeerFn) {
```
dessaya: If I understood correclty, `shouldSendToPeer` is always the result of `shouldSend`? Then does it need to be a parameter? I'm asking because the logic is somewhat difficult to follow with the dynamic closures.

packages/chain/consensus/batch_proposal.go
```diff
@@ -153,24 +162,33 @@ func (c *Consensus) calcBatchParameters(props []*BatchProposal) (*consensusBatch
 // calcIntersection a simple algorithm to calculate acceptable intersection. It simply takes all requests
 // seen by 1/3+1 node. The assumptions is there can be at max 1/3 of bizantine nodes, so if something is reported
 // by more that 1/3 of nodes it means it is correct
+const keyLen = ledgerstate.OutputIDLength + 32
+
```
dessaya: The godoc comment above applies to the `calcIntersection` function

```suggestion
const keyLen = ledgerstate.OutputIDLength + 32

// calcIntersection a simple algorithm to calculate acceptable intersection. It simply takes all requests
// seen by 1/3+1 node. The assumptions is there can be at max 1/3 of bizantine nodes, so if something is reported
// by more that 1/3 of nodes it means it is correct
```

packages/chain/messages/missing_req_msg.go
```diff
@@ -0,0 +1,121 @@
...
+	"io/ioutil"
+
+	"github.com/iotaledger/goshimmer/packages/ledgerstate"
+	"github.com/iotaledger/wasp/packages/coretypes"
+	"github.com/iotaledger/wasp/packages/coretypes/request"
+	"github.com/iotaledger/wasp/packages/util"
+	"golang.org/x/xerrors"
+)
+
+// region MissingRequestsMsg ///////////////////////////////////////////////////
```
dessaya:
```suggestion
// region MissingRequestIDsMsg ///////////////////////////////////////////////////
```

packages/chain/messages/missing_req_msg.go
```diff
@@ -0,0 +1,121 @@
...
+func MissingRequestIDsMsgFromBytes(buf []byte) (MissingRequestIDsMsg, error) {
+	r := bytes.NewReader(buf)
+	msg := MissingRequestIDsMsg{}
+	err := msg.read(r)
+	return msg, err
+}
+
+// endregion ///////////////////////////////////////////////////////////////////
+
+// region RequestMsg ///////////////////////////////////////////////////
```
dessaya:
```suggestion
// region MissingRequestMsg ///////////////////////////////////////////////////
```

dessaya (approved):

packages/chain/mempool/mempool.go
jorgemmsilva: you are right, added mutex usage here 👌 
calling getRequest would have some mutex lock/unlock overhead (for each reqID),  better just lock/unlock once

packages/chain/chainimpl/interface.go
jorgemmsilva: yeah, it seemed like a good idea at the time of writing, but I see its hard to read. Will refactor to a more readable version. 👍

jorgemmsilva: @dessaya should be better now

packages/parameters/parameters.go
```diff
@@ -60,13 +62,15 @@ func InitFlags() {
...
+	flag.StringSlice(PeeringNeighbors, []string{}, "list of neighbors: known peer netIDs")
+	flag.Bool(PullMissingRequestsFromCommittee, true, "whether or not to pull missing requests from other committee members")
 
 	flag.Int(NanomsgPublisherPort, 5550, "the port for nanomsg even publisher")
 
 	flag.String(IpfsGatewayAddress, "https://ipfs.io/", "the address of HTTP(s) gateway to which download from ipfs requests will be forwarded")
 
-	flag.Int(OffledgerGossipUpToNPeers, 10, "number of peers an offledger request is gossiped to")
-	flag.Int(OffledgerBroadcastInterVal, 1000, "time between re-broadcast of offledger requests (in ms)")
+	flag.Int(OffledgerBroadcastUpToNPeers, 10, "number of peers an offledger request is broadcasted to")
```
lunfardo314, line 72: not sure of we need it as node config parameters. Constants would be enough?

packages/coretypes/request/requestimpl.go
```diff
@@ -376,7 +500,16 @@ func (req *RequestOffLedger) Essence() []byte {
...
+	var buf bytes.Buffer
+	buf.WriteByte(OffLedgerRequestType)
+	buf.Write(req.Essence())
+	buf.Write(req.signature[:])
+	return buf.Bytes()
+}
+
+// only used for consensus
+func (req *RequestOffLedger) Hash() [32]byte {
+	return blake2b.Sum256(req.Bytes())
```
lunfardo314: to be consistent better use `hashing.HashData()` instead of `blake2b` package directly.

lunfardo314 (approved):

packages/parameters/parameters.go
jorgemmsilva: seems like something we might want to tweak in a testing environment, that's why I though it would be a good idea to have it as a parameter

PR #230, opened 2021-07-06
Wasmvm

BugFreeSoftware:


contracts/rust/wasmlib/src/corecontracts/coreblob/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+use crate::*;
+use crate::corecontracts::coreblob::*;
+
+pub struct StoreBlobCall {
+    pub func:    ScFunc,
+    pub params:  MutableStoreBlobParams,
+    pub results: ImmutableStoreBlobResults,
+}
+
+impl StoreBlobCall {
```
lunfardo314, line 23: note that storing a blob from wasm is not a normal thing

lunfardo314 (approved):

PR #231, opened 2021-07-06
Fix `wasp-cli mint` command (fixes #211)

### LINKED ISSUE #211: mint error
I use the wasp-cluster. Balance is 1mi but with mint i get the following error message
`error: prepareColoredBalancesOutput: no tokens to transfer`

dessaya:


tools/cluster/tests/wasp-cli_test.go
```diff
@@ -199,3 +200,38 @@ func TestWaspCLIBlobRegistry(t *testing.T) {
...
+	{
+		found := false
+		for _, line := range out {
+			if strings.Contains(line, color) {
+				require.Regexp(t, fmt.Sprintf("(?m)%s:[[:space:]]+1000$", color), line)
+				found = true
+			}
+		}
+		require.True(t, found)
+	}
```
jorgemmsilva: testing by checking console output seems like a slippery slope. couldn't we just check if the tokens were minted instead?

jorgemmsilva (approved):

lunfardo314 (approved):

tools/cluster/tests/wasp-cli_test.go
dessaya: The purpose of the test is to verify the behavior of `wasp-cli`, and the output is part of what is being tested here. We want to assert that the command shows a helpful message when it succeeds.

dessaya: Although it's debatable that this test is testing two things: the `mint` and `balance` commands, and maybe it should be split in two. I'll do that.

PR #232, opened 2021-07-06
Fix dashboard

dessaya:


packages/dashboard/dashboard_test.go
```diff
@@ -54,10 +62,10 @@ func TestDashboardChainAccount(t *testing.T) {
 		"chainid": chainid.RandomChainID().Base58(),
 		"agentid": strings.Replace(coretypes.NewRandomAgentID().String(), "/", ":", 1),
 	})
-
-	// make sure we are using .Base58()
-	require.NotContains(t, html.Text(), "OutputID {")
-	require.NotContains(t, html.Text(), "Address {")
+	checkProperConversionsToString(t, html)
+	require.Equal(t, 1, html.Find("#on-chain-account dl").Length())
+	require.Equal(t, "AgentID", html.Find("#on-chain-account dl dt").Text())
```
jorgemmsilva: this kind of selectors are brittle and should be avoided: `#id dl dt`. generally it's better to just add a test class to the element in question

jorgemmsilva (approved):

lunfardo314 (approved):

packages/dashboard/dashboard_test.go
dessaya: I added the `on-chain-account` id specifically for this test, it's not used anywhere else. Isn't that the same? What would be the benefit of making it a class instead?

PR #236, opened 2021-07-07
fix state checks in cluster tests

dessaya:
Funny story: The cluster function `WithSCState()` was not working (for several months at least!). Any assertions made using that function would always pass.

This PR:

* Removes the `WithSCState()` function. It is replaced by a simple loop to assert the state in all nodes: `for i := range chain.CommitteeNodes { ... }`
* Removes old webapi endpoints that allowed to dump the full state (dangerous/slow when the state is big)
* Adds a new webapi endpoint that allows to get a single value from the state given the raw key (poor man's alternative for view calls)
* Attempts to update and fix all cluster tests that used `WithSCState()`

The following tests are **not passing** and I can't figure out why, maybe it's unrelated to this change:

* `TestDeployContractAndSpawn`
* `TestIncRepeatManyIncrement`


tools/cluster/tests/deploy_test.go
```diff
@@ -38,14 +36,56 @@ func TestDeployChain(t *testing.T) {
...
+		require.EqualValues(t, 1, blockIndex)
+
+		contractRegistry, err := chain1.ContractRegistry(i)
+		require.NoError(t, err)
+
+		// contract was not deployed:
+		require.EqualValues(t, len(core.AllCoreContractsByHash), len(contractRegistry))
+	}
+
+	// TODO: check error message
```
jorgemmsilva, line 88: TODO here 👀

jorgemmsilva (approved):

tools/cluster/tests/deploy_test.go
dessaya: Yeah, I'll do that in a following PR (it was my initial intention before I realized most of the tests were broken :P).

PR #241, opened 2021-07-08
EVM improvements

dessaya:
* configurable chainID (default 1074)
* net_version responds the same value as chainID
* add txpool service (necessary for chain explorer)

### DIFF
--- a/packages/evm/jsonrpc/service.go
+++ b/packages/evm/jsonrpc/service.go
@@ -8,6 +8,7 @@ package jsonrpc
 import (
 	"fmt"
 	"math/big"
+	"strconv"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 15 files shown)


BugFreeSoftware (approved):

packages/evm/jsonrpc/service.go
```diff
@@ -282,14 +282,16 @@ func (e *EthService) CompileSolidity()
...
 
-func NewNetService() *NetService {
-	return &NetService{}
+func NewNetService(chainID int) *NetService {
+	return &NetService{chainID}
 }
 
 func (s *NetService) Version() string {
-	return "1074" // IOTA -- get it?
+	return strconv.Itoa(s.chainID)
```
lunfardo314, line 294: So it means there are no 2 chains with same version? Should’t the version be a version of the evm native contract? Some static constant in the code? And, btw, the committee nodes via the chain should enforce all nodes have compatible (same) versions?

dessaya: This chainID is not the ISCP chainID, it is an arbitrary number specified for the EVM chain, configured when initializing the evmchain contract.
This `Version` method corresponds to the `net_version` JSON-RPC endpoint, and is supposed to return the so called NetworkID which in most cases should be equal to the ChainID.

PR #242, opened 2021-07-09
Improvement: better offledger API info

jorgemmsilva:
Adds checks for common mistakes when sending offledger requests: invalid signature or empty account balance.

### DIFF
--- a/packages/webapi/request/request.go
+++ b/packages/webapi/request/request.go
@@ -6,7 +6,9 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/iotaledger/goshimmer/packages/ledgerstate"
 	"github.com/iotaledger/wasp/packages/chain"
+	"github.com/iotaledger/wasp/packages/coretypes"
 	"github.com/iotaledger/wasp/packages/coretypes/chainid"
 	"github.com/iotaledger/wasp/packages/coretypes/request"
 	"github.com/iotaledger/wasp/packages/webapi/httperrors"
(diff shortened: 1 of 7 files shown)


packages/webapi/request/request.go
```diff
@@ -43,10 +50,27 @@ func (o *offLedgerReqAPI) handleNewRequest(c echo.Context) error {
...
 	ch := o.getChain(chainID)
 	if ch == nil {
 		return httperrors.NotFound(fmt.Sprintf("Unknown chain: %s", chainID.Base58()))
 	}
+
+	// check user has on-chain balance
+	balances, err := o.getAccountBalance(ch, offLedgerReq.SenderAccount())
+	if err != nil {
+		return httperrors.ServerError("Unable to get account balance")
+	}
```
dessaya, line 68: Shouldn't this check be done by the client instead of the server? I imagine that it makes some kind of ~request to goshimmer~ view call for this, and it makes the off-ledger request somewhat slower.

packages/webapi/request/request.go
```diff
@@ -43,10 +50,27 @@ func (o *offLedgerReqAPI) handleNewRequest(c echo.Context) error {
...
+
+	if len(balances) == 0 {
```
dessaya, line 70: Instead of comparing against 0, shouldn't it check against the `Transfer` component of the request?

packages/webapi/request/request.go
jorgemmsilva: no, this is "wasp chain balance". if the client does this it would just hit the webapi on a different endpoint.
also doing this prevents the requests to be added to mempool/broadcasted, etc

packages/webapi/request/request.go
jorgemmsilva: can also check that, but this check is needed. Doesn't matter if the request has `Transfer` if the account is empty 😄 . Also I think we automatically charge exec fees from the sender account, so the `Transfer` is not mandatory

lunfardo314: "Account is empty" means sender is not known to the chain. It is a fast first line protection against blind spamming attack. Is it enough tokens etc is checked by the VM.

lunfardo314 (approved):

dessaya (approved):

PR #245, opened 2021-07-11
Signature acknowledgement

lunfardo314:
in consensus: each signed VM result message is now acknowledged by special message and re-broadcasting stops

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/mr-tron/base58 v1.2.0
 	github.com/pangpanglabs/echoswagger/v2 v2.1.0
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.7.0
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.0
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,8 @@ module github.com/iotaledger/wasp
 go 1.16
 
 require (
+	github.com/NebulousLabs/errors v0.0.0-20181203160057-9f787ce8f69e // indirect
+	github.com/NebulousLabs/fastrand v0.0.0-20181203155948-6fb6489aac4e // indirect
 	github.com/PuerkitoBio/goquery v1.6.1
 	github.com/anthdm/hbbft v0.0.0-20190702061856-0826ffdcf567
 	github.com/bytecodealliance/wasmtime-go v0.21.0
@@ -14,6 +16,7 @@ require (
 	github.com/mr-tron/base58 v1.2.0
 	github.com/pangpanglabs/echoswagger/v2 v2.1.0
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.7.0
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.0
--- a/packages/chain/chainimpl/chainimpl.go
+++ b/packages/chain/chainimpl/chainimpl.go
@@ -216,15 +216,16 @@ func (c *chainObj) processPeerMessage(msg *peering.PeerMessage) {
 			c.consensus.EventSignedResultMsg(msgt)
 		}
 	case messages.MsgSignedResultAck:
+		if c.consensus == nil {
+			return	
+		}
 		msgt := &messages.SignedResultAckMsg{}
 		if err := msgt.Read(rdr); err != nil {
 			c.log.Error(err)
 			return
 		}
 		msgt.SenderIndex = msg.SenderIndex
-		if c.consensus != nil {
-			c.consensus.EventSignedResultAckMsg(msgt)
-		}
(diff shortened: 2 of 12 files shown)


go.mod
```diff
@@ -14,6 +16,7 @@ require (
 	github.com/mr-tron/base58 v1.2.0
 	github.com/pangpanglabs/echoswagger/v2 v2.1.0
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.7.0
```
dessaya: What is this new dependency for?

packages/chain/consensus/eventproc.go
```diff
@@ -1,4 +1,4 @@
-package consensus
+package consensus //nolint:nolintlint,gofmt
```
dessaya: Can we fix the warnings instead of supressing the linter? Or disable the checks if they mostly give false positives.

packages/peering/group/group.go
```diff
@@ -181,17 +181,36 @@ func (g *groupImpl) ExchangeRound(
...
+	exceptions := make(map[uint16]struct{})
+	for _, i := range except {
+		exceptions[i] = struct{}{}
+	}
+	for i := range g.nodes {
+		if _, ok := exceptions[uint16(i)]; !ok {
+			ret[uint16(i)] = g.nodes[i]
+		}
+	}
+	return ret
```
dessaya: This block looks very similar to the one above, in `AllNodes`. Can it be refactored into a single function?

packages/peering/group/group.go
```diff
@@ -181,17 +181,36 @@ func (g *groupImpl) ExchangeRound(
...
+func (g *groupImpl) OtherNodes(except ...uint16) map[uint16]peering.PeerSender {
+	if len(except) == 0 {
+		return g.other
+	}
+	ret := make(map[uint16]peering.PeerSender)
+	exceptions := make(map[uint16]struct{})
+	for _, i := range except {
+		exceptions[i] = struct{}{}
+	}
+	for i := range g.nodes {
```
dessaya:
```suggestion
	for i := range g.other {
```

go.mod
lunfardo314: hm, no idea... Will fix

packages/chain/consensus/eventproc.go
lunfardo314: if only I knew how... I'll try my best :)

packages/chain/chainimpl/chainimpl.go
```diff
@@ -215,6 +215,16 @@ func (c *chainObj) processPeerMessage(msg *peering.PeerMessage) {
...
+	case messages.MsgSignedResultAck:
+		msgt := &messages.SignedResultAckMsg{}
+		if err := msgt.Read(rdr); err != nil {
+			c.log.Error(err)
+			return
+		}
+		msgt.SenderIndex = msg.SenderIndex
+		if c.consensus != nil {
+			c.consensus.EventSignedResultAckMsg(msgt)
+		}
```
jorgemmsilva: suggestion: early return
```suggestion
		if c.consensus == nil {
			return	
		}
		msgt := &messages.SignedResultAckMsg{}
		if err := msgt.Read(rdr); err != nil {
			c.log.Error(err)
			return
		}
		msgt.SenderIndex = msg.SenderIndex
		c.consensus.EventSignedResultAckMsg(msgt)
```

packages/chain/messages/peermsg.go
```diff
@@ -138,3 +145,23 @@ func (msg *SignedResultMsg) Read(r io.Reader) error {
...
+}
+
+func (msg *SignedResultAckMsg) Read(r io.Reader) error {
+	if err := util.ReadHashValue(r, &msg.EssenceHash); err != nil {
+		return err
+	}
+	if err := util.ReadOutputID(r, &msg.ChainInputID); /* nolint:revive */ err != nil {
+		return err
+	}
+	return nil
```
jorgemmsilva: uneeded if:

```suggestion
	return util.ReadOutputID(r, &msg.ChainInputID)
```

packages/chain/consensus/consensus.go
```diff
@@ -89,10 +91,12 @@ func New(chainCore chain.ChainCore, mempool chain.Mempool, committee chain.Commi
 		nodeConn:                         nodeConn,
 		vmRunner:                         runvm.NewVMRunner(),
 		resultSignatures:                 make([]*messages.SignedResultMsg, committee.Size()),
+		resultSigAck:                     make([]uint16, 0, committee.Size()),
```
jorgemmsilva, line 94: this could potentially be a map: `map[uint16]bool`, so some logic is simplified (creating the `expected` map on some function)

jorgemmsilva (approved):

packages/chain/consensus/consensus.go
lunfardo314: I did it first but then decided this one. The reason is optional parameters on other calls `except ...int` they need slice, not a map

packages/peering/group/group.go
lunfardo314: wanted minimal intrusion into Karolis' code...

PR #250, opened 2021-07-13
Updated off-ledger requests and added tests

lunfardo314:
refactored `nonce` logic in off-ledger requests and added access node tests off-ledger.


client/chainclient/chainclient.go
```diff
@@ -74,7 +77,11 @@ func (c *Client) PostOffLedgerRequest(
 	if len(params) > 0 {
 		par = params[0]
 	}
+	if par.Nonce == 0 {
+		par.Nonce = uint64(time.Now().UnixNano())
+	}
 	offledgerReq := request.NewRequestOffLedger(contractHname, entrypoint, par.Args).WithTransfer(par.Transfer)
+	offledgerReq.WithNonce(par.Nonce)
 	offledgerReq.Sign(c.KeyPair)
```
jorgemmsilva, line 85: an idea: should the nonce be added automatically with `Sign`?

jorgemmsilva (approved):

kape1395 (approved):

client/chainclient/chainclient.go
lunfardo314: may be..

PR #252, opened 2021-07-13
disabled order checking for off-ledger requests in VM 

lunfardo314:


packages/chain/mempool/mempool.go
```diff
@@ -182,7 +182,7 @@ func (m *Mempool) RemoveRequests(reqs ...coretypes.RequestID) {
 	}
 }
 
-const traceInOut = false
+const traceInOut = true
```
jorgemmsilva: is this intended?

jorgemmsilva (approved):

packages/chain/mempool/mempool.go
lunfardo314: it is, but I can disable it. Maybe you are right

PR #254, opened 2021-07-13
Dockerize

lmoe:


Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+EXPOSE 7000/tcp
+EXPOSE 9090/tcp
+EXPOSE 5550/tcp
+EXPOSE 4000/udp
+
+# Config is overridable via volume mount to /run/config.json
+COPY docker_config.json /run/config.json
+
+COPY --from=build /wasp/wasp /usr/bin/wasp
+COPY --from=build /wasp/wasp-cli /usr/bin/wasp-cli
```
dessaya, line 64: You can skip copying wasp-cli here?

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+COPY --from=build /wasp/ /run
+
+CMD go test -tags rocksdb -short ./...
+
+
+
+# Wasp CLI build
+FROM golang:1.16.5-buster as wasp-cli
+COPY --from=build /wasp/wasp-cli /usr/bin/wasp-cli
+ENTRYPOINT ["wasp-cli"]
```
jorgemmsilva, line 46: this should probably be part of the main build phase

docs/docs/deploy.md
```diff
@@ -8,11 +8,12 @@ contract to the chain.
 
 ---
 
-First we need to tell `wasp-cli` the location of the Goshimmer node and the
+First we need to tell `wasp-cli` the location of the Goshimmer node, the faucet PoW target and the
 committee of Wasp nodes:
 
 ```
 $ wasp-cli set goshimmer.api 127.0.0.1:8080
+$ wasp-cli set goshimmer.faucetPoWTarget 22
```
dessaya: It is best to set it to -1 so that it uses the default value, although the command is a bit cumbersome:

```
wasp-cli set -- goshimmer.faucetPoWTarget -1
```

You need the `--` so that it doesn't take `-1` as a flag.

We should also probably make this the default (and set it manually to 0 when testing against mocked goshimmer).

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# Make sure that modules only get pulled when the module file has changed
+COPY go.mod go.sum /wasp/
+RUN go mod download
+RUN go mod verify
+
+# Project build stage
+COPY . .
+
+RUN go build -tags rocksdb
+RUN go build -tags rocksdb ./tools/wasp-cli
```
jorgemmsilva: we could probably use `ARG` to enable custom build via terminal (with/without rocksdb, evm flags, etc)
https://docs.docker.com/engine/reference/builder/#arg

docs/docs/docker.md
```diff
@@ -0,0 +1,89 @@
+DRAFT
+
```
jorgemmsilva:
```suggestion
```

jorgemmsilva (approved):
nice 👍

tools/wasp-cli/README.md
```diff
@@ -45,7 +45,7 @@ set:
 
 ```
 wasp-cli set goshimmer.api goshimmer.node:8080
-wasp-cli set goshimmer.faucetPoWTarget -1
+wasp-cli set goshimmer.faucetPoWTarget 22
```
dessaya: Same thing here, it was just missing the `--`.

dessaya (approved):

Dockerfile
lmoe: I could. But if someone deploys a wasp node it might be required to use the cli though. 

They could use the wasp-cli on the same container with `docker exec wasp -it wasp-cli <args>` instead of deploying a second image they need to connect with the wasp node through a bridged network.

I think we should keep it in there. If you don't, I'll remove it.

PR #255, opened 2021-07-13
Repeat call view on "virtual state has been invalidated"

Juliusan:
A quick update to repeat call view in `counterEquals` function. In the spirit of https://github.com/iotaledger/wasp/pull/248.


kape1395 (approved):

tools/cluster/tests/util.go
```diff
@@ -239,11 +240,27 @@ func waitTrue(timeout time.Duration, fun func() bool) bool {
 	}
 }
 
+func repeatIfInvalidated(fun func() (dict.Dict, error), deadline time.Time) (dict.Dict, error) {
+	result, err := fun()
+	if err != nil {
+		errStr := fmt.Sprintf("%v", err)
+		if strings.Contains(errStr, "virtual state has been invalidated") {
```
jorgemmsilva, line 247: not a fan of checking the error by string matching 😕 
could we create a special kind of error and use `errors.Is`? 
https://pkg.go.dev/errors#Is

tools/cluster/tests/util.go
```diff
@@ -239,11 +240,27 @@ func waitTrue(timeout time.Duration, fun func() bool) bool {
...
+			if time.Now().Before(deadline) {
+				return repeatIfInvalidated(fun, deadline)
```
jorgemmsilva, line 249: should we add some waiting time here before re-trying?

tools/cluster/tests/util.go
Juliusan: Not a fan either. That, I believe, should depend on error code returned from the server. It is done more elegantly on the server side, but do we want to do it here or there? That's a discussion I wanted.

tools/cluster/tests/util.go
Juliusan: Waiting is done on `RepeatOnceIfUnlucky`. I hoped to do without it here... :-)

tools/cluster/tests/util.go
jorgemmsilva: oh okay, I didn't realize this was checking on a http result

lunfardo314 (approved):

PR #256, opened 2021-07-13
Partially convert existing docu and propose a new docu structure.

aleksei-korolev:
Hi!

I made basic setup of docusaurus documentation (in `docOps` folder). Also I moved some parts of old documentation there, so you could use it as a reference to introduce your own changes (because I am very kind, totally not because I misunderstood the task completely). You can find a file structure with empty .md files in the `docOps/docs/in_progress` folder. Other points of interests are `docOps/docusaurus.config.js` (general settings), `docOps/sidebars.js` (ToC), `docOps/src/components/LandingpageComponents` (landing page behavior), `docOps/static` for files other than .md. Also there is a new github workflow to deploy the docs on github pages.

To test the new docs locally, go to `docOps` and run `yarn start`.

I kept the old `docs` folder untouched for now, but the idea is that `docOps/docs` replaces it completely.


dessaya:
I think it will be better to avoid duplicating information / tools / etc. Can we remove the old mdBook configuration if we are not using it anymore? Also remove the `.md` files that were moved to `docOps/`?

aleksei-korolev:
@dessaya could you do it for me please? I moved most of the .md files to the new structure, but there are, for example, source code folder in the tutorial folder and technical specification pdf and a few TODO files (`new` folder and `intro/todo` folders), and I don't really what to do with setup_wasp (docaurus does not support `#include` directive). I want to be sure that I do not lose any information that you need.

You can safely evaporize any traces of mdbooks, we will not use it anymore (I never worked with mdbooks and have no clue where to look).

dessaya:
@aleksei-korolev Ok, let's merge it and then I'll remove the obsolete / duplicate stuff.

dessaya (approved):

PR #258, opened 2021-07-14
Retry on "virtual state has been invalidated"

Juliusan:
It's a continuation of https://github.com/iotaledger/wasp/pull/255. The retry is done on server till some constant timeout. Server returns 409 on "virtual state has been invalidated". The client also retries on 409 until some timeout. I am thinking, maybe the timeout must be passed to http client so the request wouldn't take longer because of longer server timeout.

### DIFF
--- a/packages/webapi/state/callview.go
+++ b/packages/webapi/state/callview.go
@@ -3,6 +3,7 @@ package state
 import (
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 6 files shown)


packages/webapi/state/callview.go
```diff
@@ -72,6 +73,9 @@ func handleCallView(c echo.Context) error {
 	return c.JSON(http.StatusOK, ret)
 }
 
+const retryOnStateInvalidatedRetry = 100 * time.Millisecond //nolint:gofumpt
```
jorgemmsilva, line 76: why the lint ignore?

jorgemmsilva (approved):

lunfardo314 (approved):

packages/webapi/state/callview.go
Juliusan: More clear for me :-) gofumpt offers
```
const(
   a = 1
   b = 2
)
```
but I don't like the grouping :-)

PR #261, opened 2021-07-15
replay protection of off-ledger and MEV prevention

lunfardo314:


packages/vm/vmcontext/runreq.go
```diff
@@ -172,22 +178,20 @@ func (vmctx *VMContext) validateRequest() bool {
...
-	//}
+	// this is a replay protection measure for off-ledger requests assuming in the batch order of requests is random.
+	// See rfc [replay-off-ledger.md]
 
-	return true
+	maxAssumed := accounts.GetMaxAssumedNonce(vmctx.State(), req.SenderAddress())
+	if maxAssumed < OffLedgerNonceStrictOrderTolerance {
+		return true
+	}
+	return req.Nonce() > maxAssumed-OffLedgerNonceStrictOrderTolerance
```
jorgemmsilva, line 194: something about the nonce "range check":
Given a valid nonce (in the accepted range), I can submit N requests (to different SCs), with the same (valid) nonce, and all will accepted right?

I find this a weird property, but maybe it's okay.
What do you think?

packages/vm/vmcontext/runreq.go
```diff
@@ -172,22 +178,20 @@ func (vmctx *VMContext) validateRequest() bool {
...
-	//	vmctx.lastError = fmt.Errorf("validateRequest: invalid order %d <= %d of request %s",
-	//		vmctx.req.Nonce(), lastNonce, req.ID().String())
-	//	return false
-	//}
+	// this is a replay protection measure for off-ledger requests assuming in the batch order of requests is random.
+	// See rfc [replay-off-ledger.md]
 
-	return true
+	maxAssumed := accounts.GetMaxAssumedNonce(vmctx.State(), req.SenderAddress())
+	if maxAssumed < OffLedgerNonceStrictOrderTolerance {
```
jorgemmsilva, line 191: should it be `<=`?
```suggestion
	if maxAssumed <= OffLedgerNonceStrictOrderTolerance {
```

jorgemmsilva (approved):

docs/rfc-tmp/prevent-mev.md
```diff
@@ -0,0 +1,25 @@
+# Prevention of MEV in ISCP
```
dessaya: Documentation is now moved to `docOps` (apparently this is standard across IF repositories).

Propose moving these `.md` files to `docOps/docs/rfc`

packages/vm/core/accounts/internal.go
```diff
@@ -235,11 +235,15 @@ func DecodeBalances(balances dict.Dict) (map[ledgerstate.Color]uint64, error) {
 	return ret, nil
 }
 
-func GetLastNonce(state kv.KVStore, address ledgerstate.Address) uint64 {
-	order, _, _ := codec.DecodeUint64(state.MustGet(kv.Key(address.Bytes()) + "ord"))
-	return order
+func GetMaxAssumedNonce(state kv.KVStore, address ledgerstate.Address) uint64 {
+	nonce, _, _ := codec.DecodeUint64(state.MustGet(kv.Key(address.Bytes()) + "non"))
```
dessaya: `"non"` could be a const?

packages/vm/vmcontext/runreq.go
lunfardo314: yes, it is ok. They can have same nonces but if request ids (hashes) will be the same, it was prevented. However, with each request `maxAssumed` will be increasing, so you wont be able to do it forever, will need to jump to another nonce range.

packages/vm/vmcontext/runreq.go
lunfardo314: I think it is not important. because constant is arbitrary anyway, right?

docs/rfc-tmp/prevent-mev.md
lunfardo314: ok, will do, this was temporary anyway

packages/vm/core/accounts/internal.go
lunfardo314: I just took it from Eric's code :)  Will make a constant

PR #262, opened 2021-07-15
Adds profiling plugin

jorgemmsilva:
instructions on `profiling.md`

### DIFF
--- a/plugins/profiling/plugin.go
+++ b/plugins/profiling/plugin.go
@@ -1,0 +1,82 @@
+package profiling
+
+import (
+	"net/http"
+	"runtime"
+	"sync"
+
+	// import required to profile
+	_ "net/http/pprof"
+
+	profile "github.com/bygui86/multi-profile/v2"
+	"github.com/iotaledger/hive.go/daemon"
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/hive.go/node"
+	"github.com/iotaledger/wasp/packages/parameters"
+)
+
+// PluginName is the name of the profiling plugin.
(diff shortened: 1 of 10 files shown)


plugins/profiling/plugin.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/iotaledger/wasp/packages/parameters"
+)
+
+// PluginName is the name of the profiling plugin.
+const PluginName = "Profiling"
+
+var (
+	// plugin is the profiling plugin.
+	plugin *node.Plugin
+	once   sync.Once
```
dessaya: Is `once` really needed? The `Init` function should be called once by the plugin system anyway.
Also, `plugin` doesn't need to be global I think.

tools/cluster/tests/advanced_inccounter_test.go
```diff
@@ -172,6 +172,10 @@ func testAccessNodesOffLedger(t *testing.T, numRequests, numValidatorNodes, clus
 
 	clu1, chain1 := setupAdvancedInccounterTest(t, clusterSize, cmt)
 
+	defer func() {
+		clu1.Stop()
+	}()
```
dessaya: `defer clu1.Stop()`

dessaya (approved):

tools/cluster/tests/advanced_inccounter_test.go
jorgemmsilva: thanks, it was for testing. After that I realized its called on `t.Cleanup` automatically from `NewCluster`.
removed

PR #268, opened 2021-07-15
Implement prometheus plugin

ivange94:
Implement prometheus plugin


plugins/prometheus/plugin.go
```diff
@@ -0,0 +1,82 @@
...
+				},
+			)
+			handler.ServeHTTP(c.Response(), c.Request())
+			return nil
+		})
+
+		bindAddr := parameters.GetString(parameters.PrometheusBindAddress)
+		server = &http.Server{Addr: bindAddr, Handler: e}
+
+		stopped := make(chan struct{})
```
jorgemmsilva, line 57: Why do we need the stopped chan? Doesn't seem to be used.

jorgemmsilva (approved):
LGTM, just 1 question

plugins/prometheus/plugin.go
ivange94: It's used on line 62. `defer close(stopped)`. But it's effect is actually minor so I can remove it if you don't think it's necessary. So basically what it's doing is shutting down the plugin if the was an error starting the server instead of waiting for the shutdown signal to shutdown the plugin. See line 68-70.

jorgemmsilva: sure, makes sense 👍

plugins/prometheus/plugin.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/labstack/echo/v4/middleware"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const PluginName = "Prometheus"
+
+var (
+	log    *logger.Logger
+	server *http.Server
```
dessaya: Does this need to be a global var? Maybe it can be local in the background worker.

dessaya (approved):

PR #269, opened 2021-07-16
Refactor coreutil

dessaya:
This is an attempt to improve the boilerplate for native contracts' interface and implementation:

* Interface and implementation of native contracts are now separate. In theory they could be in separate packages, and external code could import the interface definition without having to import the implementation code.
* This also clears up the potential initialization cycle (interface depended on implementation and vice-versa).
* `Interface` is renamed to `Contract`. E.g., things like `root.Interface.Name` becomes `root.Contract.Name`
* The implementation of a contract is now defined entirely in `impl.go` (including the list of function handlers), and stored in an object called `Processor`. E.g., `root.Processor` is the implementation of the `root.Contract` interface.
* Function declarations in the interface are now instances of `EntryPointInfo` or `ViewEntryPointInfo` (instead of being just strings)

### DIFF
--- a/client/chainclient/uploadblob.go
+++ b/client/chainclient/uploadblob.go
@@ -37,7 +37,7 @@ func (c *Client) UploadBlob(fields dict.Dict, waspHosts []string, quorum int, op
 	blobHash := blob.MustGetBlobHash(fields)
 
 	reqTx, err := c.Post1Request(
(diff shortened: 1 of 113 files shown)


lunfardo314 (approved):

jorgemmsilva (approved):
nice!

PR #275, opened 2021-07-20
Integration tests

lmoe:
Test setup was modified to be compatible to GoShimmer. 

The following changes were made:

* Generate new seeds on each testrun
* Reuse the same address as long as possible, create a new random address when funds are too low
* Move address logic into a separate function
* Update wait times to accommodate GoShimmer response times
* Add additional cluster tests
* Update cluster configuration to use the GoShimmer configuration
* Add new arguments 
    * -goshimmer-hostname=`<string>` 
    * -goshimmer-txport=`<int>`
    * -goshimmer-use-provided-node `<bool>`
 * [BREAKING] renamed `goshimmer-provided` to `goshimmer-use-provided-node` to make it more clear what it does => Use a real GoShimmer node, or a simulated one


tools/cluster/config.go
```diff
@@ -13,7 +13,8 @@ import (
 type GoshimmerConfig struct {
 	TxStreamPort int
 	APIPort      int
-	Provided     bool
+	UseNode      bool
```
jorgemmsilva: `UseNode` is not an improvement imo, still ambiguous. `useMockedNode` or `useProvidedNode`

tools/cluster/tests/advanced_inccounter_test.go
```diff
@@ -107,24 +113,56 @@ func TestAccessNodesOnLedger(t *testing.T) {
...
+func createNewClient(t *testing.T, clu1 *cluster.Cluster, chain1 *cluster.Chain) *scclient.SCClient {
+	keyPair, _ := getOrCreateAddress(t, clu1)
+
+	client := chain1.SCClient(iscp.Hn(incCounterSCName), keyPair)
+
+	return client
+}
+
+func getOrCreateAddress(t *testing.T, clu1 *cluster.Cluster) (*ed25519.KeyPair, *ledgerstate.ED25519Address) {
+	minimalTokenAmountBeforeRequestingNewFunds := uint64(1000)
```
jorgemmsilva: could be a const, also would rename to`min` instead of `minimal`

tools/cluster/tests/advanced_inccounter_test.go
```diff
@@ -107,24 +113,56 @@ func TestAccessNodesOnLedger(t *testing.T) {
...
-	kp := wallet.KeyPair(1)
-	myAddress := ledgerstate.NewED25519Address(kp.PublicKey)
-	err = requestFunds(clu1, myAddress, "myAddress")
 	require.NoError(t, err)
 
-	myClient := chain1.SCClient(iscp.Hn(incCounterSCName), kp)
+	if funds <= minimalTokenAmountBeforeRequestingNewFunds {
+		// Requesting new token requires a new address
+
+		addressCount = rand.New(randomAddress).Uint64()
```
jorgemmsilva: seems like this is not an "address count", but rather a random "seed index", right?

jorgemmsilva (approved):
LGTM, just a few comments

dessaya (approved):

tools/cluster/tests/advanced_inccounter_test.go
lmoe: It's more an addressIndex. Is that fine?

tools/cluster/config.go
lmoe: Sure, should we keep `-goshimmer-use-node` or extend this as well like `-goshimmer-use-provided-node`?

jorgemmsilva: as you wish, I'm just making a suggestion. feel free to ignore 🙃

lmoe: If someone thinks there is room for clarification, I think that should be applied. 
If some thinks the command is too long, '-g' can be used.  :)

PR #276, opened 2021-07-20
Fix: callview error check

jorgemmsilva:


lunfardo314 (approved):

client/callview.go
```diff
@@ -33,7 +32,7 @@ func (c *WaspClient) CallView(chainID iscp.ChainID, hContract iscp.Hname, functi
 		switch {
 		case err == nil:
 			return res, err
-		case errors.Is(err, optimism.ErrStateHasBeenInvalidated):
+		case strings.Contains(err.Error(), "virtual state has been invalidated"):
```
Juliusan, line 35: Why is string comparison needed here? What's wrong with `errors.Is`?

jorgemmsilva: its on client side, the error never arrived here, so 'Is' didn't work (it now retries on both client and server side)

PR #281, opened 2021-07-21
minor adjustments after review

lunfardo314:


dessaya (approved):

packages/solo/req.go
```diff
@@ -236,13 +230,17 @@ func (ch *Chain) PostRequestSyncTx(req *CallParams, keyPair *ed25519.KeyPair) (*
...
+	return tx, res, ch.mustGetErrorFromReceipt(reqid)
+}
+
+func (ch *Chain) mustGetErrorFromReceipt(reqid iscp.RequestID) error {
 	rec, _, _, ok := ch.GetRequestReceipt(reqid)
 	require.True(ch.Env.T, ok)
-	err = nil
+	var err error
 	if len(rec.LogData) > 0 {
 		err = xerrors.New(string(rec.LogData))
```
jorgemmsilva: is `rec.LogData` always empty `""`, or an error? 
should be renamed to `rec.Error` if this is the case

jorgemmsilva (approved):

packages/solo/req.go
lunfardo314: `LogData` is not always an error. It is a `receipt`. We first have to specify what we what to have in receipt. It could contain more information than just an error message. Let's not rename until then

PR #293, opened 2021-07-25
Abstract core code from `ledgerstate.Color` and `ledgerstate.ColoredBalances`

lunfardo314:
* refactoring is a first step to decouple Wasp data structures and dependencies from Goshimmer code.  I believe Color and related stuff is the biggest dependency.
* abstracted code from `ledgerstate.Color` (almost completely, Some left in +build ignore files and in client part. Wasp now uses `colored.Color` instead
* abstracted code from `ledgerstate.ColoredBalances` (completely). Wasp now uses own type `colored.Balances` which is more convenient, I believe
* `ledegerstate.Color` and related dependencies are concentrated in the new `iscp/colored` package.
improved some parts of code, mostly related to marshaling and requests
* some cleanup of unused code


tools/wasp-cli/wallet/send.go
```diff
@@ -13,7 +15,7 @@ import (
 
 var sendFundsCmd = &cobra.Command{
 	Use:   "send-funds <target-address> <color> <amount>",
-	Short: "Transfer tokens",
+	Short: "Tokens tokens",
```
dessaya: ?

packages/vm/vmcontext/general.go
```diff
@@ -109,12 +110,14 @@ func (vmctx *VMContext) RequestID() iscp.RequestID {
 
 const maxParamSize = 512
 
-func (vmctx *VMContext) Send(target ledgerstate.Address, tokens *ledgerstate.ColoredBalances, metadata *iscp.SendMetadata, options ...iscp.SendOptions) bool {
-	if tokens == nil || tokens.Size() == 0 {
+// TODO implement send option
```
dessaya: This TODO left intentionally?

packages/vm/core/accounts/impl.go
```diff
@@ -86,35 +86,32 @@ func withdraw(ctx iscp.Sandbox) (dict.Dict, error) {
...
-	ctx.Log().Debugf("accounts.withdraw.success. Sent to address %s", tokensToSend.String())
+	ctx.Log().Debugf("accounts.withdraw.success. Sent to address %s", tokensToWithdraw.String())
 	return nil, nil
 }
 
 // owner of the chain moves all tokens from the common account to its own account
 // Params:
 //   ParamWithdrawAmount if do not exist or is 0 means withdraw all balance
-//   ParamWithdrawColor color to withdraw if amount is specified. Defaults to ledgerstate.ColorIOTA
+//   ParamWithdrawColor dummyColor to withdraw if amount is specified. Defaults to colored.IOTA
```
dessaya: dummyColor change intentional?

packages/vm/core/accounts/accounts_intern_test.go
```diff
@@ -17,9 +18,17 @@ func TestBasic(t *testing.T) {
...
-var color = ledgerstate.Color(hashing.HashStrings("dummy string"))
+var dummyColor colored.Color
+
+func init() {
+	var err error
+	dummyColor, err = colored.ColorFromBase58EncodedString(hashing.HashStrings("dummy string").Base58())
+	if err != nil {
+		panic(err)
+	}
+}
```
dessaya, line 29: Does it need to be a specific color other than the zero value?

```suggestion
const dummyColor = colored.Color{}
```

packages/iscp/colored/color.go
```diff
@@ -0,0 +1,311 @@
+package colored
```
dessaya, line 1: `colored.Color` seems a little stuttery? Suggest renaming to something like `level1` (`level1.Color`, `level1.ColoredBalances`) or `ledger` (`ledger.Color`, `ledger.ColoredBalances`).
Or even keeping it under `iscp` (`iscp.Color`, `iscp.ColoredBalances`).

packages/iscp/colored/color.go
```diff
@@ -0,0 +1,311 @@
...
+
+// Compare offers a comparator for Colors which returns -1 if otherColor is bigger, 1 if it is smaller and 0 if they are
+// the same.
+func (c Color) Compare(otherColor Color) int {
+	return bytes.Compare(c[:], otherColor[:])
+}
+
+// endregion ///////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// region Balances //////////////////////////////////////////////////////////////////////////////////////////////
```
dessaya, line 92: Any reason to prefer using regions instead of files? `color.go` + `coloredbalances.go`

dessaya (approved):

packages/vm/vmcontext/general.go
lunfardo314: yes

packages/vm/core/accounts/accounts_intern_test.go
lunfardo314: `colored.Color{}` is equal to `colored.Iota` which is bad for the test

packages/iscp/colored/color.go
lunfardo314: expected reaction :)  I chose `colored` over `color` and `level1` because I like sound of `colored.Balances`. `level1.Balances` doesn't make sense. I think `colored.IOTA` and `colored.Color` even if a (little) bit stuttery, however point to the domain of colored assets, which makes sense. 
Besides, it is not something about level 1. The concept of level 1 and 2 not clearly defined in general

packages/iscp/colored/color.go
lunfardo314: no reason other than to mimic goshimmer repo. The topic is opinionated ;)

jorgemmsilva (approved):

PR #296, opened 2021-07-27
Off ledger fix

kape1395:
TestAccessNodesOffLedger runs, at least on my machine.


kape1395:
- libp2p peering implemented.
 - rules for calculating batch timestamp changed.
 - More debug logging added.

packages/peering/udp/udpPeer.go
```diff
@@ -152,6 +152,9 @@ func (p *peer) maintenanceCheck() {
 		p.sendHandshake(true)
 	} else {
 		p.accessLock.RUnlock()
+		// if p.numUsers == 0 && p.lastMsgRecv.Before(old) {
+		// TODO: Do a peer cleanup.
+		// }
```
jorgemmsilva, line 157: commented code with a TODO, not sure if intended

packages/peering/udp/udpNetImpl.go
```diff
@@ -393,10 +394,10 @@ func (n *NetImpl) receiveUserMsg(msg *peering.PeerMessage, peerUDPAddr *net.UDPA
...
 	remoteUDPAddrStr := peerUDPAddr.String()
-	n.peersLock.RLock()
-	defer n.peersLock.RUnlock()
 
+	n.peersLock.RLock()
 	if p, ok := n.peersByAddr[remoteUDPAddrStr]; ok {
+		n.peersLock.RUnlock()
 		p.noteReceived()
 		msg.SenderNetID = p.NetID()
 		n.recvQueue <- &peering.RecvEvent{
```
jorgemmsilva, line 403: probably silly question: is it possible that this channel overflows? should we log that case?

Makefile
```diff
@@ -6,7 +6,7 @@ build:
 build-lint: build lint
 
 test: install
-	go test -tags rocksdb ./... --timeout 20m --count 1
+	go test -tags rocksdb ./... --timeout 30m --count 1
```
jorgemmsilva: we could add `-failfast` so it stops running tests as soon as there is an error
```suggestion
	go test -tags rocksdb ./... --timeout 30m --count 1 -failfast
```

packages/chain/chainimpl/interface.go
```diff
@@ -85,15 +84,36 @@ func (c *chainObj) IsDismissed() bool {
...
+		err := recover()
+		if err == "send on closed channel" {
+			c.log.Warnf("Failed to receive message, reason=%v", err)
+			return
+		}
+		if err != nil {
+			panic(err)
+		}
+	}()
 	if !c.IsDismissed() {
```
jorgemmsilva: I know this is existing code, but we could invert the condition and return early to reduce nesting:
```
	if c.IsDismissed() {
		return
	}
```

packages/chain/chainimpl/interface.go
```diff
@@ -85,15 +84,36 @@ func (c *chainObj) IsDismissed() bool {
...
+	defer func() { // This is needed to handle possible write to a closed channel.
+		err := recover()
+		if err == "send on closed channel" {
+			c.log.Warnf("Failed to receive message, reason=%v", err)
+			return
+		}
+		if err != nil {
+			panic(err)
+		}
+	}()
```
jorgemmsilva: this feels like a code smell. Does it ever happen?

we should know that the channel is closed because of `c.dismissed`, right?, or is there any other way it can be closed?

packages/peering/lpp/lppNetImpl.go
```diff
@@ -0,0 +1,549 @@
...
+	libp2pquic "github.com/libp2p/go-libp2p-quic-transport"
+	"github.com/libp2p/go-tcp-transport"
+	"github.com/multiformats/go-multiaddr"
+	"golang.org/x/xerrors"
+)
+
+const (
+	maintenancePeriod = 1 * time.Second
+	recvQueueSize     = 1024 * 5
+	// recvBlockingDuration = 3 * time.Second
```
jorgemmsilva:
```suggestion
```

packages/peering/lpp/lppNetImpl.go
```diff
@@ -0,0 +1,549 @@
...
+			addrs = append(addrs, addr)
+		}
+		for j := range peerIP6 {
+			addr, err := multiaddr.NewMultiaddr(fmt.Sprintf(addrPatterns[i], "ip6", peerIP6[j], peerPort))
+			if err != nil {
+				return libp2ppeer.ID(""), xerrors.Errorf("failed to make libp2p address for NetID=%v, error: %w", trustedPeer.NetID, err)
+			}
+			addrs = append(addrs, addr)
+		}
+	}
```
jorgemmsilva: seems a bit of duplicated logic here (for ip4/6), would it make sense to do something like this? (and we could remove the ip4/ip6 lists split)


```suggestion
func getIpVersion( ip IP) string {
	if ip.To4() != nil {
		return "ip4"
	}
	return "ip6"
}

// ---------
		for _, ip := range peerIPs {
			addr, err := multiaddr.NewMultiaddr(fmt.Sprintf(addrPatterns[i], getIpVersion(ip),ip, peerPort))
			if err != nil {
				return libp2ppeer.ID(""), xerrors.Errorf("failed to make libp2p address for NetID=%v, error: %w", trustedPeer.NetID, err)
			}
			addrs = append(addrs, addr)
		}
	}
```

packages/peering/lpp/lppPeer.go
```diff
@@ -0,0 +1,219 @@
...
+		p.accessLock.RUnlock()
+		p.net.lppHeartbeatSend(p, true)
+	} else {
+		p.accessLock.RUnlock()
+	}
+	if p.numUsers == 0 && !p.trusted && p.lastMsgRecv.Before(old) {
+		p.net.delPeer(p)
+		close(p.sendCh)
+	}
+}
```
jorgemmsilva: this conditional unlock logic is weird.
```
	p.accessLock.RLock()
	if p.numUsers > 0 && p.lastMsgRecv.Before(old) {
		p.accessLock.RUnlock()
		p.net.lppHeartbeatSend(p, true)
	} else {
		p.accessLock.RUnlock()
	}
```

shouldn't we also lock/unlock `p.accessLock` to read on line #80 ?
```
	if p.numUsers == 0 && !p.trusted && p.lastMsgRecv.Before(old) {
```

what do you think about the following?
```suggestion
func (p *peer) maintenanceCheck() {
	now := time.Now()
	old := now.Add(-inactivePingTime)
	
	p.accessLock.RLock()
	numUsers := p.numUsers
	isLastMsgOld := p.lastMsgRecv.Before(old)
	trusted := p.trusted
	p.accessLock.RUnlock()
	
	if numUsers > 0 &&  isLastMsgOld {
		p.net.lppHeartbeatSend(p, true)
	}
	if numUsers == 0 && !trusted && isLastMsgOld {
		p.net.delPeer(p)
		close(p.sendCh)
	}
}
```

packages/peering/lpp/lppPeer.go
```diff
@@ -0,0 +1,219 @@
...
+			p.log.Warnf("Failed to send message, reason=%v", err)
+			return
+		}
+		if err != nil {
+			p.log.Errorf("Failed to send message, reason=%v", err)
+			panic(err)
+		}
+	}
+	//
+	defer catch()
```
jorgemmsilva, line 128: similar to the other comment, this feels weird, shouldn't we know when the channel is closed instead of handling "send on closed chan" panics?

packages/peering/lpp/lppPeer.go
```diff
@@ -0,0 +1,219 @@
...
+// Return true if is alive and average latencyRingBuf in nanosec.
+func (p *peer) IsAlive() bool {
+	p.accessLock.RLock()
+	defer p.accessLock.RUnlock()
+	return p.remotePubKey != nil && p.lastMsgRecv.After(time.Now().Add(-inactiveDeadline))
+}
+
+// Await implements peering.PeerSender interface for the remote peers.
+func (p *peer) Await(timeout time.Duration) error {
+	if p.trusted {
```
jorgemmsilva, line 185:
```suggestion
	p.accessLock.RLock()
	defer p.accessLock.RUnlock()
	if p.trusted {
```

jorgemmsilva (approved):
left some observations that might be worth looking into, but it LGTM.
its awesome how well libp2p fits without any destructive changes needed! 🚀

packages/peering/udp/udpPeer.go
kape1395: That's in the UDP implementation. Maybe it should be removed along with the TCP implementation?
I would like to wait a bit to see if everything works fine with the libp2p.

packages/peering/udp/udpNetImpl.go
kape1395: Again, that's in the based UDP implementation.
The overflow is possible. This blocks the receive queue for some time.
During such periods the incoming packets are either buffered by the OS, or dropped.
I would like to leave it as is and later remove UDP and TCP based peering implementations.

packages/chain/chainimpl/interface.go
kape1395: Good point. Done.

packages/chain/chainimpl/interface.go
kape1395: Tried to use dismissed. But the errors are apparent in high load tests.
The scenario is like this:

  - Thread 1 checks if chain is  dismissed, gets false.
  - Thread 2 marks chain dismissed and closes the channel.
  - Thread 1 attempts to send a message to the channel and then gets an error.

Another approach would be to use locks, but I don't want to use locks on channel operations, because those can block and leave the locks taken for a long time.

I'm not happy with that recover() solution either, but I don't know better approaches to that,

packages/peering/lpp/lppNetImpl.go
kape1395: Updated.

packages/peering/lpp/lppPeer.go
kape1395: Nice pattern. Will use it.

packages/peering/lpp/lppPeer.go
kape1395: The comment is the same as above.

packages/peering/lpp/lppPeer.go
kape1395: Good catch.

packages/chain/chainimpl/interface.go
lunfardo314: use lock with timeout? That would be more expensive of course

dessaya: Probably the way to go is making sure that the thread that writes to the channel is the one that closes it.

```go
ch := make(chan string)
done := make(chan bool)

// thread 2
func dismiss() {
  close(done)
}

// thread 1
func doStuff() {
  for {
    select {
      case <- done:
        close(ch)
        break
      default:
        ch <- "hi"
    }
  }
}
```

PR #298, opened 2021-07-27
Fix: reboot with db

jorgemmsilva:
fix node crash when we reboot with and existing database:
 - fixes race condition in goshimmer node connection and chain activation
 - fixes statemgr solid state fetched from DB not being applied

### DIFF
--- a/packages/chain/chain.go
+++ b/packages/chain/chain.go
@@ -31,6 +31,7 @@ type ChainCore interface {
 	Processors() *processors.Cache
 	GlobalStateSync() coreutil.ChainStateSync
 	GetStateReader() state.OptimisticStateReader
+	SetGlobalStateIndex(index uint32)
 	Log() *logger.Logger
 }
 
--- a/packages/chain/chain.go
+++ b/packages/chain/chain.go
@@ -31,7 +31,6 @@ type ChainCore interface {
 	Processors() *processors.Cache
 	GlobalStateSync() coreutil.ChainStateSync
 	GetStateReader() state.OptimisticStateReader
-	SetGlobalStateIndex(index uint32)
 	Log() *logger.Logger
 }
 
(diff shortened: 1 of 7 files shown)


packages/chain/chain.go
```diff
@@ -31,6 +31,7 @@ type ChainCore interface {
 	Processors() *processors.Cache
 	GlobalStateSync() coreutil.ChainStateSync
 	GetStateReader() state.OptimisticStateReader
+	SetGlobalStateIndex(index uint32)
```
dessaya: Isn't this the same as `GlobalStateSync().SetSolidIndex(index)`? Does it need to be a separate method?

dessaya (approved):

plugins/chains/plugin.go
```diff
@@ -44,11 +44,11 @@ func run(_ *node.Plugin) {
 		database.GetOrCreateKVStore,
 	)
 	err := daemon.BackgroundWorker(PluginName, func(shutdownSignal <-chan struct{}) {
+		allChains.Attach(nodeconn.NodeConnection())
```
lunfardo314, line 47: `nodeconn.NodeConnection()` is waiting untill nodconn plugin is ready. No need dor additional channel to sync the readiness,

packages/chains/chains.go
```diff
@@ -113,6 +116,7 @@ func (c *Chains) ActivateAllFromRegistry(registryProvider registry.Provider) err
 // - insert it into the runtime registry
 // - subscribes for related transactions in he IOTA node
 func (c *Chains) Activate(chr *registry.ChainRecord, registryProvider registry.Provider) error {
+	<-c.nodeConnEstablished
```
lunfardo314: No need. See comment below

lunfardo314 (approved):

PR #300, opened 2021-07-27
improvement: Protection against basic spam attacks

jorgemmsilva:
a few strategies implemented to prevent abuse of resources.

root contract now has specific limits for the chain:
- `MaxEventsPerRequest`  - default 1000
- `DefaultMaxEventSize` - default 2000   // 2Kb
- `DefaultMaxBlobSize` - default 2000000 // 2Mb

offledger requests API now saves an expiring cache of processed requests (5 min by default, but can be configured using config.json)

--- 

root contract implementation was moved to `rootimpl`, so that root interface exported constants can be used on other contracts without causing an import cycle.


packages/vm/vmcontext/internal.go
```diff
@@ -145,11 +147,16 @@ func (vmctx *VMContext) mustLogRequestToBlockLog(errProvided error) {
...
+		vmctx.Panicf("too many events issued for contract: %s, request index: %d", contract.String(), vmctx.requestIndex)
+	}
 
 	vmctx.log.Debugf("MustSaveEvent/%s: msg: '%s'", contract.String(), msg)
 	err := blocklog.SaveEvent(vmctx.State(), msg, vmctx.eventLookupKey(), contract)
 	if err != nil {
 		vmctx.Panicf("MustSaveEvent: %v", err)
 	}
-	vmctx.requestEventIndex++
+	if vmctx.requestEventIndex < math.MaxUint8 {
```
dessaya: This should always be true? Then I think it would be better to panic if it's false (like above) or remove the `if`.

dessaya (approved):

packages/vm/vmcontext/internal.go
jorgemmsilva: I think it overflows. In my test after 255 it starts writing from 0

jorgemmsilva: nevermind, I see what you mean 👍

dessaya: But if I understand correctly, you already added a check above so it should be impossible to overflow now?

lunfardo314:
as per latest discussion:
- to add max limit on event data size (panics if exceeds)
- (probably) to return to `uint16` for max number of events in the request. 
- Optionally: config parameters `MaxEvents`, `MaxEventSize`, `MaxBlobSize` at the chain level. With reasonable defaults. Updatable by the chain owner

tools/wasp-cli/chain/listcontracts.go
```diff
@@ -17,10 +18,10 @@ var listContractsCmd = &cobra.Command{
 		info, err := SCClient(root.Contract.Hname()).CallView(root.FuncGetChainInfo.Name, nil)
 		log.Check(err)
 
-		contracts, err := root.DecodeContractRegistry(collections.NewMapReadOnly(info, root.VarContractRegistry))
+		contracts, err := rootimpl.DecodeContractRegistry(collections.NewMapReadOnly(info, root.VarContractRegistry))
```
dessaya: Shouldn't public data types and encode/decode functions be in the `root` package instead of `rootimpl`?

I mean, I understand the reason for the split is to have `rootimpl` contain only the private implementation of the contract, right? So `rootimpl` should never be imported from outside.

jorgemmsilva: you're right. will refactor 👍

PR #311, opened 2021-07-29
Expose metrics for counting requests per chain

ivange94:
Mempool metrics per chain.


packages/chain/mempool/mempool.go
```diff
@@ -184,6 +198,7 @@ func (m *Mempool) RemoveRequests(reqs ...iscp.RequestID) {
 			continue
 		}
 		m.outPoolCounter++
+		m.mempoolMetrics.ProcessRequest()
```
lunfardo314: maybe names like `CountOffLedgerRequestIn`, `CountOnLedgerRequestIn`, `CountRequestOut` are more self describing?

packages/metrics/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+func (c *chainMetrics) ProcessRequest() {
+	c.metrics.processedRequestCounter.With(prometheus.Labels{"chain": c.chainID.String()}).Inc()
+}
+
+func (m *Metrics) registerMempoolMetrics() {
+	m.log.Info("Registering mempool metrics to prometheus")
+	m.offLedgerRequestCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "wasp_off_ledger_request_counter",
+		Help: "Number of ledger requests made to chains",
+	}, []string{"chain"})
```
lunfardo314, line 31: typo: "Number of off-ledger requests made to chains"

packages/metrics/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+	prometheus.MustRegister(m.offLedgerRequestCounter)
+
+	m.onLedgerRequestCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "wasp_on_ledger_request_counter",
+		Help: "Number of on ledger requests made to chain",
+	}, []string{"chain"})
```
lunfardo314, line 37: on-ledger

packages/metrics/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+	log *logger.Logger
+}
+
+func DefaultMempoolMetrics(log *logger.Logger) MempoolMetrics {
+	return &defaultMempoolMetrics{log}
+}
+
+func (m *defaultMempoolMetrics) NewOffLedgerRequest() {
+	m.log.Info("No counter configured for off ledger requests.")
+}
```
lunfardo314: logging is temporary, right?

packages/parameters/parameters.go
```diff
@@ -48,8 +48,8 @@ const (
 	ProfilingBindAddress = "profiling.bindAddress"
 	ProfilingEnabled     = "profiling.enabled"
 
-	PrometheusBindAddress = "prometheus.bindAddress"
-	PrometheusEnabled     = "prometheus.enabled"
```
lunfardo314: it seems use of `prometheus` makes sense here because server exposes metrics specifically for prometheus through this bind address

lunfardo314 (approved):

packages/chain/mempool/mempool.go
ivange94: I agree. Will update.

packages/metrics/mempool.go
ivange94: yeah that's temporary.

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/pangpanglabs/echoswagger/v2 v2.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.7.0
+	github.com/shirou/gopsutil v3.21.4-0.20210419000835-c7a38de76ee5+incompatible
```
dessaya: I don't see this used anywhere? Is it really a dependency?

packages/metrics/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+type defaultMempoolMetrics struct {
+	log *logger.Logger
+}
+
+func DefaultMempoolMetrics(log *logger.Logger) MempoolMetrics {
+	return &defaultMempoolMetrics{log}
+}
+
+func (m *defaultMempoolMetrics) CountOffLedgerRequestIn() {
+	m.log.Info("No counter configured for off ledger requests.")
```
dessaya: Maybe this should use `log.Warn` or `log.Debug` instead? The logs are already too spammy :sweat_smile:

packages/metrics/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type MempoolMetrics interface {
+	CountOffLedgerRequestIn()
+	CountOnLedgerRequestIn()
+	CountRequestOut()
+}
+
+func (c *chainMetrics) CountOffLedgerRequestIn() {
```
dessaya, line 14: Organization of structs and methods is a bit confusing IMO. I'd expect to have the declaration of `chainMetrics` and all its methods in a single file / location? Same for `Metrics` and all other structs. Is there a logic that I'm missing?

packages/webapi/admapi/activatechain.go
```diff
@@ -7,15 +7,16 @@ import (
 	"github.com/iotaledger/goshimmer/packages/ledgerstate"
 	"github.com/iotaledger/wasp/packages/chains"
 	"github.com/iotaledger/wasp/packages/iscp"
+	metricspkg "github.com/iotaledger/wasp/packages/metrics"
```
dessaya: Is the rename really necessary?

plugins/metrics/plugin.go
```diff
@@ -0,0 +1,67 @@
...
+			log.Errorf("Error stopping: %s", err)
+		}
+	}, parameters.PriorityMetrics); err != nil {
+		log.Warnf("Error starting as daemon: %s", err)
+	}
+}
+
+func AllMetrics() *metrics.Metrics {
+	initialized.MustWait(5 * time.Second)
+	return allMetrics
```
dessaya: Is the initializer mutex really necessary?

The `allMetrics` variable is set in the `configure` phase, so as long as no one calls `AllMetrics()` before the `configure` phase is finished, this should always return a valid object.

Even if `AllMetrics()` is needed during the configure phase, there is no need to wait for the metrics server to be ready; the `initialized.SetReady()` can be moved to the `configure` function. Unless I'm missing something?

packages/metrics/mempool.go
ivange94: Yeah I tried to make the organisation as easy to follow as could. I realised having method definitions on `chainMetrics` inside mempool.go can be very confusing. But `MempoolMetrics` is just a sub interface of ChainMetrics so it uses the same struct. And even the prometheus metrics variables here are specific to mempool. In future I may have a concensus.go, statemanager.go with metrics specific for those areas all of which will still define methods on `*chainMetrics` reason why I split it this way.

packages/webapi/admapi/activatechain.go
ivange94: Maybe, maybe not. I could have as well renamed my variable. I did create a variable name `metrics` and this was causing problems with the package import.

plugins/metrics/plugin.go
ivange94: Thanks for this clarification. Yes you're right. Given that no code calls the `AllMetrics()` before and within the configure stage, I think it's safe to remove the `initialized` usage here.

dessaya (approved):

packages/chain/mempool/mempool.go
```diff
@@ -163,6 +174,9 @@ func (m *Mempool) ReceiveRequest(req iscp.Request) bool {
 	if m.checkInBuffer(req) {
 		return false
 	}
+	if req.IsOffLedger() {
+		m.mempoolMetrics.CountOffLedgerRequestIn()
+	}
```
jorgemmsilva: this is currently only used for offledger requests, so the check could be potentially removed (and probably renamed it to `receiveOffledgerRequest`).

packages/metrics/mempool.go
jorgemmsilva: Actually, should be a single chain, right? 😄 
```
"Number of off-ledger requests made to the chain",
```

packages/metrics/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+
+	m.onLedgerRequestCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "wasp_on_ledger_request_counter",
+		Help: "Number of on-ledger requests made to chain",
+	}, []string{"chain"})
+	prometheus.MustRegister(m.onLedgerRequestCounter)
+
+	m.processedRequestCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "wasp_processed_on_ledger_request_counter",
+		Help: "Number of requests processed on ledger",
```
jorgemmsilva:
```suggestion
		Help: "Number of on-ledger requests processed",
```

packages/metrics/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+	m.log.Debug("No counter configured for off ledger requests.")
+}
+
+func (m *defaultMempoolMetrics) CountOnLedgerRequestIn() {
+	m.log.Debug("No counter configured for on ledger requests.")
+}
+
+func (m *defaultMempoolMetrics) CountRequestOut() {
+	m.log.Debug("No counter configured for processed requests.")
+}
```
jorgemmsilva: I would avoid all this logging, doesn't make much sense imo, imo we can just log something saying "metrics not configured" when the `DefaultMempoolMetrics` gets created.

packages/metrics/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+		e.GET("/metrics", func(c echo.Context) error {
+			handler := promhttp.Handler()
+			handler.ServeHTTP(c.Response(), c.Request())
+			return nil
+		})
+		m.server = &http.Server{Addr: addr, Handler: e}
+	}
+	m.registerMempoolMetrics()
+	m.log.Infof("Prometheus metrics accessible at: %s", addr)
+	return m.server.ListenAndServe()
```
jorgemmsilva: could use `sync.once.Do`. Even with the nil check there is a possibility of a race condition if Start called multiple times

packages/webapi/admapi/activatechain.go
jorgemmsilva: I would just rename the variable in this case

jorgemmsilva (approved):
nice 🚀  just some minor comments

packages/metrics/mempool.go
ivange94: How do you suggest we handle the default implementation of these methods then? I need a default implementation of the MempoolMetrics Interface so I don't have to do `if mempoolMetrics == nil` before calling those methods here https://github.com/iotaledger/wasp/pull/311/files#diff-68311bbde608a0ded53c2bcfc06ef2533845ce5fe1a1bbe9b06600da61b7fc04R161 as that adds another layer of nesting. Personally I don't like those logging either but I wasn't sure how else to provide a default implementation. Having empty methods felt weird but I can put that instead. What do you think?

ivange94: The default behaviour in my opinion is to not have any metrics exposed as it might be an unnecessary overhead when prometheus is turned off.

packages/metrics/metrics.go
ivange94: About this, `once.Do` takes an argument of type `func()` meanwhile `ListenAndServe` has type `func() error`. So to get this to work I had to wrap `ListenAndServe` inside a function and then pass that function to `once.Do`. But that hides the error.

packages/metrics/mempool.go
jorgemmsilva: I think it should just be empty functions. Logging a million times "metrics not configured" won't add any new information 😄

packages/metrics/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+	m.registerMempoolMetrics()
+	m.log.Infof("Prometheus metrics accessible at: %s", addr)
+	var once sync.Once
+	runServer := func() {
+		if err := m.server.ListenAndServe(); err != nil {
+			m.log.Error("Failed to start metrics server", err)
+		}
+	}
+	once.Do(runServer)
+}
```
jorgemmsilva: I think the current "once.Do" impl won't work (`var once sync.Once` gets re-declared any time the function runs).

What do you think about the following?

```suggestion
var once sync.Once

func (m *Metrics) Start(addr string) {
	once.Do(func() {
		e := echo.New()
		e.HideBanner = true
		e.Use(middleware.Recover())
		e.GET("/metrics", func(c echo.Context) error {
			handler := promhttp.Handler()
			handler.ServeHTTP(c.Response(), c.Request())
			return nil
		})
		m.registerMempoolMetrics()
		m.log.Infof("Prometheus metrics accessible at: %s", addr)
		m.server = &http.Server{Addr: addr, Handler: e}
		if err := m.server.ListenAndServe(); err != nil {
			m.log.Error("Failed to start metrics server", err)
		}
	})
}
```

ivange94: This makes sense.

PR #312, opened 2021-07-29
Improve performance of EVM emulator

dessaya:
This PR fixes the quadratic behavior of EVMEmulator when creating blocks with multiple transactions.

With this change, the `BenchmarkEVMStorageAsync` benchmark goes from ~250 TPS to ~750 TPS.

### DIFF
--- a/packages/evm/emulator.go
+++ b/packages/evm/emulator.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/common/math"
+	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/ethash"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/bloombits"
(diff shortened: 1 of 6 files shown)


packages/evm/emulator.go
```diff
@@ -113,28 +137,65 @@ func (e *EVMEmulator) Close() error {
...
+		e.pending.header,
+		e.pending.state,
+		e.pending.txs,
+		nil,
+		e.pending.receipts,
+	)
+	if err != nil {
+		panic(err)
+	}
+	root, err := e.pending.state.Commit(e.blockchain.Config().IsEIP158(e.pending.header.Number))
```
jorgemmsilva: do we need the `.IsEIP158` stuff?
edit: we can probably just pass `true`

packages/evm/emulator.go
```diff
@@ -113,28 +137,65 @@ func (e *EVMEmulator) Close() error {
...
-	e.pendingState, _ = state.New(e.pendingBlock.Root(), e.blockchain.StateCache(), nil)
+func (e *EVMEmulator) Rollback(timestamp uint64) {
+	statedb, err := e.blockchain.State()
+	if err != nil {
+		panic(err)
+	}
+
+	parent := e.blockchain.CurrentBlock()
+	header := &types.Header{
+		Root:       statedb.IntermediateRoot(e.blockchain.Config().IsEIP158(parent.Number())),
```
jorgemmsilva: I think we can just pass `true` here
```suggestion
		Root:       statedb.IntermediateRoot(true),
```

jorgemmsilva (approved):

PR #313, opened 2021-07-30
put previous state hash into the block

lunfardo314:
- Each block of mutations to the state now contains the previous state hash. This creates chain of blocks and makes it possible to use state hash as key for the block in the DB. This, in turn, makes it possible to validate whole chain of blocks knowing only the starting state and the final state hash fetched from the ledger. The rest can be fetched from the DB without access to the Tangle
- `blocklog` now contains whole sequence state hashes as a part of the `BlockInfo` record


packages/vm/core/blocklog/interface.go
```diff
@@ -119,7 +128,13 @@ func (bi *BlockInfo) Read(r io.Reader) error {
...
 		return err
 	}
-	return util.ReadUint16(r, &bi.NumOffLedgerRequests)
+	if err := util.ReadUint16(r, &bi.NumOffLedgerRequests); err != nil {
+		return err
+	}
+	if err := util.ReadHashValue(r, &bi.PreviousStateHash); err != nil {
+		return err
+	}
+	return nil
```
jorgemmsilva:
```suggestion
	return util.ReadHashValue(r, &bi.PreviousStateHash)
```

lunfardo314: haha, I corrected this way on purpose. Before it was like you corrected. Because it makes a lot easier the editing and so on. That is why I am not in very good relations with the linter

packages/state/stateupdate.go
```diff
@@ -168,10 +176,34 @@ func findTimestampMutation(stateUpdates []*StateUpdateImpl) (time.Time, error) {
...
+	}
+	for i := len(stateUpdates) - 1; i >= 0; i-- {
+		hashBin, exists := stateUpdates[i].Mutations().Get(kv.Key(coreutil.StatePrefixPrevStateHash))
+		if !exists || hashBin == nil {
+			continue
+		}
+		prevStateHash, ok, err := codec.DecodeHashValue(hashBin)
+		if !ok || err != nil {
+			return hashing.NilHash, err
+		}
```
jorgemmsilva: could we use. `su.PreviousStateHashMutation()` here?

jorgemmsilva (approved):

dessaya (approved):

Issue #310, opened 2021-07-29
Deploying to evm fails with invalid memory address


dahof1:
Branch: Develop

Deploying a Smart Contract to EVM Chain fails with the following error:
```
GET http://127.0.0.1:9090/chain/mbzESjyjzkzbLEcs1JeXKYzT4gpNLA41NQqbkxSBq2Rj/contract/adc164b5/callview/getBlockByNumber: 400: View call failed: viewcontext: panic in VM: runtime error: invalid memory address or nil pointer dereference
```

dessaya:
@dahof1 I can't reproduce this issue. Can you provide more details? How are you deploying the contract? What are the exact steps? Is there some log you can share?

dahof1:
Hi @dessaya,

it is strange... so here are the (complete) steps:

(Pre-step: `wasp-cluster start`)

```bash
wasp-cli request-funds
wasp-cli chain deploy --committee=0,1,2,3 --quorum=3 --chain=mychain --description="My chain"
wasp-cli chain deposit IOTA:1000
wasp-cli chain evm deploy --chainid 1074 --alloc "<ADDRESS>:100000000000000000000000000"
wasp-cli chain evm jsonrpc --chainid 1074
````

Deploy a smart contract with truffle:
```
truffle migrate --network wasptest
```

The migration fails with: (At this time i had a privateKey as deployer key)
```
GET http://127.0.0.1:9090/chain/mbzESjyjzkzbLEcs1JeXKYzT4gpNLA41NQqbkxSBq2Rj/contract/adc164b5/callview/getBlockByNumber: 400: View call failed: viewcontext: panic in VM: runtime error: invalid memory address or nil pointer dereference
```

Update 1:

Then, i switched  to mnemonic as hd wallet key and afterwards the error was gone. BUT now the deployment runs like "forever". So it feels like there is no block confirmation coming. In the log i see a lot of `/callview/getBlockByNumber` calls, but the deployment runs...

My truffle configuration:

```javascript
networks: {
wasptest01 : {
      provider: () => new HDWalletProvider(process.env.WASPTEST01_DEPLOYER_KEY, process.env.WASPTEST01_RPC),
      network_id: 1074,
      skipDryRun: true,
      confirmations: 0,
      timeoutBlocks: 0,
      from: <address>
    },
}
```

Update 2:

At some time (back and forth of trying :) ) i receive this error
```
[...]

PR #314, opened 2021-07-30
Misc fixes

dessaya:
Two fixes to address issue #310

- Add stack trace of failed view calls to debug log
- After posting a request, `wasp-cli` shows a hint to the user to run `wasp-cli chain request <request-id>` to see the result of the request.

### DIFF
--- a/tools/cluster/tests/wasp-cli_test.go
+++ b/tools/cluster/tests/wasp-cli_test.go
@@ -139,6 +139,17 @@ func TestWaspCLIContract(t *testing.T) {
 	checkCounter(45)
 }
 
+func findRequestIDInOutput(out []string) string {
+	for _, line := range out {
+		m := regexp.MustCompile(`(?m)#\d+ \(check result with: wasp-cli chain request (\w+)\)$`).FindStringSubmatch(line)
+		if len(m) == 0 {
+			continue
+		}
+		return m[1]
+	}
+	return ""
+}
+
 func TestWaspCLIBlockLog(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ivange94 (approved):

tools/cluster/tests/wasp-cli_test.go
```diff
@@ -139,6 +139,17 @@ func TestWaspCLIContract(t *testing.T) {
...
+func findRequestIDInOutput(out []string) string {
+	for _, line := range out {
+		m := regexp.MustCompile(`(?m)#\d+ \(check result with: wasp-cli chain request (\w+)\)$`).FindStringSubmatch(line)
+		if len(m) == 0 {
+			continue
+		}
+		return m[1]
+	}
+	return ""
+}
```
jorgemmsilva, line 151: regexp could be created once and re-used

```suggestion
var rxp = regexp.MustCompile(`(?m)#\d+ \(check result with: wasp-cli chain request (\w+)\)$`)
func findRequestIDInOutput(out []string) string {
	for _, line := range out {
		m := rxp.FindStringSubmatch(line)
		if len(m) == 0 {
			continue
		}
		return m[1]
	}
	return ""
}
```

jorgemmsilva (approved):

tools/cluster/tests/wasp-cli_test.go
dessaya: True, but this being test code I'd argue that the performance gain is not important. Personally I prefer to leave it as it is.

PR #315, opened 2021-08-02
Cluster stability integration tests

lmoe:
This PR will introduce a new set of tests which validate the stability and consensus capabilities of a Wasp cluster.

This includes:
* Sudden crashes of a fixed set of nodes while smart contracts are being invoked
* Freeze and unfreezing of nodes (simulating network issues, bottlenecking)
* Validating the quorum limit and testing the handling below that
* Combining a lot requests with heavy cluster instability


packages/peering/domain/domain.go
```diff
@@ -120,6 +120,8 @@ func (d *DomainImpl) AddPeer(netID string) error {
 	}
 	d.nodes[netID] = peer
 	d.reshufflePeers()
+	util.LogExtended(d.log, util.ExtendedMessage{Message: "Added Peer", Tag: "domain", Args: map[string]string{"netID": netID}})
```
dessaya: I think this would produce a result similar to `log.Debugw("Added Peer", "tag", "domain", "args", map[string]string{"netID": netID}})`. In that case, no need for `LogExtended`?

packages/webapi/admapi/peering.go
```diff
@@ -47,6 +47,10 @@ func addPeeringEndpoints(adm echoswagger.ApiGroup, network peering.NetworkProvid
 		AddResponse(http.StatusOK, "Trusted peer info.", listExample[0], nil).
 		SetSummary("Trust the specified peer, the pub key is passed via the path.")
 
+	adm.GET(routes.PeeringGetStatus(), handlePeeringGetStatus, addCtx).
+		AddResponse(http.StatusOK, "This node as a peer.", listExample[0], nil).
```
dessaya: `listExample[0]` suggests that the response value is of type `*model.PeeringTrustedNode`, but the implementation below returns a list of `model.PeeringNodeStatus`.

Note: the information here is used to autogenerate the API documentation, accessible at the `/doc` endpoint.

tools/cluster/cluster.go
```diff
@@ -309,6 +325,71 @@ func (clu *Cluster) start(dataPath string) error {
...
+			if err == nil {
+				clu.waspCmds[nodeIndex] = nil
+			}
+
+			return err
+		}
+
+		return nil
+	}
+	return fmt.Errorf("[cluster] Wasp node with index %d not found", nodeIndex)
```
dessaya: It's more idiomatic to return early and avoid indentation levels. Same for the other functions below. Instead of 

```
if condition {
  do stuff
}
return error
```

it's often better:

```
if !condition {
  return error
}
do stuff
```


```suggestion
	if nodeIndex >= len(clu.waspCmds) {
		return fmt.Errorf("[cluster] Wasp node with index %d not found", nodeIndex)
	}

	process := clu.waspCmds[nodeIndex]
	if process == nil {
		return nil
	}

	err := process.Process.Kill()
	if err != nil {
		return err
	}

	clu.waspCmds[nodeIndex] = nil
```

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,347 @@
...
+	t.Run("cluster=1,numValidators=1,req=35", func(t *testing.T) {
+		const numRequests = 35
+		const numValidators = 1
+		const clusterSize = 1
+
+		env := InitializeStabilityTest(t, numValidators, clusterSize)
+		env.sendRequests(numRequests)
+		waitUntil(t, env.chainEnv.counterEquals(numRequests), env.getActiveNodeList(), 120*time.Second, "contract to be deployed")
+	})
+
```
dessaya: This block seems to be duplicated with different parameters? Can it be generalized? It could be in a local anonymous function:

```
runTest := func(numRequests, numValidators, clusterSize int) {
  t.Run(fmt.Sprintf("cluster=%d,numValidators=%d,req=%d", numRequests, numValidators, clusterSize), func(t *testing.T) {
    ...
  }
}

runTest(35, 1, 1)
runTest(35, 4, 5)
...
```

tools/cluster/tests/env.go
```diff
@@ -95,6 +99,40 @@ func (e *chainEnv) deployContract(wasmName, scDescription string, initParams map
...
+	myAddress := ledgerstate.NewED25519Address(keyPair.PublicKey)
+
+	funds, err := e.clu.GoshimmerClient().BalanceIOTA(myAddress)
+
+	require.NoError(e.t, err)
+
+	if funds <= minTokenAmountBeforeRequestingNewFunds {
+		// Requesting new token requires a new address
+
+		e.addressIndex = rand.New(randomAddress).Uint64()
```
dessaya, line 123: Just curious: why random instead of just increasing the previous index?

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,347 @@
...
+		env := InitializeStabilityTest(t, numValidators, clusterSize)
+		env.setSabotageValidators(numBrokenNodes)
+
+		wg := env.sabotageNodes(5*time.Second, 1*time.Second)
+		env.sendRequests(numRequests)
+
+		wg.Wait()
+		// quorum is not met, incCounter should not equal numRequests
+		waitUntil(t, env.chainEnv.counterNotEquals(numRequests), env.getActiveNodeList(), 120*time.Second, "contract to be deployed")
```
dessaya: I don't understand something here: does it make sense to use `waitUntil` with `counterNotEquals`? I mean, this polls repeatedly until the counter is found to be not equal to `numRequests`. I think this would make sense if the counter is expected to be equal to `numRequests` and to change some time in the future, but it is not this case, right? Here we expect the counter to be not equal right now, so why bother with `waitUntil`?

packages/peering/domain/domain.go
lmoe: Currently I'm just experimenting how to get proper data from the cluster nodes into the tests as I want to see how many nodes are actually used. This will most likely completely be replaced by https://github.com/iotaledger/wasp/pull/315/files#diff-314b0da5b7ddafa17c62ac1b35ad6f188c9bee10209d1e898aaf639e8b506c23R50 before I change the PR draft status and therefore LogExtended can also get removed.

tools/cluster/tests/cluster_stability_test.go
lmoe: When I'm done with all test cases I will put them all into generalized functions. No worries. :)

tools/cluster/tests/env.go
lmoe: That was basically one of my first commits, so I might was wrong. 
Just increasing the counter by one didn't always work, especially when you were running all tests at once. It seemed like that some tests were using an old, used address even though it should have been a new one. Maybe due to the way the cluster gets set up? Some thread synchronization issue? All I know is that using a random number for each renew did the trick. Maybe worth to take a deeper look into?

tools/cluster/tests/cluster_stability_test.go
lmoe: You are completely right. What I wanted to do is to poll the incCounter for x minutes and check that not all requests went through (As the quorum is not reached). But just negating the check does not make any sense.

tools/cluster/cluster.go
```diff
@@ -308,6 +324,73 @@ func (clu *Cluster) start(dataPath string) error {
...
+
+		if err == nil {
+			clu.waspCmds[nodeIndex] = nil
+		}
+
+		return err
+	}
+
+	return nil
+}
```
jorgemmsilva: we have a `stopNode` function, any reason to not use that?

https://github.com/iotaledger/wasp/blob/develop/tools/cluster/cluster.go#L386-L395

lmoe: Yeah, that stops the node gracefully. I've thought we need a bit destruction here.

tools/cluster/cluster.go
```diff
@@ -308,6 +324,73 @@ func (clu *Cluster) start(dataPath string) error {
...
+
+	cmd, err := clu.startServer("wasp", waspNodeDataPath(clu.DataPath, nodeIndex), nodeIndex, initOk)
+	if err != nil {
+		return err
+	}
+
+	select {
+	case <-initOk:
+	case <-time.After(10 * time.Second):
+		return fmt.Errorf("Timeout starting wasp nodes\n") //nolint:revive
```
jorgemmsilva: I think `xerrors` is prefered (instead of fmt.Errorf)
```suggestion
		return xerrors.Errorf("Timeout starting wasp nodes\n")
```

lmoe: Everywhere or just in such functions? Then I replace that in the whole cluster class.

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
+/**
+This test will test the recovery capabilities of the clusterized nodes, where some nodes can break/restart/lay dead at any time.
+*/
+
+package tests
```
jorgemmsilva, line 5: tests that use freeze/unfreeze should probably be skipped on windows:
```go
if runtime.GOOS == "windows" {
	t.Skip()
}
```

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+	from := clusterSize - e.NumValidators
+	to := from + breakCount - 1
+
+	for i := from; i <= to; i++ {
+		nodeList = append(nodeList, i)
+	}
+
+	e.From = from
+	e.To = to
```
jorgemmsilva: `from`, `to` seem to not be used and could be removed

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+	e.SabotageList = nodeList
+}
```
jorgemmsilva: I think both `setSabotageValidators` and `setSabotageAll` could be deprecated in favour of just passing a `[]int` list of nodeIndexes to sabotage. Unless I'm missing something.

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+type SabotageOption int
+
+const (
+	SabotageByKill SabotageOption = iota
+	// Important: Frozen nodes need to get killed/unfrozen manually after the test is done, otherwise they stay alive after the test has finished
+	SabotageByFreeze SabotageOption = iota
+)
```
jorgemmsilva, line 74:
```suggestion
const (
	SabotageByKill SabotageOption = iota
	// Important: Frozen nodes need to get killed/unfrozen manually after the test is done, otherwise they stay alive after the test has finished
	SabotageByFreeze SabotageOption
)
```

both have the same value, right? 😅

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+		runTestSuccessfulIncCounterIncreaseWithoutInstability(t, clusterSize, numValidators, numRequests)
+	})
+}
+
+func runTestSuccessfulIncCounterIncreaseWithMildInstability(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequests int) {
+	env := InitializeStabilityTest(t, numValidators, clusterSize)
+	env.setSabotageValidators(numBrokenNodes)
+
+	wg := env.sabotageNodes(SabotageByKill, 4*time.Second, 1*time.Second)
```
jorgemmsilva, line 181: any reason to have `setSabotage` and `sabotage` in different steps?
`env.sabotageNodes` could receive the list of nodes to sabotage directly

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+	env.sendRequests(numRequests, time.Millisecond*250)
+
+	wg.Wait()
```
jorgemmsilva, line 184: shouldn't we send the requests *after* the nodes are sabotaged? 😅 
```suggestion
	wg.Wait()
	env.sendRequests(numRequests, time.Millisecond*250)
```

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+	t.Run("cluster=8,numValidators=6,req=35", func(t *testing.T) {
+		const clusterSize = 8
+		const numValidators = 6
+		const numRequests = 35
+
+		runTestSuccessfulIncCounterIncreaseWithoutInstability(t, clusterSize, numValidators, numRequests)
+	})
+}
+
+func runTestSuccessfulIncCounterIncreaseWithMildInstability(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequests int) {
```
jorgemmsilva: a bit verbose 😄 
```suggestion
func testIncCounterWithMildInstability(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequests int) {
```

tools/cluster/cluster.go
```diff
@@ -308,6 +324,73 @@ func (clu *Cluster) start(dataPath string) error {
...
+func (clu *Cluster) FreezeNode(nodeIndex int) error {
+	if nodeIndex >= len(clu.waspCmds) {
+		return fmt.Errorf("[cluster] Wasp node with index %d not found, active processes: %v", nodeIndex, len(clu.waspCmds))
+	}
+
+	process := clu.waspCmds[nodeIndex]
+
+	err := process.Process.Signal(syscall.SIGSTOP)
+
+	return err
```
dessaya, line 356: Nit-picking: `return process.Process.Signal(syscall.SIGSTOP)`

tools/cluster/cluster.go
jorgemmsilva: I think everywhere, but there are still a low of `fmt.Errorf` across the repo. I'm not sure what's the reason for this decision, just that's what we're using 😄 .
cc: @dessaya

tools/cluster/cluster.go
jorgemmsilva: ahah I see 😄 
fair enough, but there might be cases where the node won't come up again due to corrupt DB or similar. I guess we also need to be able to resist such cases 👍

tools/cluster/tests/env.go
dessaya: AFAIK all cluster tests run synchronously in a single thread (unless you explicitly start a goroutine), so I think it would be a godd idea to look into that, because it could reveal a more serious issue.

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+		const clusterSize = 14
+		const numValidators = 12
+		const numBrokenNodes = 11
+		const numRequests = 35
+
+		runTestFailsIncCounterIncreaseAsQuorumNotMet(t, clusterSize, numValidators, numBrokenNodes, numRequests)
+	})
+}
+
+func runTestSuccessfulConsenseusWithReconnectingNodesAndNoQuorum(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequestsBeforeFailure, numRequestsAfterFailure int) {
```
jorgemmsilva:
```suggestion
func testConsenseusReconnectingNodesNoQuorum(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequestsBeforeFailure, numRequestsAfterFailure int) {
```

tools/cluster/cluster.go
dessaya: It's preferred because when the error is printed, using `xerrors` will print the file:line info and all wrapped errors (when using `%w`), so you get something like a stack trace of wrapped errors.

dessaya (approved):

tools/cluster/tests/cluster_stability_test.go
lmoe: No not right :) 
You can simplify it by leaving the second iota out. But the result is the same. 

https://yourbasic.org/golang/iota/

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+	// quorum is not met, work stops, incCounter should not equal numRequests
+	require.NotEqual(env.chainEnv.t, numRequestsBeforeFailure+numRequestsAfterFailure, int(counter))
+
+	// unfreeze nodes, after bootstrapping it is expected to reach a full quorum leading to an equal incCounter
+	env.unfreezeNodes()
+
+	waitUntil(t, env.chainEnv.counterEquals(int64(numRequestsBeforeFailure+numRequestsAfterFailure)), env.getActiveNodeList(), 60*time.Second, "incCounter matches expectation")
+}
+
+func runTestSuccessfulConsenseusWithReconnectingNodesAndHighQuorum(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequestsBeforeFailure, numRequestsAfterFailure int) {
```
jorgemmsilva:
```suggestion
func testConsenseusReconnectingNodesHighQuorum(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequestsBeforeFailure, numRequestsAfterFailure int) {
```

tools/cluster/tests/cluster_stability_test.go
lmoe: The idea is that the nodes break away *while* you send messages. That's why the node kills are in a separate thread.

tools/cluster/tests/cluster_stability_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+		runTestSuccessfulConsenseusWithReconnectingNodesAndHighQuorum(t, clusterSize, numValidators, numBrokenNodes, numRequestsBeforeFailure, numRequestsAfterFailure)
+	})
+}
+
+func runTestOneFailingNodeAfterTheOther(t *testing.T, clusterSize, numValidators, numBrokenNodes, numRequestsEachStep int) {
+	quorum := (2*numValidators)/3 + 1
+
+	t.Logf("Quorum: %v", quorum)
+	t.Logf("Maximal allowed broken nodes: %v", numValidators-quorum)
```
jorgemmsilva:
```suggestion
	t.Logf("Maximum allowed broken nodes: %v", numValidators-quorum)
```

jorgemmsilva (approved):
Good stuff 🚀

tools/cluster/cluster.go
lmoe: I really prefer to put the return in a variable before returning it, as it makes debugging easier when you need to step into the function.

Do you do it differently in the Wasp team? Then I just do it that way. :)

tools/cluster/tests/cluster_stability_test.go
lmoe: I had that this way first, but I think creating a new list for each test case is tedious, especially when you want to change the parameters again later. (Had to do this a few times already) That's why I've created these functions, which based on the node count, creates those lists. I think just passing a number is nicer than to recreate a list every time.

I can remove them and replace everything with lists, but really would like to keep it :D

tools/cluster/tests/cluster_stability_test.go
lmoe: Yeah I need the sabotageList also in other functions like UnfreezeNodes. It felt wrong to put the list directly into the sabotageFunction which also sets state to be used by other functions. I thought the separation makes sense there.

tools/cluster/cluster.go
lmoe: I think we already do, kind of :) 

The testIncCounterWithMildInstability/runTestFailsIncCounterIncreaseAsQuorumNotMet is not restarting nodes after they got killed, therefore they don't come up online again. Is this what you mean?

jorgemmsilva: I was thinking what should be the node's behaviour after its stopped abruptly, and started again.
either:
1. keep in broken state until manual maintenance
2. destroy broken db for the chain and re-sync state

not sure how we want to handle it, but if it's 2 - we could use these kinds of tests to make sure it is able to withstand it

tools/cluster/tests/cluster_stability_test.go
jorgemmsilva: I don't see how that's easier, but if you prefer it lets keep it. 😄 

btw, there's an utility function to generate sequential lists of int's: 
https://github.com/iotaledger/wasp/blob/1a8a1c8467f5b9da2ffe4b64708557feec4ba297/packages/util/misc.go#L36

tools/cluster/tests/env.go
lmoe: Agreed.

As this was introduced in another PR I would like to merge this here. I will address that in the following PRs :)

PR #319, opened 2021-08-03
better default value for faucet pow parameter

dessaya:
* `wasp-cli` now has faucetPoW = -1 by default, which makes sense for production usage
* `wasp-cluster` now has faucetPoW = 0 by default, which makes sense for testing usage
* Cluster tests will automatically set faucetPoW to -1 if goshimmer is provided


lmoe (approved):

tools/cluster/config.go
```diff
@@ -51,9 +51,9 @@ func DefaultConfig() *ClusterConfig {
 			TxStreamPort:    5000,
 			APIPort:         8080,
 			UseProvidedNode: false,
+			FaucetPoWTarget: 0,
```
jorgemmsilva, line 54: If I recall correctly, setting this to 0 resulted in cluster test being slowed down. 😞

jorgemmsilva: nevermind, I see you handle that case 👍

jorgemmsilva (approved):

PR #328, opened 2021-08-10
Improvement: handle excess outputs

jorgemmsilva:
There is a limit of outputs a tx can have, if we exceed this limit the state transition tx won't be accepted by L1.

There is no way to know how many outputs a request will produce without executing it. Because of this, we can only cap the number of outputs produced during batch execution.

This PR adds the following behaviours for excess number of outputs produced:
 - if a single request produces more outputs than can fit in a tx, it fails with an error
 - if a request produces outputs that overflow the current limit (ex: 1 output too many), we rollback that request and keep processing the rest of the current batch

This PR also introduces:
 - limit of inputs that can fit in a tx ensured by the VM (this is already ensured in consensus, but its good to have a robust self-contained logic for the VM)
- a nonce of for outputs per request included in the request metadata (so we don't produce identical outputs, which results in an error)

depends on: https://github.com/iotaledger/goshimmer/pull/1614


packages/vm/vmcontext/general.go
```diff
@@ -58,61 +58,24 @@ func (vmctx *VMContext) Entropy() hashing.HashValue {
...
 	return vmctx.req.ID()
 }
 
 const maxParamSize = 512
 
 // TODO implement send options
-//goland:noinspection GoUnusedParameter
 func (vmctx *VMContext) Send(target ledgerstate.Address, tokens colored.Balances, metadata *iscp.SendMetadata, options ...iscp.SendOptions) bool {
+	if vmctx.requestOutputCount >= MaxBlockOutputCount {
+		vmctx.log.Panicf("request with ID %s exceeded max number of allowed outputs (%d)", vmctx.req.ID().Base58(), MaxBlockOutputCount)
```
lunfardo314, line 70: not good. It means some requests can panic for no reason for the caller and fee is still charged.
I think correct approach would be somehow signal the main loop in the VM not to include this request at all. It can be done for example with special panic code which could be intercepted and treated accordingly

lunfardo314: oops, sorry, it seems it is exactly how it is implemented :)

packages/vm/vmcontext/runreq.go
```diff
@@ -79,6 +78,14 @@ func (vmctx *VMContext) RunTheRequest(req iscp.Request, requestIndex uint16) {
 		vmctx.mustCallFromRequest()
 	}()
 
+	if vmctx.blockOutputCount > MaxBlockOutputCount {
+		vmctx.shouldStopRunningBatch = true
```
lunfardo314: batch run should not be stopped, only no more on-ledger requests should be accepted (if inputs exceeded) and requests should be excluded if they want to create outputs but output count exceeded

packages/vm/vmcontext/runreq.go
```diff
@@ -301,6 +311,9 @@ func (vmctx *VMContext) mustUpdateOffledgerRequestMaxAssumedNonce() {
 }
 
 func (vmctx *VMContext) mustFinalizeRequestCall() {
+	if vmctx.shouldStopRunningBatch {
```
lunfardo314: ??

packages/vm/vmcontext/vmcontext.go
```diff
@@ -20,6 +20,11 @@ import (
 	"golang.org/x/xerrors"
 )
 
+var (
```
lunfardo314, line 23: why not `const`?

lunfardo314 (changes requested):

packages/vm/vmcontext/vmcontext.go
jorgemmsilva: it was const initially, but I decided to use var in order to better test it:
https://github.com/jorgemmsilva/wasp/blob/impr/produced-outputs/packages/vm/core/testcore/output_limits_test.go#L73

packages/vm/vmcontext/runreq.go
jorgemmsilva: makes sense, updated 👍

packages/vm/vmcontext/runreq.go
jorgemmsilva: removed

lunfardo314 (approved):

PR #329, opened 2021-08-10
Fix failing tests

ivange94:
`metrics. NewChainMetrics` was getting called on a `nil` receiver and this returned a `chainMetricsObj` that didn't have the metrics variables initialised and this caused the tests to fail when mempool tried to update metrics using the `chainMetricsObjt` returned. To fix this, I am checking whether the receiver is `nil` and if nil, I return a defaultChainmetrics object.

I also did some reorganisation to make the code more easy to follow.


packages/metrics/metrics.go
```diff
@@ -22,13 +22,29 @@ type Metrics struct {
...
-type chainMetrics struct {
-	metrics *Metrics
-	chainID *iscp.ChainID
-}
-
 func (m *Metrics) NewChainMetrics(chainID *iscp.ChainID) ChainMetrics {
-	return &chainMetrics{
+	if m == nil {
+		return DefaultChainMetrics()
+	}
```
jorgemmsilva, line 28: this cannot happen

ivange94: I don't understand

ivange94: That's actually the fix. A lot of the changes here is reorganisation. Was weird but it did happen. This method was getting called on a nil receiver and it didn't crash. It just returned the `chainMetricsObj` below which didn't have the metrics variables initialised.

packages/chains/chains.go
```diff
@@ -137,6 +137,7 @@ func (c *Chains) Activate(chr *registry.ChainRecord, registryProvider registry.P
 
 	defaultRegistry := registryProvider()
 	chainKVStore := c.getOrCreateKVStore(chr.ChainID)
+	chainMetrics := allMetrics.NewChainMetrics(chr.ChainID)
```
jorgemmsilva, line 140: this blows up when `allMetrics` is nil, right?

ivange94: So I did put log statements in there. `allMetrics == nil` returns true but code didn't crash. It actually successfully returned a `chainMetrics` object whose contents were nil though.

ivange94: > this blows up when allMetrics is nil, right?

This was my expectation but that's not what happened.

packages/metrics/metrics.go
jorgemmsilva: oh wow, you're right! I didn't know about this o.O

jorgemmsilva (approved):

PR #333, opened 2021-08-11
DOCS - Minor edit on accounts.md

lucas-tortora:
# Description of change

Minor text edit and formatting.  

https://github.com/iotaledger/smartcontract-development/issues/28

## Type of change

- Documentation Enhancement

It's just some minor text edits, but the site was built and run locally.

## Change checklist

- [ ] I have followed the contribution guidelines for this project

### DIFF
--- a/documentation/docs/contract_core/accounts.md
+++ b/documentation/docs/contract_core/accounts.md
@@ -1,42 +1,40 @@
-# The `accounts` contract
+# The `accounts` Contract
 
 The `accounts` contract is one of the [core contracts](overview.md) on each ISCP
 chain.
 
-The function of the `accounts` contract is to keep in its state a consistent ledger of
-on-chain accounts for the agents that control them. There are two types of agents who can do it:
-L1 addresses and smart contracts.
+The `accounts` contract keeps a consistent ledger of on-chain accounts in its state for the agents that control them. There are two types of agents who can do it: L1 addresses and smart contracts.
 
-The `accounts` contract provides functions to deposit and withdraw tokens.
-It also provides information about the assets deposited on the chain.  
+## Entry Points
+
+The `accounts` contract provides functions to deposit and withdraw tokens, and also provides information about the assets deposited on the chain.  
+
 Note that the ledger of accounts on the chain is consistently maintained behind scenes by the VM.
-The `accounts` contract provides a front-end of authorized access to those
-accounts for outside usersof the chain.
 
-### Entry Points
+### deposit
 
-* **deposit** moves tokens attached as a transfer to a target account on the
-  chain. If the agent ID parameter `a` is specified the target account is the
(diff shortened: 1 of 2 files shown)


lucas-tortora:
@iotaledger/tech-writers , could someone please review this before passing over to the WASP team?

charlesthompson3 (approved):
LGTM!

documentation/docs/contract_core/accounts.md
```diff
@@ -1,42 +1,40 @@
...
-The `accounts` contract provides functions to deposit and withdraw tokens.
-It also provides information about the assets deposited on the chain.  
 Note that the ledger of accounts on the chain is consistently maintained behind scenes by the VM.
-The `accounts` contract provides a front-end of authorized access to those
-accounts for outside usersof the chain.
 
-### Entry Points
+### deposit
+
+Moves tokens attached as a transfer to a target account on the chain. If the agent ID parameter `a` is specified, the target account is the account controlled by that agent ID. Otherwise, the target account is the account controlled by the caller. This makes sense only if it is a request, not if it is an on-chain call.
```
jorgemmsilva: I would word this sligtly differently, what do you think abou the following?:
```suggestion
Moves tokens attached as a transfer to a target account on the chain. By default the funds are deposited to the caller account, optionally a different target account can be specified with the agend ID parameter `a`.
This makes sense only if it is a request, not if it is an on-chain call.
```

> This makes sense only if it is a request, not if it is an on-chain call.

why is that? a SC can make an internal on-chain call to this endpoint. I would probably remove the last phrase
edit: I think this is based on a previous text by Evaldas, right?

jorgemmsilva (approved):

documentation/docs/contract_core/accounts.md
lucas-tortora: Hi @jorgemmsilva , it is indeed based on the previous text.  I definitely like your suggestion :) , I'll update it in a new commit as there's a small tipo in agent.

documentation/docs/contract_core/accounts.md
```diff
@@ -1,42 +1,40 @@
...
 chain.
 
-The function of the `accounts` contract is to keep in its state a consistent ledger of
-on-chain accounts for the agents that control them. There are two types of agents who can do it:
-L1 addresses and smart contracts.
+The `accounts` contract keeps a consistent ledger of on-chain accounts in its state for the agents that control them. There are two types of agents who can do it: L1 addresses and smart contracts.
+
+## Entry Points
+
+The `accounts` contract provides functions to deposit and withdraw tokens, and also provides information about the assets deposited on the chain.
```
aleksei-korolev, line 10: "The `accounts` contract" three times in a row. Sorry for reviewing too late.

PR #335, opened 2021-08-12
Feat: (mempool) handle fallback options

jorgemmsilva:


packages/chain/mempool/mempool_test.go
```diff
@@ -409,6 +409,49 @@ func TestTimeLock(t *testing.T) {
 	testStatsFun()
 }
 
+func TestFallbackOptions(t *testing.T) {
+	glb := coreutil.NewChainStateSync().SetSolidIndex(0)
+	rdr, _ := createStateReader(t, glb)
+	pool := New(rdr, iscp.NewInMemoryBlobCache(), testlogger.NewLogger(t), nil)
```
ivange94: You're calling `New` with `nil` MempoolMetrics. There is `mempool.DefaultChainMetrics()`

packages/chain/mempool/mempool.go
```diff
@@ -233,13 +233,22 @@ func (m *Mempool) traceOut(reqid iscp.RequestID) {
 	}
 }
 
+// don't process any request which deadline will expire within 10 minutes
+const FallbackDeadlineMinAlowedInterval = time.Minute * 10
```
ivange94: Possible typo  `Alowed` -> `Allowed`

packages/iscp/request.go
```diff
@@ -31,6 +31,10 @@ type Request interface {
 	Target() (Hname, Hname)
 	// returns time lock time or zero time if no time lock
 	TimeLock() time.Time
+	// retunrs the fallback address or nil if not set
```
lunfardo314: probably does not make much sense to have it the `Request` interface because fallback logic is only applicable to on-ledger requests.

lunfardo314 (approved):

PR #344, opened 2021-08-23
Add docs for interacting with smart contract from solo and wasp-cli

ivange94:
Ticket https://github.com/iotaledger/smartcontract-development/issues/29

### DIFF
--- a/documentation/docs/guide/core_concepts/smartcontract-interaction/example.md
+++ b/documentation/docs/guide/core_concepts/smartcontract-interaction/example.md
@@ -1,0 +1,69 @@
+# Example smart contract
+
+This example native smart contract stores a string in the state upon request and allows to retrive the store string using a view.
+
+The code below creates a new smart contract and initialize a processor with the required entry points and initializer method. The initializer method sets initial state.
+
+```go
+package helloworld
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


documentation/docs/guide/core_concepts/smartcontract-interaction/example.md
```diff
@@ -0,0 +1,69 @@
+# Example smart contract
+
+This example native smart contract stores a string in the state upon request and allows to retrive the store string using a view.
+
+The code below creates a new smart contract and initialize a processor with the required entry points and initializer method. The initializer method sets initial state.
+
+```go
```
jorgemmsilva, line 7: I don't think we should give a `built-in` or `native` contract as an example 😕 . Final users shouldn't be using this.

Imo we should give the users instructions for a normal rust/tiny-go smart contract.

lunfardo314: I agree with Jorge. This is good as internal PoC and example, we have some of this from older times. However, we should not put into the docs of how to develop contracts. It is misleading at best

sascha1337: Indeed been initially confusing, can confirm !

PR #347, opened 2021-08-24
Statemgr getCandidates optimisation

Juliusan:
State manager while picking which blocks to apply checks the state hash in each iteration. This should improve performance.

### DIFF
--- a/packages/chain/statemgr/syncmgr.go
+++ b/packages/chain/statemgr/syncmgr.go
@@ -95,7 +95,7 @@ func (sm *stateManager) doSyncActionIfNeeded() {
 		}
 		if approvedBlockCandidatesCount > 0 {
 			sm.log.Debugf("doSyncAction: trying to find candidates to commit from index %v to %v", startSyncFromIndex, i)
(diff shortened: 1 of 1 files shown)


kape1395:
I'm not sure, if it is OK to use recursion instead of a loop. Otherwise looks OK to me.

packages/chain/statemgr/syncmgr.go
```diff
@@ -143,7 +132,20 @@ func (sm *stateManager) getCandidatesToCommit(candidateAcc []*candidateBlock, fr
...
+				fromStateIndex, toStateIndex, candidatePrevStateHash.String(), calculatedPrevStateHash.String())
+			return nil, nil, false
+		}
+		calculatedState, err := stateCandidateBlock.getNextState(calculatedPrevState)
+		if err != nil {
+			sm.log.Errorf("getCandidatesToCommit from %v to %v: failed to apply synced block index #%d: %v",
+				fromStateIndex, toStateIndex, stateCandidateBlock.getBlock().BlockIndex(), err)
+			return nil, nil, false
+		}
+		resultBlocks, tentativeState, ok := sm.getCandidatesToCommit(append(candidateAcc, stateCandidateBlock), calculatedState, fromStateIndex+1, toStateIndex)
```
lunfardo314, line 148: I also prefer loop, not recursion

lunfardo314 (approved):

PR #352, opened 2021-08-25
Refactor: root -> governance

jorgemmsilva:
Moves most administrative functionality out of the `root` SC, into `governance`.
`root` is now only responsible for the SCs deployment logic.


lunfardo314 (approved):

packages/dashboard/templates/chaincontract.tmpl
```diff
@@ -13,18 +13,10 @@
 			<dt>Program hash</dt><dd><code>{{$c.ProgramHash.String}}</code></dd>
 			{{if $c.HasCreator}}<dt>Creator</dt><dd>{{ template "agentid" (args $chainid $c.Creator) }}</dd>{{end}}
 			<dt>Owner fee</dt><dd>
-				{{- if $c.OwnerFee -}}
-					<code>{{- $c.OwnerFee }} {{ $rootinfo.FeeColor -}}</code>
-				{{- else -}}
-					<code>{{- $rootinfo.DefaultOwnerFee }} {{ $rootinfo.FeeColor }}</code> (chain default)
-				{{- end -}}
+					<code>{{- .OwnerFee }} {{ $rootinfo.FeeColor -}}</code>
```
dessaya, line 16: Just checking: this behavior change is intentional? Previously if the contract did not have specific values for the fees, the dashbord here would show something like `owner fee: 300 IOTA (chain default)`.

Now `.OwnerFee` is guaranteed to have a value?

packages/vm/core/governance/contractfees.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"github.com/iotaledger/hive.go/marshalutil"
+)
+
+// ContractFeesRecord is a structure which contains the fee information for a contract
+type ContractFeesRecord struct {
+	// Chain owner part of the fee. If it is 0, it means chain-global default is in effect
+	OwnerFee uint64
+	// Validator part of the fee. If it is 0, it means chain-global default is in effect
+	ValidatorFee uint64 // validator part of the fee
```
dessaya, line 12: Nit: the comment is duplicated

packages/vm/core/governance/governanceimpl/fees.go
```diff
@@ -0,0 +1,65 @@
...
+	params := kvdecoder.New(ctx.Params(), ctx.Log())
+
+	hname := params.MustGetHname(governance.ParamHname)
+	rec := governance.FindContractFees(ctx.State(), hname)
+	if rec == nil {
+		rec = governance.NewContractFeesRecord(0, 0)
+	}
+
+	ownerFee := params.MustGetUint64(governance.ParamOwnerFee, 0)
+	ownerFeeSet := ownerFee > 0
```
dessaya, line 31: And what if you want to unset the fee? :thinking:

packages/vm/core/governance/governanceimpl/fees.go
```diff
@@ -0,0 +1,65 @@
...
+	params := kvdecoder.New(ctx.Params())
+	hname, err := params.GetHname(governance.ParamHname)
+	if err != nil {
+		return nil, err
+	}
+	feeColor, ownerFee, validatorFee := governance.GetFeeInfo(ctx, hname)
+	ret := dict.New()
+	ret.Set(governance.VarFeeColor, codec.EncodeColor(feeColor))
+	ret.Set(governance.VarOwnerFee, codec.EncodeUint64(ownerFee))
+	ret.Set(governance.VarValidatorFee, codec.EncodeUint64(validatorFee))
```
dessaya, line 63: If i understand correctly, this method returns the fees set in the contract, and if they are 0, the chain default fees are in effect, right?

Would it be more convenient if this method also returned the effective fees? To save the caller from doing `if owner fee == 0 then fetch the chain default`. Something like:

```
effectiveOwnerFee := ownerFee
if ownerFee == 0 {
   effectiveOwnerFee = fetchDefaultOwnerFee()
}
ret.Set(governance.VarEffectiveOwnerFee, codec.EncodeUint64(effectiveOwnerFee))
```

packages/dashboard/templates/chaincontract.tmpl
jorgemmsilva: fees are acquired via `FuncGetFeeInfo` from `governance`:
https://github.com/iotaledger/wasp/blob/develop/packages/dashboard/chaincontract.go#L59-L61

which will return the default fee in case it isn't specified for that contract:
https://github.com/iotaledger/wasp/blob/develop/packages/vm/core/governance/internal.go#L83-L98

packages/vm/core/governance/governanceimpl/fees.go
jorgemmsilva: it already does that 😄  ` governance.GetFeeInfo` does that logic internally

packages/vm/core/governance/governanceimpl/fees.go
jorgemmsilva: thanks for pointing this out. This is not correct. Fixing in a following PR and adding a test case for it.

PR #354, opened 2021-08-27
Add more metrics

ivange94:
Metrics for chain messages and vm run time.

### DIFF
--- a/packages/chain/consensus/action.go
+++ b/packages/chain/consensus/action.go
@@ -118,7 +118,6 @@ func (c *Consensus) runVMIfNeeded() {
 	}
 
 	c.log.Debugf("runVM needed: total number of requests = %d", len(reqs))
-	c.vmRunStartTime = time.Now()
 	// here reqs as a set is deterministic. Must be sorted to have fully deterministic list
 	c.sortBatch(reqs)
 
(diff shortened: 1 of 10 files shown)


packages/chain/consensus/action.go
```diff
@@ -230,6 +231,8 @@ func (c *Consensus) prepareVMTask(reqs []iscp.Request) *vm.VMTask {
 		c.chain.ReceiveMessage(&messages.VMResultMsg{
 			Task: task,
 		})
+		elapse := time.Since(c.vmRunStartTime)
+		c.consensusMetrics.RecordVMRunTime(elapse)
```
dessaya:
```suggestion
		elapsed := time.Since(c.vmRunStartTime)
		c.consensusMetrics.RecordVMRunTime(elapsed)
```

packages/metrics/chain.go
```diff
@@ -1,31 +1,44 @@
...
+	CountRequestAckMessages()
 	MempoolMetrics
+	ConsensusMetrics
 }
 
 type MempoolMetrics interface {
 	CountOffLedgerRequestIn()
 	CountOnLedgerRequestIn()
 	CountRequestOut()
+	RequestProcessingTime(iscp.RequestID, time.Duration)
```
dessaya: Suggest renaming to `RecordRequestProcessingTime` or similar, to make it consistent with `RecordVMRunTime`

dessaya (approved):

PR #366, opened 2021-09-01
Fix cluster test for windows

lmoe:
The stability tests freeze selected Wasp nodes using SIGSTOP/SIGCONT, this only works for unix based operating systems. 
As the syscall package is os dependent, we need to separate the calls for unix and windows.
In best case, we will have a solution for Windows at some point. Right now the function just throws an error on Windows systems.

### DIFF
--- a/packages/chain/chainimpl/chainimpl.go
+++ b/packages/chain/chainimpl/chainimpl.go
@@ -5,7 +5,6 @@ package chainimpl
 
 import (
 	"bytes"
-	logpkg "log"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


jorgemmsilva (approved):

dessaya (approved):

packages/chain/chainimpl/chainimpl.go
```diff
@@ -297,7 +296,7 @@ func (c *chainObj) processChainTransition(msg *chain.ChainTransitionEventData) {
 				// The error means a database error. The optimistic state read failure can't occur here
 				// because the state transition message is only sent only after state is committed and before consensus
 				// start new round
-				logpkg.Panicf("processChainTransition. unexpected error: %v", err)
+				c.log.Panicf("processChainTransition. unexpected error: %v", err)
```
lunfardo314, line 299: 👍

PR #367, opened 2021-09-01
Bugfixes

Juliusan:
Several small BugFixes:
* a sentence in docs to save someone a day searching the profiles
* error handling fix
* state manager fix to avoid "virtual state has been invalidated" error

### DIFF
--- a/packages/chain/chainimpl/chainimpl.go
+++ b/packages/chain/chainimpl/chainimpl.go
@@ -297,6 +297,7 @@ func (c *chainObj) processChainTransition(msg *chain.ChainTransitionEventData) {
 				// because the state transition message is only sent only after state is committed and before consensus
 				// start new round
 				c.log.Panicf("processChainTransition. unexpected error: %v", err)
+				return // to avoid "possible nil pointer dereference" in later use of `reqids`
 			}
 			// remove processed requests from the mempool
 			c.log.Debugf("processChainTransition state %d cleaning state %d: removing %d requests", stateIndex, i, len(reqids))
--- a/packages/chain/chainimpl/interface.go
+++ b/packages/chain/chainimpl/interface.go
@@ -94,8 +94,8 @@ func (c *chainObj) receiveMessage(msg interface{}, blocking bool) {
 		return
 	}
 	defer func() { // This is needed to handle possible write to a closed channel.
-		err := recover()
-		if err == "send on closed channel" {
+		err, ok := recover().(error)
+		if ok && err.Error() == "send on closed channel" {
 			c.log.Warnf("Failed to receive message, reason=%v", err)
 			return
 		}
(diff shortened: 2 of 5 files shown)


jorgemmsilva (approved):

packages/chain/chainimpl/chainimpl.go
```diff
@@ -297,6 +297,7 @@ func (c *chainObj) processChainTransition(msg *chain.ChainTransitionEventData) {
 				// because the state transition message is only sent only after state is committed and before consensus
 				// start new round
 				c.log.Panicf("processChainTransition. unexpected error: %v", err)
+				return // to avoid "possible nil pointer dereference" in later use of `reqids`
```
dessaya, line 300: `Panicf()` does not panic by itself? :thinking:

packages/peering/lpp/lppPeer.go
```diff
@@ -114,8 +114,8 @@ func (p *peer) SendMsg(msg *peering.PeerMessage) {
 	}
 	p.accessLock.RUnlock()
 	catch := func() {
-		err := recover()
-		if err == "send on closed channel" {
+		err, ok := recover().(error)
+		if ok && err.Error() == "send on closed channel" {
 			p.log.Warnf("Failed to send message, reason=%v", err)
 			return
 		}
```
dessaya, line 121: Same here.

packages/chain/chainimpl/interface.go
```diff
@@ -94,8 +94,8 @@ func (c *chainObj) receiveMessage(msg interface{}, blocking bool) {
 		return
 	}
 	defer func() { // This is needed to handle possible write to a closed channel.
-		err := recover()
-		if err == "send on closed channel" {
+		err, ok := recover().(error)
+		if ok && err.Error() == "send on closed channel" {
 			c.log.Warnf("Failed to receive message, reason=%v", err)
 			return
 		}
```
dessaya, line 101: I think this changes the behavior of the if block right below (github does not allow me to add a comment there), if recover() returns something that is not an error instance.

packages/chain/chainimpl/chainimpl.go
Juliusan: I think, it does. But somehow linter doesn't catch this. That's the linter warning I try to avoid.

packages/chain/chainimpl/interface.go
Juliusan: I did a lot of testing and `if recover() == "send on closed channel"` does not work for me. It simply doesn't catch "send on closed channel" error. This peace of code, however, works. So I am simply not aware of a better way, that works.

lunfardo314: it is indeed strange way of handling, based on message, however I cannot propose the better way.
I actually do not see writing to the closed channel anything natural, the code has to prevent any of this. It may be not very easy, yes

lunfardo314 (approved):

PR #371, opened 2021-09-01
Add prometheus and grafana dashboards

ivange94:


dessaya (approved):

tools/monitoring/config.json
```diff
@@ -0,0 +1,49 @@
+{
+  "database": {
+    "directory": "waspdb"
+  },
+  "logger": {
+    "level": "debug",
```
jorgemmsilva, line 6: ~maybe let's set `info` as the default? debug is too spammy for normal operations, I think~

edit: on a second thought I can see why we want users to run with debug logging for now

jorgemmsilva: btw, why not use the default `config.json` that is on the root of the project?

jorgemmsilva (approved):
Nice! 🚀

tools/monitoring/config.json
ivange94: The config in the root project doesn't have the metrics plugin turned on. I wasn't sure if it's something we'll always want to have turned on for the root config thats why I created this one.

jorgemmsilva: okay, makes sense 👍

PR #374, opened 2021-09-05
Abstracted colors and more

lunfardo314:
- EDIT. created next abstraction layer for `colored.Color` and `colored.Balances` with `+build` tags . 

- fixed some bugs in core contracts and even one in consensus

- added `SkipHeavy` function to testutils, to be able to skip heavy tests by adding `tags=skipheavy` into the `go test` command line


packages/iscp/colored/color.go
```diff
@@ -3,60 +3,90 @@ package colored
...
-	"github.com/iotaledger/hive.go/stringify"
 	"github.com/mr-tron/base58"
 	"golang.org/x/xerrors"
 )
 
-// region Color ////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Color is abstract color code used in ISCP.
+// It can be mapped into specific implementations of Goshimmer or Chrysalis by calling Init
+type Color []byte //nolint:gofumpt
+type ColorKey string
```
dessaya: Maybe add a comment explaining the purpose of `ColorKey`? Also, why not just have `type Color string`?

packages/iscp/colored/color.go
```diff
@@ -3,60 +3,90 @@ package colored
...
+}
+
+// Init initializes module with the specific length of the color code and specific code used as IOTA color code
+func Init(colLen int, iotaColor Color) {
+	if colorLength != 0 {
+		panic("color20.init called twice")
+	}
+	colorLength = colLen
+	IOTA = iotaColor
+}
```
dessaya: This makes it compulsory to call Init() from your main function or else it will not work. 

I think Jorge proposed the following, did you consider it?: have everything specific to each different implementation in a separate .go file with a build tag:

colored_iota20.go
```
// +build !l1_15
const ColorLength 32
```

colored_iota15.go
```
// +build l1_15
const ColorLength 666
```

And everything that's common here:

color.go
```
type Color[ColorLength]byte
```

I think it's a slightly better approach, doesn't require calling an Init function, and it's also less bug-prone since Color will still be a fixed size array instead of a slice.  What do you think?

packages/testutil/skip_heavy_true.go
```diff
@@ -0,0 +1,8 @@
+// only included if tag=skipheavy
+//+build skipheavy
+
+package testutil
+
+func init() {
+	skipHeavy = true
+}
```
dessaya: In this case we can also avoid having a global variable and init function with something like this:

skip_heavy_true.go
```
// +build skipheavy
func SkipHeavy(t *testing.T) {
	t.Logf("skipping heavy test %s", t.Name())
	t.SkipNow()
}
```

skip_heavy_false.go
```
// +build !skipheavy
func SkipHeavy(t *testing.T) {}
```

dessaya: BTW, another approach without the need for a new testutils function, is just moving all heavy tests to a separate .go file with `// +build !skipheavy`:

foo_test.go
```
func TestFoo(t *testing.T) {
  ...
}
```

foo_heavy_test.go
```
// +build !skipheavy

func TestFooHeavy(t *testing.T) {
  ...
}
```

dessaya (approved):

packages/iscp/colored/color.go
dessaya: Also, why the need for nolint here?

lunfardo314: No idea why linter is not happy :)

lunfardo314: Yes, string may be also an option. An equivalent one. Except that ColorKey wouldn’t be needed. Wanted closer to the byte array thing but maybe it make sense to rethink it.

packages/iscp/colored/color.go
lunfardo314: Considered it but I was sure linter or my IDE won’t allow it. Maybe I am wrong. Have to try

packages/testutil/skip_heavy_true.go
lunfardo314: I agree, but my concern about linter above still stands. Can’t confirm without trying. It is double declaration, may be

packages/iscp/colored/color.go
jorgemmsilva: you might have to provide the build flags to gopls, so that the IDE highlight works correctly

lunfardo314: Yeah, maybe that works. Will rewrite, if so

packages/iscp/colored/colored_15.go
```diff
@@ -0,0 +1,20 @@
...
+	MINT = Color{}
+)
+
+func init() {
+	for i := range MINT {
+		MINT[i] = 0xFF
+	}
+}
+
+// TODO TODO
```
jorgemmsilva, line 22: what is this for?

jorgemmsilva (approved):

packages/iscp/colored/colored_15.go
lunfardo314: it is just example of another file which declares variables but is never included into the build. 
The whole file is dummy not included into the build, just a template for the future (which will change anyway). 
Put a comment on that

PR #378, opened 2021-09-06
Use TLS instead of explicit message signing.

kape1395:


packages/peering/peering.go
```diff
@@ -265,22 +265,23 @@ func NewPeerMessageFromBytes(buf []byte, peerPubKey *ed25519.PublicKey) (*PeerMe
...
-		}
-		var signature ed25519.Signature
-		if signature, _, err = ed25519.SignatureFromBytes(signatureBin); err != nil {
-			return nil, xerrors.Errorf("failed to parse signature: %w", err)
-		}
-		if !peerPubKey.VerifySignature(buf[:len(buf)-lenBeforeSig], signature) {
-			return nil, xerrors.New("signature verification failed")
+		if peerPubKey != nil {
+			// Check the signature, if key is provided.
+			if peerPubKey == nil {
```
jorgemmsilva: this doesn't seem right. We're checking if `peerPubKey` is `nil` after checking it is not `nil`

kape1395: nice catch.

jorgemmsilva (approved):

PR #388, opened 2021-09-09
update docs on downloading goshimmer snapshot.bin

dessaya:


jorgemmsilva (approved):
we should probably just remove this and point directly to the goshimmer repo, tho :/

dessaya:
> we should probably just remove this and point directly to the goshimmer repo, tho :/

I agree, but this infirmation is buried in the middle of the README file in Goshimmer; I can't think of a way to link to that specific line in a friendly way.

PR #389, opened 2021-09-14
Fixed chain name not passed on

BugFreeSoftware:


contracts/common/deploy.go
```diff
@@ -26,14 +26,14 @@ func DeployWasmContractByName(chain *solo.Chain, scName string, params ...interf
 	return chain.DeployWasmContract(nil, scName, wasmFile, params...)
 }
 
-func StartChain(t *testing.T, scName string) *solo.Chain {
+func StartChain(t *testing.T, chainName string) *solo.Chain {
```
ivange94, line 29: For consistency I think this parameter should either go away or `StartChainAndDeployWasmContractByName` should also expect this parameter. Right now it's hardcoded to always start a chain with name `chain1`.

contracts/rust/erc20/test/init_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func TestDeployErc20(t *testing.T) {
-	chain := common.StartChain(t, ScName)
+	chain := common.StartChain(t, "chain1")
```
ivange94, line 13: Why not also make the chain name a global const like ScName?

ivange94 (approved):
LGTM,  just a minor suggestions.

jorgemmsilva (approved):

contracts/rust/erc20/test/init_test.go
BugFreeSoftware: It will change with the new way of using the schema tool anyway.

contracts/common/deploy.go
BugFreeSoftware: The parameter is necessary to be able to start a different chain. SC name is also necessary to be able to deploy a SC under a different name.

PR #391, opened 2021-09-14
Channels

Juliusan:
Replaced a simple Go channel by "infinite" channel in the most congested places: `chainimpl` and `lpp` implementation. This increased test performance and now all the integration tests pass for me consistently.

### DIFF
--- a/packages/chain/chainimpl/chainimpl.go
+++ b/packages/chain/chainimpl/chainimpl.go
@@ -57,7 +57,6 @@ type chainObj struct {
 	stateReader                      state.OptimisticStateReader
 	procset                          *processors.Cache
 	chMsg                            *channels.InfiniteChannel
-	chMsgOverflow                    atomic.Int32 // To log a level of a channel overflow.
 	stateMgr                         chain.StateManager
 	consensus                        chain.Consensus
 	log                              *logger.Logger
--- a/packages/chain/chainimpl/chainimpl.go
+++ b/packages/chain/chainimpl/chainimpl.go
@@ -35,6 +35,7 @@ import (
 	"github.com/iotaledger/wasp/packages/vm/processors"
 	"go.uber.org/atomic"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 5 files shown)


packages/chain/chainimpl/chainimpl.go
```diff
@@ -55,8 +56,7 @@ type chainObj struct {
 	chainStateSync                   coreutil.ChainStateSync
 	stateReader                      state.OptimisticStateReader
 	procset                          *processors.Cache
-	chMsg                            chan interface{}
-	chMsgOverflow                    atomic.Int32 // To log a level of a channel overflow.
+	chMsg                            *channels.InfiniteChannel
```
jorgemmsilva, line 59: > Note: Several types in this package provide so-called "infinite" buffers. Be very careful using these, as no buffer is truly infinite. If such a buffer grows too large your program will run out of memory and crash. Caveat emptor.

this seems like a slippery slope :grimacing:

Juliusan: Yes, it is true, I've read that. But instead I think about it as a trade-off. I've improved the performance and paid by increased memory usage. At least tests run now. Before, they were failing constantly at least for me.

Of course, we could improve on that concept. The simplest solution, I can think of for now, would be to limit the growth of the buffer simply dropping new messages or overwriting old ones by new ones if the maximum buffer capacity is reached. 

But the biggest advantage of this solution IMO is than channel is no longer used for synchronisation: it is used simply to pass information (which is its purpose). The sender is no longer required to stop and wait for receiver and the receiver is no longer pushed by sender to read the message ASAP. They can be really independent of each other except for information, they need to share.

kape1395:
I have asked Julius to do this change.
Please wait for my review.
We will need to add some kind of back-pressure with this approach, but for now this way looks much better than locking of channels.
At least we will be able to measure and monitor, and localize botlenecks.

kape1395 (approved):

PR #393, opened 2021-09-15
use go:generate to compile solidity contracts

dessaya:
This allows to recompile all solidity example contracts with a single command: `go generate` (provided that `solc` is installed).


jorgemmsilva (approved):
nice!

ivange94 (approved):

PR #396, opened 2021-09-20
Fix: DB bloat

jorgemmsilva:
Fixes the db bloat by collapsing all `stateUpdates` of each `block` into a single `stateUpdate` (thus saving space on keys that are re-used between updates).

also introduces the `dbinspector` tool, which helps with inspecting the contents of a wasp db

adds spam/stress tests

some documentation updates

small improvements: panics node when out of disk space, renders the dashboard even when committee info is not available


documentation/rfc/replay-off-ledger.md
```diff
@@ -58,7 +58,7 @@ In this case requests `3,4` will fail for no reason for the user.
 
 ## Proposed solution
 
-We propose to combine the two methods: by checking processed requests in the state and same time requiring and enforcing
+We propose to combine the two methods: by checking processed requests in the state and at same time requiring and enforcing
```
dessaya:
```suggestion
We propose to combine the two methods: by checking the processed requests in the state and at the same time requiring and enforcing an
```

packages/chain/statemgr/syncmgr.go
```diff
@@ -179,7 +180,10 @@ func (sm *stateManager) commitCandidates(candidates []*candidateBlock, tentative
 	sm.chain.GlobalStateSync().SetSolidIndex(tentativeState.BlockIndex())
 
 	if err != nil {
-		sm.log.Errorf("commitCandidates: failed to commit synced changes into DB. Restart syncing")
+		sm.log.Errorf("commitCandidates: failed to commit synced changes into DB. Restart syncing. %w", err)
+		if strings.Contains(err.Error(), "space left on device") {
+			sm.log.Panicf("Terminating WASP, no space left on disc.")
+		}
```
dessaya, line 186: Is there another case where it makes sense _not_ to panic?

tools/cluster/templates/waspconfig.go
```diff
@@ -16,13 +16,13 @@ type WaspConfigParams struct {
...
   "database": {
-    "inMemory": true,
+    "inMemory": false,
     "directory": "waspdb"
   },
   "logger": {
     "level": "info",
     "disableCaller": false,
-    "disableStacktrace": true,
+    "disableStacktrace": false,
```
dessaya: Is this intentional?

dessaya: Tip: you can still override any setting with something like:

```
$ DATABASE_INMEMORY=false LOGGER_DISABLESTACKTRACE=false wasp-cluster
```

dessaya (approved):

packages/chain/statemgr/syncmgr.go
jorgemmsilva: not sure, is it possible that there's some error writing to disk, but the next write works fine?

tools/cluster/templates/waspconfig.go
jorgemmsilva: reverting, thanks.

Juliusan (approved):

packages/chain/consensus/action.go
```diff
@@ -73,8 +73,8 @@ func (c *Consensus) proposeBatchIfNeeded() {
 		})
 	})
 
-	c.log.Infof("proposeBatch: proposed batch len = %d, ACS session ID: %d, state index: %d, proposal: %+v",
-		len(reqs), c.acsSessionID, c.stateOutput.GetStateIndex(), iscp.ShortRequestIDs(proposal.RequestIDs))
+	c.log.Infof("proposeBatch: proposed batch len = %d, ACS session ID: %d, state index: %d",
+		len(reqs), c.acsSessionID, c.stateOutput.GetStateIndex())
 	c.workflow.batchProposalSent = true
```
Juliusan, line 78: Sad to see it gone. It was very convenient to find the proposal visually in logs. But otherwise request ID logging was useless :-)

packages/chain/statemgr/syncmgr.go
```diff
@@ -179,7 +180,10 @@ func (sm *stateManager) commitCandidates(candidates []*candidateBlock, tentative
 	sm.chain.GlobalStateSync().SetSolidIndex(tentativeState.BlockIndex())
 
 	if err != nil {
-		sm.log.Errorf("commitCandidates: failed to commit synced changes into DB. Restart syncing")
+		sm.log.Errorf("commitCandidates: failed to commit synced changes into DB. Restart syncing. %w", err)
+		if strings.Contains(err.Error(), "space left on device") {
```
Juliusan, line 184: Is it possible to write this using `errors.Is(...)`. Everywhere I read, it says that error string comparison is bad practice. But I am not sure of working code, which is a good practice :-D

Juliusan (approved):

packages/chain/statemgr/syncmgr.go
jorgemmsilva: I wanted to do that actually, but didn't find an error to compare to :confused:

packages/chain/statemgr/syncmgr.go
lunfardo314: I think the only error from the state which can be tolerated is failure of the optimistic read. Otherwise: panic. We have a distributed system, if one node crashes with DB error, so be it

lunfardo314 (approved):

Issue #398, opened 2021-09-21
Websockets seem to halt the execution of smart contracts


lmoe:
Unfortunately I can't offer an extensive log, but just speak out of the experience I made throughout the testing.

The execution of the current fairroulette contract emits a few events which are sent via Nanomsg. Our current Websocket implementation listens to these events and publishes them through a Websocket port.

Somewhere in these stages, (this could also mean that the Nanomsg itself is the reason), Wasp behaves very uncoorperative. For example: Messages being dropped, whole execution halts at any random moment, the WS connection suddenly dies. Sometimes this is unrecoverable and requires the deletion of the Waspdb. Once I got this error message after restarting wasp `storeSyncingData not needed: stateOutput is nil`

I have experienced that, as soon as I drop off the Websocket connection and restart Wasp, the execution works reliable again.

dessaya:
Fixed in #399

PR #399, opened 2021-09-21
remove duplication in websocket code

dessaya:


packages/publisher/publisherws/publisherws.go
```diff
@@ -0,0 +1,87 @@
...
+			v, ok := p.clients.Load(chainID)
+
+			if !ok {
+				return
+			}
+			chainWsClients := v.(*sync.Map)
+
+			msg := msgType + " " + strings.Join(parts, " ")
+			chainWsClients.Range(func(key interface{}, clientCh interface{}) bool {
+				clientCh.(chan string) <- msg
```
dessaya: @lmoe regarding #398 
I don't know how to reproduce the issue, but my hunch is that this line might be the culprit. Can you test if any of these changes fixes it?

```go
go func() { clientCh.(chan string) <- msg }()
```

Or:

```go
select {
case clientCh.(chan string) <- msg:
default:
}
```

jorgemmsilva (approved):

lmoe (approved):
Yay!

PR #402, opened 2021-09-23
Wasmvm

BugFreeSoftware:
Fix to core contracts interface for SoloContext


BugFreeSoftware:
Note that a lot of the changed files were auto-generated by schema tool and don't really need review.
Especially all contract.go and main.go have the same transformation due to a change in the schema tool.

jorgemmsilva (approved):

PR #403, opened 2021-09-23
Roulette poc

lmoe:
This PR is one part of a two or three step PR to keep them a bit shorter in total. 
The next PR will contain more refactorings and documentation.

The roulette PoC is a simple roulette game. It contains a reference implementation written in TypeScript and Svelte to demonstrate the practical use of the Wasp node. 
In this game, OnLedger requests are used only. An offLedger method is available in the client library to play with though.


.gitignore
```diff
@@ -14,6 +14,6 @@ wasp-cli.json
 ./tools/wasp-cli/wasp-cli
 .docusaurus
 node_modules
-./wasp-cli
-./wasp
+wasp-cli
+wasp
```
dessaya: This will ignore any change under `tools/wasp-cli/` folder. Is that intended?

If you need to ignore an executable that is located somewhere else, maybe it's better to create a `.gitignore` file there with `./filename`. Or add the explicit path here, like the one in line 14.

contracts/rust/fairroulette/frontend/src/App.svelte
```diff
@@ -0,0 +1,542 @@
...
+    });
+  }
+
+  function isBroke(balance: bigint) {
+    return balance < 200;
+  }
+
+  function isWealthy(balance: bigint) {
+    return balance > 200;
+  }
```
jorgemmsilva: nit: 1 func should be enough for this, also what happens when balance is `200`? :smile:

contracts/rust/fairroulette/frontend/src/wasp_client/basic_wallet.ts
```diff
@@ -0,0 +1,192 @@
...
+    [address: string]: { [outputID: string]: IWalletOutput; };
+  } {
+    const outputsToConsume: { [address: string]: { [outputID: string]: IWalletOutput; }; } = {};
+
+    let iotasLeft = iotas;
+
+    for (const unspentOutput of unspentOutputs) {
+      let outputsFromAddressSpent = false;
+
+      for (const output of unspentOutput.outputs) {
```
jorgemmsilva, line 68: nit: splitting into functions would be nice to improve readability

contracts/rust/fairroulette/frontend/src/wasp_client/basic_wallet.ts
```diff
@@ -0,0 +1,192 @@
...
+        let requiredColorFoundInOutput = false;
+
+        if (!output.balances[Colors.IOTA_COLOR_STRING]) {
+          continue;
+        }
+
+        const balance = output.balances[Colors.IOTA_COLOR_STRING];
+
+        if (iotasLeft > 0n) {
```
jorgemmsilva, line 77: could exit loop once we have enough iotas, right?

jorgemmsilva (approved):
hard to do a thorough review, but overall it looks good to me :rocket: :slot_machine:

.gitignore
lmoe: At least on my instance, it will only ignore the executables wasp and wasp-cli. Changes on ./tools/wasp-cli are reflected still. 

/edit:

(I have reset it for now) :)

PR #404, opened 2021-09-23
Improvement: single State update

jorgemmsilva:
Changes to use a single stateUpdate instead of an array of stateUpdates, this helps when blocks are BIG so we can now have a single mutation (also automatically compresses many changes to the same key to a single mutation)

includes some minor improvements: use sequential nonce on waspclient, ads a "writeProfiles" option to the profiling plugin, and a minor tweak to the offledger broadcast logic.


packages/chain/chainimpl/chainimpl.go
```diff
@@ -346,7 +346,7 @@ func (c *chainObj) publishNewBlockEvents(blockIndex uint32) {
 
 	for _, msg := range evts {
 		c.log.Infof("publishNewBlockEvents: '%s'", msg)
-		publisher.Publish("vmmsg", c.chainID.Base58(), msg)
+		go publisher.Publish("vmmsg", c.chainID.Base58(), msg)
```
dessaya: This will not guarantee that events are triggered in order. Is this an issue?

packages/state/state.go
```diff
@@ -139,61 +138,48 @@ func (vs *virtualState) ApplyBlock(b Block) error {
...
-		upds[i] = b.(*blockImpl).stateUpdates[i]
-	}
-	vs.ApplyStateUpdates(upds...)
+	vs.ApplyStateUpdates(b.(*blockImpl).stateUpdate)
 	vs.empty = false
 	return nil
 }
 
-// ApplyStateUpdate applies one state update. Doesn't change state hash: it can be changed bu Apply block
+// ApplyStateUpdate applies one state update. Doesn't change state hash: it can be changed by Apply block
```
dessaya:
```suggestion
// ApplyStateUpdate applies one state update. Doesn't change the state hash: it can be changed by Apply block
```

packages/state/state.go
```diff
@@ -139,61 +138,48 @@ func (vs *virtualState) ApplyBlock(b Block) error {
...
-
-func (vs *virtualState) regenHash() {
-	compactUpdate := NewStateUpdate()
-	for _, su := range vs.updateLog {
-		for k, v := range su.Mutations().Sets {
-			compactUpdate.mutations.Set(k, v)
-		}
-		for k := range su.Mutations().Dels {
-			compactUpdate.mutations.Del(k)
+// StateCommitment return hash of the state. It is recursive hashing of the previous state hash and the block
```
dessaya:
```suggestion
// StateCommitment returns the hash of the state, calculated as a recursive hashing of the previous state hash and the block.
```

packages/state/state.go
```diff
@@ -139,61 +138,48 @@ func (vs *virtualState) ApplyBlock(b Block) error {
...
+func (vs *virtualState) StateCommitment() hashing.HashValue {
```
dessaya, line 173: Why the name change?

dessaya (approved):

packages/state/state.go
jorgemmsilva: I think this is in preparation to the next stage where state will be identified by something other than a simple hash (a verkle tree or similar)

packages/chain/chainimpl/chainimpl.go
jorgemmsilva: good point, fixing

PR #409, opened 2021-09-27
Fixing Typo in tutorial.md

Phyloiota:
fix typo


Phyloiota:
Currently, the Wiki is built from the Master branch. I saw that the Typo is already corrected in 'develop', but should be fixed here so the wiki is correct

lmoe (approved):

PR #417, opened 2021-10-01
Add initial fairroulette documentation

lmoe:
The inital fairroulette documentation. It's being worked at, but can already land on develop in this stage.


fijter (approved):
lgtm

PR #441, opened 2021-10-09
add cname to gh_action

huhn511:
Please remove this file after merge:

https://github.com/iotaledger/wasp/blob/develop/documentation/static/CNAME

with this PR the gh_action will set it


jorgemmsilva (approved):
@huhn511 can we remove the static CNAME file in this PR, so both changes happen a the same time?

PR #453, opened 2021-10-12
Channels

Juliusan:
Communication via channels improved once more. Infinite channel dependency changed to custom implementation based on that dependency. For clarity reasons (Evaldas said that "channel" is a wrong name for that) infinite channel is renamed to pipe. The message queue of the pipe can be limited (in the 2 use cases in wasp it is limited to 10000 messages). The messages can be prioritised. Tests showed, that pipe in wasp usually get congested, when not priority messages are passed, which is a desirable property. More tests might be needed too.

The only thing that is missing, is the ability to drop messages, which are already queued. Current implementation is based on `==` equality, which simply does not work for messages, which are usually pointers. Places for future changes are marked with `TODO`. A method must be found to get a "hash" (not necessarily a number - hence the quotes) such that if `msg1` is different than `msg2`, then `hash(msg1) != hash(msg2)` and `hash(_)` always returns something, that can be put as a key of a map. Using pointer of the message as a hash satisfy this condition. However it would be beneficial to have that if `msg1` is the same as `msg2`, then `hash(msg1) == hash(msg2)` at least to some extent. Using pointer fails in this regard as the same message created on different points of program execution usually have different addresses.


dessaya:
I'm out of the loop here, so I'm not aware of all the requirements of the problem to be solved here. I'd like to understand the motivation for this change, could you explain? Because looks like a lot of code and I am surprised that it is not something that can be solved using regular Go channels, or something like https://pkg.go.dev/container/heap .
Thanks!

packages/chain/chainimpl/chainimpl.go
```diff
@@ -31,13 +31,15 @@ import (
...
 	"github.com/iotaledger/wasp/packages/util"
+	"github.com/iotaledger/wasp/packages/util/pipe"
 	"github.com/iotaledger/wasp/packages/vm/core/blocklog"
 	"github.com/iotaledger/wasp/packages/vm/processors"
 	"go.uber.org/atomic"
 	"golang.org/x/xerrors"
-	"gopkg.in/eapache/channels.v1"
 )
 
+const maxMsgBuffer = 10000
```
lunfardo314, line 41: does it mean the pipe still has a limited buffer size? Or I am misunderstanding, then comment on the constant would be helpful

packages/chain/chainimpl/chainimpl.go
```diff
@@ -103,10 +105,29 @@ func NewChain(
...
+				messages.MsgSignedResultAck:
+				return true
+			default:
+				return false
+			}
+		default:
+			return true
+		}
+	}
+	messageHashFun := func(msg interface{}) interface{} { // TODO
```
lunfardo314: I am not sure how we want to hash it. If that would be `[]byte` , an ordinary `blake2b` hash would be ok. With `interface{}` there's no proper way of doing that or other way of detecting duplicates.

lunfardo314: so, `messageHashFun := func(msg interface{}) interface{}` is not needed, just use `hashing.HashData()`

lunfardo314 (approved):

packages/chain/chainimpl/chainimpl.go
Juliusan: Yes, for now buffer size is limited. But the limit can be easily removed (there is a constructor without a buffer limit). This functionality is just addressing a concern that without a limit pipe can use up all the memory and then Wasp would crash on out of memory error.

packages/chain/chainimpl/chainimpl.go
Juliusan: Everything is `[]byte` in computers, so can be hashed :-D But we need to take only the information that uniquely identify message (and not go object). Sadly, this information cannot contain `[]byte`.

Juliusan:
@dessaya The main requirement is that the sender should never be forced to wait for the receiver to read the older message. Earlier this was solved using go routines, which were started just to wait for the channel to be ready to accept, which caused a lot of confusion and was not scaling well. The infinite queue solution (`gopkg.in/eapache/channels.v1 v1.1.0`) supported much larger throughput.

The secondary goals were:
1. Ability to prioritise some (not a lot) messages
2. Ability to limit the queue size in order not to obtain out of memory error
3. Ability to reject messages, which are already queued for sending
4. Implement all of this, so it can easily be configured (as well as switched of) to the actual needs

I decided to take `gopkg.in/eapache/channels.v1 v1.1.0` implementation and improve on that. Currently everything is solved, except 3 which holds only partially due to naive approach to hashing, as I mentioned.

I'll try to experiment with heap, but it will save only code for 1.

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	go.uber.org/zap v1.16.0
 	golang.org/x/crypto v0.0.0-20210506145944-38f3c27a63bf
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
+	gopkg.in/yaml.v2 v2.4.0
```
kape1395, line 39: Is that intended?

kape1395 (approved):

PR #468, opened 2021-10-14
EVM: add new `evmlight` flavor

dessaya:
This PR is intended to be merged after the beta release. It is a large change, maybe we should do a meeting where I walk through the code.

----

Adds an alternative EVM implementation called `evmlight`.

With this change, one can choose between two different 'flavors' when deploying an EVM chain: `evmchain` (the old one) and `evmlight` (the new one).

The `evmlight` implementation is more space efficient, and also includes a proof of concept support for integration with ISCP, in the form of an `iscp` EVM contract and a forked/modified EVM interpreter.

I tried to duplicate code as little as possible.


contracts/native/evm/evmchain/impl.go
```diff
@@ -0,0 +1,231 @@
...
+
+func initialize(ctx iscp.Sandbox) (dict.Dict, error) {
+	a := assert.NewAssert(ctx.Log())
+	genesisAlloc, err := evmtypes.DecodeGenesisAlloc(ctx.Params().MustGet(evm.FieldGenesisAlloc))
+	a.RequireNoError(err)
+	chainID, err := codec.DecodeUint16(ctx.Params().MustGet(evm.FieldChainID), evm.DefaultChainID)
+	a.RequireNoError(err)
+	emulator.InitGenesis(
+		int(chainID),
+		rawdb.NewDatabase(emulator.NewKVAdapter(ctx.State())), // TODO: use subrealm to avoid collisions with evm management
```
jorgemmsilva, line 50: TODO left here.

btw, I think we should probably deprecate the "EVM management" stuff, and just move `gasPerIota` to the main `governance` contract (maybe also rename it to `evmGasPerIota`, to not be confused with the wasm gas, or find a way to make both "gases" ~equivalent)

dessaya: Thanks for reviewing!

I already addressed the TODO but I plan to push this change in my next PR. I think it's safe to merge anyway since it passes all tests and it's the same code that is in production anyway :P

Re: EVM management: Agree, but again, maybe let's do that in a later PR?

jorgemmsilva: yes of course, just writing what came to mind!

contracts/native/evm/evmchain/internal.go
```diff
@@ -0,0 +1,160 @@
...
+	_ = emu.Close()
+}
+
+func withEmulatorR(ctx iscp.SandboxView, f func(*emulator.EVMEmulator) dict.Dict) (dict.Dict, error) {
+	emu := emulator.NewEVMEmulator(
+		rawdb.NewDatabase(emulator.NewKVAdapter(buffered.NewBufferedKVStoreAccess(ctx.State()))),
+		timestamp(ctx),
+	)
+	defer emu.Close()
+	return f(emu), nil
```
jorgemmsilva: isn't `error` always `nil`? :thinking:

dessaya: Yes, but it returns `error` to make it compatible with the return value of the contract handler functions, so that it is possible to do

```go
func someHanlder(ctx iscp.SandboxView) (dict.Dict, error) {
  return withEmulatorR(...)
}
```

But maybe it is silly, I'm open to changing it.

jorgemmsilva: I was thinking we could change the func signature to the following:

```go
func withEmulatorR(ctx iscp.SandboxView, f func(*emulator.EVMEmulator) (dict.Dict, error)) (dict.Dict, error) {
```

and then just `return f(emu)`

does that make sense?
(edit: fix code block)

dessaya: Yes, I considered that initially but then all uses of this would have `return something, nil`. Again, I'm not very opinionated, so if it makes the code clearer I'll change it.

IMO the root issue here is probably that the `error` return value in handlers is useless. Is it being used at all? Does any of our handlers return an error?

jorgemmsilva: >IMO the root issue here is probably that the error return value in handlers is useless. 

you're right. I found a couple of places where we do return an error, but most of the time we just panic.

I think there's a fundamental difference between "panic'ing" and returning an error ( panic returns the fees iirc), but maybe this doesn't make sense, especially when have support for languages that don't have multiple returns.

@lunfardo314 what do you think about this?

packages/evm/evmtypes/receipt.go
```diff
@@ -0,0 +1,79 @@
...
+	return receipt, err
+}
+
+// EncodeReceiptFull encodes the receipt including fields not serialized by EncodeReceipt
+func EncodeReceiptFull(r *types.Receipt) []byte {
+	m := marshalutil.New()
+	writeBytes(m, EncodeReceipt(r))
+	m.WriteBytes(r.TxHash.Bytes())
+	m.WriteBytes(r.BlockHash.Bytes())
+	writeBytes(m, r.BlockNumber.Bytes())
```
jorgemmsilva, line 38:
```suggestion
	m.WriteBytes(r.BlockNumber.Bytes())
```
ocd

jorgemmsilva (approved):

packages/evm/evmtypes/receipt.go
dessaya: `writeBytes()` is a helper function that actually writes two things: size + bytes; then `readBytes` does the reading. This is for cases when the size is variable, such as `*big.Int`.

jorgemmsilva: Okay, makes sense. But isn't the blocksize bytesize constant?

dessaya: You mean the blocknumber? I looked into the implementation of big.Int and it seems to be dynamic. It makes sense because big.Int represents an arbitrarily large number.

PR #471, opened 2021-10-15
Update Dockerfile and documentation

lmoe:


Dockerfile
```diff
@@ -22,4 +22,12 @@ RUN go build -o . -tags=${BUILD_TAGS} -ldflags="${BUILD_LD_FLAGS}" ./...
 # Wasp build
 FROM gcr.io/distroless/cc
 
+EXPOSE 7000/tcp
+EXPOSE 9090/tcp
+EXPOSE 5550/tcp
+EXPOSE 4000/udp
+
 COPY --from=build /wasp/wasp /usr/bin/wasp
+COPY docker_config.json /etc/wasp_config.json
```
jorgemmsilva, line 31: I think the removal of this file might have been intentional, let's wait for @luke-thorne 
anyway, it makes sense to just link the proper config when spinning up an image

lmoe: I had a talk with him about it already.
We should ship a default config since the config differs from the normal one.

jorgemmsilva (approved):

PR #474, opened 2021-10-15
Roulette poc

lmoe:
This is the second (and last) part of the Roulette PoC containing an UX/UI overhaul, cleaned up code, updated smart contracts, bugfixes, wasp api additions, dockerization, documentation and a lot of heart.

For reference, see: #403


documentation/docs/guide/example_projects/fair_roulette.md
```diff
@@ -3,287 +3,312 @@ keywords:
...
-The fair roulette is an example reference implementation demonstrating the development, setup and interaction with a smart contract.
+Fair roulette is an example reference implementation which demonstrates the development, setup, and interaction with a smart contract.
 
 ## Introduction
 
-For this example, we have created a very simple betting game in which players can bet on a number of a certain range. 
+The Fair roulette example project is a simple betting game in which players can bet on a number within a certain range.
 
-The goal of the game is to bet on the right number, to win a share of the placed funds. This is being done in rounds. 
+The game consists of many rounds in which the player will try to bet on the right number to win a share of the placed funds.
```
luke-thorne: Would using `bet` make more sense in this context?

```suggestion
The game consists of many rounds in which the player will try to bet on the right number to win a share of the bet funds.
```

documentation/docs/guide/example_projects/fair_roulette.md
```diff
@@ -3,287 +3,312 @@ keywords:
...
+You can build the frontend by running the following commands:
+
+```bash
+cd contracts/rust/fairroulette/frontend
+npm install
+npm run build_worker
 ```
 
-It's possible to head directly to the deposit and deployment, otherwise request funds from the faucet first. `./wasp-cli request-funds`. 
+After this, you can run `npm run dev` which will run a development server that exposes the transpiled frontend on [`http://localhost:5000`](http://localhost:5000).
```
luke-thorne, line 295: There have been a couple question regarding how to change the host/port configurations. Should we add a small snippet showing how to set them with `HOST=0.0.0.0 PORT=8080 npm run dev`?

luke-thorne (commented):
I only had 2 minor suggestions. Feel free to ignore them if they don't make sense.

contracts/rust/fairroulette/fairroulette.go
```diff
@@ -32,10 +34,20 @@ const NanoTimeDivider = 1000_000_000
...
+// - 'number', which must be an Int64 number from 1 to MAX_NUMBER
 // The 'member' function will save the number together with the address of the better and
 // the amount of incoming iotas as the bet amount in its state.
 func funcPlaceBet(ctx wasmlib.ScFuncContext, f *PlaceBetContext) {
+	bets := f.State.Bets()
+
+	for i := int32(0); i < bets.Length(); i++ {
+		bet := bets.GetBet(i).Value()
+
+		if bet.Better.Address().String() == ctx.Caller().Address().String() {
```
jorgemmsilva:
```suggestion
		if bet.Better.Address() == ctx.Caller().Address() {
```
:point_up: this should work I think

contracts/rust/fairroulette/fairroulette.go
```diff
@@ -10,16 +10,18 @@
 package fairroulette
```
jorgemmsilva, line 10: why do we have a `fairroulette.go` file inside `contracts/rust/...` ? feels odd :thinking:

contracts/rust/fairroulette/frontend/src/lib/app.ts
```diff
@@ -0,0 +1,332 @@
...
+
+    if (!timestamp || !roundStartedAt) return undefined;
+
+    if (roundStartedAt == 0) {
+        return 0;
+    }
+
+    const diff = Math.round(timestamp - roundStartedAt);
+
+    // TODO: Explain.
```
jorgemmsilva: TODO here

packages/webapi/webapiutil/callview.go
```diff
@@ -1,20 +1,29 @@
...
 	"github.com/iotaledger/wasp/packages/kv/optimism"
 	"github.com/iotaledger/wasp/packages/vm/viewcontext"
 )
 
+// TODO: Find a proper solution, this is not a desired fix
+var mu sync.Mutex
+
 func CallView(ch chain.ChainCore, contractHname, viewHname iscp.Hname, params dict.Dict) (dict.Dict, error) {
+	mu.Lock()
+	defer mu.Unlock()
```
jorgemmsilva, line 18: @BugFreeSoftware did you figure out a fix for the view issue?

jorgemmsilva (approved):
this is way too big to give a meaningful review, but overall it LGTM :shipit:

contracts/rust/fairroulette/fairroulette.go
lmoe: I had the same discussion with Eric about it :) 

I fully agree that we should split it at some point in some way or another, but this needs to be done inside the Schema tool as it generates Rust and Go inside the same root folder.

At least we should rename the general rust folder into something like wasm to indicate that these contracts are for wasmtime or so. But I don't know if that breaks stuff when I just rename it.

contracts/rust/fairroulette/frontend/src/lib/app.ts
lmoe: Nice catch, can be removed alltogether.

PR #476, opened 2021-10-15
Wasmvm

BugFreeSoftware:
Refactoring in preparation of concurrency bug solution


packages/vm/wasmproc/wasmcontext.go
```diff
@@ -0,0 +1,146 @@
...
+		// init function was missing, do nothing
+		return nil, nil
+	}
+
+	if wc.function == FuncDefault {
+		// TODO default function, do nothing for now
+		return nil, nil
+	}
+
+	if wc.function == ViewCopyAllState {
```
lunfardo314: attackable for any big state, for example `blocklog`? Better to protect somehow, for example by setting maximum returned keys or something

lunfardo314 (approved):

packages/vm/wasmproc/wasmcontext.go
BugFreeSoftware: I will remove that. We don't need it at the moment.

PR #482, opened 2021-10-18
feat: allow open deployment to anyone

jorgemmsilva:


packages/vm/core/root/interface.go
```diff
@@ -31,9 +32,11 @@ const (
...
-	FuncDeployContract         = coreutil.Func("deployContract")
-	FuncGrantDeployPermission  = coreutil.Func("grantDeployPermission")
-	FuncRevokeDeployPermission = coreutil.Func("revokeDeployPermission")
-	FuncFindContract           = coreutil.ViewFunc("findContract")
-	FuncGetContractRecords     = coreutil.ViewFunc("getContractRecords")
+	FuncDeployContract          = coreutil.Func("deployContract")
+	FuncGrantDeployPermission   = coreutil.Func("grantDeployPermission")
+	FuncRevokeDeployPermission  = coreutil.Func("revokeDeployPermission")
+	FuncOpenDeploymentToAnyone  = coreutil.Func("openDeploymentToAnyone")
+	FuncCloseDeploymentToAnyone = coreutil.Func("closeDeploymentToAnyone")
```
dessaya: Grant/Revoke vs Open/Close... may be confusing?

Also, another option is to keep a single set of methods `grantDeployPermission` / `revokeDeployPermission`, and allow calling them with some predefined constant like `grantDeployPermission(ALL)`, like SQL does: https://www.techonthenet.com/mysql/grant_revoke.php

luke-thorne: Would we be able to use this in the future to grant specific addresses the ability the deploy smart contracts instead of limiting it to only the owner or allowing anyone?

jorgemmsilva: @luke-thorne that is already possible with the current implementation.

They way I see it, we have 2 ways to deploy contracts:
1) open - anyone can deploy
2) closed - only approved users can deploy

its possible to switch between the two using `open/closeDeploymentToAnyone`.

In the closed setting, its possible to manage which users can deploy using `grant`/`revoke`.

While I like the idea of using something akin to SQL `"grantDeployPermission(ALL)"`, its a bit weird because the grant/revoke entrypoints expect an agentID, so either we add an extra parameter, or chose an agentID to 'mean" ALL (all 0's maybe).

dessaya: Yeah, I agree that it would also be weird to add a parameter or to use a "special" AgentID to mean ALL.

I think that what needs improvement is naming: `openDeploymentToAnyone` / `closeDeploymentToAnyone` seem a bit off. Maybe `requireDeployPermission(true/false)?`

jorgemmsilva: sounds good :+1:  will refactor

dessaya (approved):

PR #488, opened 2021-10-19
Check for func/view hname collision in schema definition file

BugFreeSoftware:


tools/schema/generator/schema.go
```diff
@@ -133,11 +134,19 @@ func (s *Schema) compileFuncs(schemaDef *SchemaDef, params, results *FieldMap, v
 	}
 	for _, funcName := range sortedFuncDescs(templateFuncs) {
 		if views && schemaDef.Funcs[funcName] != nil {
-			return fmt.Errorf("duplicate func/view name")
+			return fmt.Errorf("duplicate func/view name: %s", funcName)
```
jorgemmsilva, line 137: just wondering: would it be possible to point to the line(s) where the errors happens in the source code? that would be nice

jorgemmsilva (approved):

tools/schema/generator/schema.go
BugFreeSoftware: The YAML/JSON parsers do not preserve line numbers.
But in most cases it is pretty clear where the error is.

PR #489, opened 2021-10-19
Bugfixes

Juliusan:


jorgemmsilva:
@Juliusan should this PR target `develop` instead of `master`?

jorgemmsilva:
I think its just a duplicated from https://github.com/iotaledger/wasp/pull/490 right? We should close this.

PR #498, opened 2021-10-19
Add forced round restart function in case of hangers

lmoe:


contracts/rust/fairroulette/lib.go
```diff
@@ -24,6 +25,24 @@ func OnLoad() {
...
 }
 
+type ForceResetContext struct {
+	State MutableFairRouletteState
+}
+
+func funcForceResetThunk(ctx wasmlib.ScFuncContext) {
+	ctx.Log("fairroulette.funcForceReset")
+	// only SC creator can restart the round forcefully
+	ctx.Require(ctx.Caller() == ctx.ContractCreator(), "no permission")
```
luke-thorne, line 35:
```suggestion
	ctx.Require(ctx.Caller() == ctx.ContractCreator(), "permission denied")
```

If the caller is not the creator, this would result in a permission denied error, right? "no permission" isn't a commonly used term.

contracts/rust/fairroulette/src/lib.rs
```diff
@@ -48,6 +49,24 @@ fn on_load() {
...
 }
 
+pub struct ForceResetContext {
+    state: MutableFairRouletteState,
+}
+
+fn func_force_reset_thunk(ctx: &ScFuncContext) {
+    ctx.log("fairroulette.funcForceReset");
+    // only SC creator can restart the round forcefully
+    ctx.require(ctx.caller() == ctx.contract_creator(), "no permission");
```
luke-thorne, line 59:
```suggestion
    ctx.require(ctx.caller() == ctx.contract_creator(), "permission denied");
```

same as above

luke-thorne (commented):
Just some ideas about different verbiage.

contracts/rust/fairroulette/lib.go
lmoe: That's part of the schema tool, that's something we need to change there. Otherwise the file gets overwritten with next use. I would merge it if it's just this part and mark @BugFreeSoftware to discuss this further. What do you think? :)

luke-thorne (approved):

PR #502, opened 2021-10-19
Feat: roulette demo frontend updates

begonaalvarezd:
This PR adds the possibility to integrate the demo with Google Analytics and adds a opt-in/opt-out cookie disclaimer. 
### Local development: 
In order to integrate the demo with GA you have to add your key `googleAnalyticsId: 'XXXXXXXX'` to `contracts/rust/fairroulette/frontend/config.dev.js`. 
If you leave it undefined or dont use it, the GA integration will not take place.
### Docker:
When you build a docker image of the frontend, if you want to include google analytics you have to build the image using: 
```bash 
docker build -t "image_name:image_tag" --build-arg GOOGLE_ANALITICS_ID="XXXXXXXX" .
```
If you dont add the the `--build-arg`, the GA integration will not take place.


contracts/rust/fairroulette/frontend/src/lib/utils.ts
```diff
@@ -7,4 +7,41 @@ export const generateRandomId = (): string => {
...
+
+  gtag('config', gaID)
+
+  const script = document.createElement('script')
+  script.src = `https://www.googletagmanager.com/gtag/js?id=${gaID}`
+  document.body.appendChild(script)
+}
+
+export const setCookie = (cookieName: string, cookieValue: string | boolean, expirationDays: number): void => {
+  var d = new Date()
```
lmoe: Lets at least get rid off the vars and use let/const please.

lmoe (approved):

contracts/rust/fairroulette/frontend/src/lib/utils.ts
jorgemmsilva: this should be an eslint check :stuck_out_tongue:  :point_up:

lmoe: Yeah it should :)

begonaalvarezd: Code updated 🙏🏼

lmoe (approved):

PR #512, opened 2021-10-20
Feat: roulette poc text updates

begonaalvarezd:


luke-thorne:
@begonaalvarezd can you change one more thing? The instructions wording after placing a bet might be better as "The next round will begin in a few seconds" instead of "The game is starting in a couple of seconds"
image

luke-thorne (approved):

PR #541, opened 2021-10-25
Ignore temporar files in git

Juliusan:
Ignore files, that are created in running `make` targets, but are not supposed to be comitted to Git.


jorgemmsilva (approved):

kape1395 (changes requested):
I'm not sure, if the ignore is not to general at least because of
```
./packages/wasp
```

kape1395 (approved):

PR #548, opened 2021-10-26
Master to dev merge

lmoe:


jorgemmsilva (approved):
the changes look fine, but we should never do " merge `master` to `develop`"

new commits should  go to develop first, if we need something in master we cherry-pick those changes and replay them on the master branch. This way the flow of changes is unidirectional `develop` => `master`

lmoe:
Yeah that was basically the point. By mistake I merged a PR which was targeted to master instead of develop. Also as you can see, some other PRs were only targeted for master as well.

I wanted to get both branches aligned again with this PR, otherwise they diverge too much from each other.

PR #550, opened 2021-10-27
Add metrics docs

ivange94:
This is still a draft because metrics is not yet enabled on the testnet.


documentation/docs/guide/chains_and_nodes/testnet.md
```diff
@@ -47,6 +47,8 @@ The testnet can be accessed via a series of endpoints that have been made availa
       - Information about GoShimmer via the API
 - https://demo.sc.iota.org
   - Our FairRoulette demo application to see a live smart contract in action
+- https://metrics.sc.iota.org
```
luke-thorne:
```suggestion
- https://metrics.wasp.sc.iota.org
```

In the other file you reference this being under the wasp subdomain. Let's use that here as well.

luke-thorne (commented):
You had 2 different metrics URLs. Let's just use 1.

luke-thorne (approved):

PR #553, opened 2021-10-27
Minor fixes

Juliusan:
Some minor changes: logging and more space for ACS data transmission, as 5k off ledger queries was too much for it to handle.

### DIFF
--- a/packages/vm/viewcontext/viewcontext.go
+++ b/packages/vm/viewcontext/viewcontext.go
@@ -58,8 +58,8 @@ func (v *Viewcontext) CallView(contractHname, epCode iscp.Hname, params dict.Dic
 			default:
 				err = xerrors.Errorf("viewcontext: panic in VM: %v", err1)
 			}
-			v.log.Errorf("CallView: %v", err)
-			v.log.Errorf(string(debug.Stack()))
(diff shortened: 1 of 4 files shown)


packages/vm/viewcontext/viewcontext.go
```diff
@@ -58,8 +58,8 @@ func (v *Viewcontext) CallView(contractHname, epCode iscp.Hname, params dict.Dic
 			default:
 				err = xerrors.Errorf("viewcontext: panic in VM: %v", err1)
 			}
-			v.log.Debugf("CallView: %v", err)
-			v.log.Debugf(string(debug.Stack()))
+			v.log.Errorf("CallView: %v", err)
+			v.log.Errorf(string(debug.Stack()))
```
dessaya: This was set to Debug because it might spam the logs, e.g. if a user calls a view function with an invalid parameter, it is not really a bug and we don't want that to be in the log; especially the stack trace.

Juliusan: Fixed that.

dessaya (approved):

jorgemmsilva (approved):

PR #572, opened 2021-11-03
Fix: Flush DB to disk more often

jorgemmsilva:
In order to prevent what happened last week with the broken access node that was forcefully stopped and couldn't recuperate, this PR makes some changes so that :
 - Chain registry changes are always persisted on disk
 - Chain blocks are written flushed to disk every time a new "state transition" happens

### DIFF
--- a/packages/database/registrykvstore/registrykvstore.go
+++ b/packages/database/registrykvstore/registrykvstore.go
@@ -1,0 +1,78 @@
+package registrykvstore
+
+import "github.com/iotaledger/hive.go/kvstore"
+
+// registrykvstore is just a wrapper to any kv store that flushes the changes to disk immediately (Sets or Dels)
+// this is to prevent that the registry database is corrupted if the node is not shutdown gracefully
+
+type RegistryKVStore struct {
+	store kvstore.KVStore
+}
+
+func New(store kvstore.KVStore) kvstore.KVStore {
+	return &RegistryKVStore{store}
+}
+
+func (s *RegistryKVStore) WithRealm(realm kvstore.Realm) kvstore.KVStore {
+	return s.store.WithRealm(realm)
+}
+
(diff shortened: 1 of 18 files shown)


packages/vm/processors/cache.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // Cache stores all initialized VMProcessor instances used by a single chain
 type Cache struct {
-	*sync.Mutex
+	Mutex      *sync.Mutex
```
dessaya: Can it be private?

packages/database/registrykvstore/registrykvstore.go
```diff
@@ -0,0 +1,78 @@
...
+func (s *RegistryKVStore) Get(key kvstore.Key) (value kvstore.Value, err error) {
+	return s.store.Get(key)
+}
+
+func (s *RegistryKVStore) Set(key kvstore.Key, value kvstore.Value) error {
+	err := s.store.Set(key, value)
+	if err != nil {
+		return err
+	}
+	return s.store.Flush()
```
dessaya, line 49: Could this result in a negative performance impact? I don't know how hive / rocksdb work but in general the DB should be "intelligent", and have its own write-ahead log, and know what's the best time to flush.

dessaya (approved):

packages/database/registrykvstore/registrykvstore.go
jorgemmsilva: writes to the chain-registry are pretty sparse. It should impact performance when that happens, but I don't think its meaningful, and it keeps the node "safer" from forced shutdowns

packages/vm/processors/cache.go
jorgemmsilva: good point, updated

PR #573, opened 2021-11-03
Disable spuriously failing test

BugFreeSoftware:


contracts/wasm/testcore/test/concurrency_test.go
```diff
@@ -32,6 +32,10 @@ func TestCounter(t *testing.T) {
 
 func TestSynchronous(t *testing.T) {
 	run2(t, func(t *testing.T, w bool) {
+		// TODO fails with 999 instead of 1000 at WaitForPendingRequests
+		if *wasmsolo.GoDebug {
```
jorgemmsilva, line 36: Will this be skipped in GH actions? Why the wasmsolo flag?

jorgemmsilva (approved):

contracts/wasm/testcore/test/concurrency_test.go
BugFreeSoftware: Yes, the run2() function runs several versions of the code. 
I'm skipping the GoVM version this way.

jorgemmsilva: Makes sense, thanks!

PR #576, opened 2021-11-05
BugFix for "ApplyBlock: b state index #x can't be applied to the empty state"

Juliusan:
A fix for a bug, which didn't allow to apply any blocks to the state, which is loaded from the DB and is not an origin state. A test, checking this exact situation, is also included. All the regular tests pass. Heavy tests not run.

### DIFF
--- a/packages/state/state.go
+++ b/packages/state/state.go
@@ -27,7 +27,6 @@ import (
 type virtualStateAccess struct {
 	chainID            *iscp.ChainID
 	db                 kvstore.KVStore
-	empty              bool
 	kvs                *buffered.BufferedKVStoreAccess
 	committedHash      hashing.HashValue
 	appliedBlockHashes []hashing.HashValue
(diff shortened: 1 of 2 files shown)


jorgemmsilva (approved):
LGTM, but maybe it can be simplified.

Can't we have a single function, instead of applyBock and applyAnyBlock?

It's just checking if the index and timestamp match not enough? Why do we need the "empty" flag?

Juliusan:
`empty` flag is needed for applying origin block to origin state. If `vs` is an origin state, it doesn't have a block index (and, possibly a timestamp), so `vs.BlockIndex()+1 != b.BlockIndex()` (more precisely `vs.BlockIndex()`) would throw an error.

Another approach would be to somehow catch this situation in `ApplyBlock` method. However, I think, my way is better: in this case, there is an exception for origin block only and no need to check this indirectly.

jorgemmsilva:
Any reason for the "origin state" to not come with the "origin block" already? :smile:

Juliusan:
Haven't dived that deep :-D

PR #587, opened 2021-11-08
skeleton for the sandbox gas

lunfardo314:
- sandbox gas skeleton added
- refactored Utils() implementation


packages/vm/core/blob/impl.go
```diff
@@ -67,8 +58,14 @@ func storeBlob(ctx iscp.Sandbox) (dict.Dict, error) {
 		blbSizes.MustSetAt([]byte(k), EncodeSize(size))
 		sizes[i] = size
 		totalSize += size
+		totalSizeWithKeys += size + uint32(len(k))
 	}
 
+	// TODO this code is experimental. Alternative approach would be burning gas directly in the state access interface
+	//  Set and Del mutation would burn net difference between size of the current key+value and the new one
+	//  Burning storage gas can be negative, which means we are saving the space and therefore adding to the available budget
```
jorgemmsilva, line 66: what happens if the execution of a requests results in negative gas, do we take funds out of the "chain account" to pay the user?

what happens in this scenario if the chain account has been "harvested" recently?

packages/vm/sandbox/sandboxview.go
```diff
@@ -69,5 +68,21 @@ func (s sandboxView) State() kv.KVStoreReader {
...
+func (s sandboxView) Burn(gas int64) {
+	s.vmctx.GasBurn(gas)
+}
+
+func (s sandboxView) Budget() int64 {
+	return s.vmctx.GasBudget()
+}
+
+func (s sandboxView) SetBudget(gasBudget int64) {
+	s.vmctx.GasSetBudget(gasBudget)
```
jorgemmsilva, line 87: views shouldn't have Gas, I think

PR #593, opened 2021-11-10
feat: show request details in dashboard & wasp-cli

dessaya:


packages/iscp/request/request.go
```diff
@@ -289,6 +289,7 @@ func (req *OnLedger) Output() ledgerstate.Output {
 
 // Params returns solid args if decoded already or nil otherwise
 func (req *OnLedger) Params() (dict.Dict, bool) {
+	// FIXME: this returns nil after deserializing a processed request (see tools/wasp-cli/chain/blocklog.go)
```
jorgemmsilva, line 292: `req.Params()` is the same as `req.Args()`, but are only available after `request.SolififyArgs` is called.
I'm not sure why this is needed, it has something to do with "downloader" :shrug: . Maybe we could remove the entire "solidification" stuff?

@lunfardo314 what do you think about this?

jorgemmsilva (approved):

packages/iscp/request/request.go
dessaya: +1 to remove that if it's not needed anymore

PR #608, opened 2021-11-17
impl: (VM) unify all UTXO types into a single onledger_request

jorgemmsilva:


packages/iscp/requestdata/request_metadata.go
```diff
@@ -0,0 +1,121 @@
+package requestdata
+
+import (
+	"github.com/iotaledger/hive.go/marshalutil"
+	iotago "github.com/iotaledger/iota.go/v3"
+	"github.com/iotaledger/wasp/packages/iscp"
+	"github.com/iotaledger/wasp/packages/kv/dict"
+)
+
+type RequestMetadata struct {
```
dessaya, line 10: Nit: "metadata" usually means "data about the data". I think that a better name for this object should be something like "RequestData" or "RequestParams" or just "Request".

packages/iscp/requestdata/requestdata.go
```diff
@@ -66,57 +63,48 @@ type RequestData interface {
...
 }
 
 type Features interface {
-	TimeLock() (TimeLockOptions, bool)
-	Expiry() (ExpiryOptions, bool)
-	ReturnAmount() (ReturnAmountOptions, bool)
+	TimeLock() *TimeData
+	Expiry() *TimeData
+	ReturnAmount() (uint64, bool)
 	SwapOption() (SwapOptions, bool) // for the new swap
```
dessaya, line 95: should be called `SwapOptions`

packages/iscp/requestdata/requestdata.go
```diff
@@ -48,7 +45,7 @@ func (t TypeCode) String() string {
 type UTXOMetaData struct {
 	UTXOInput          iotago.UTXOInput
 	MilestoneIndex     uint32
-	MilestoneTimestamp time.Time
+	MilestoneTimestamp uint64
```
dessaya, line 48: Why not use `time.Time`?
Also, is this a unix timestamp? In seconds, milliseconds, nanoseconds? IMO every timestamp that is not defined as `time.Time` should have a comment specifying the type, or a name like `MilestoneTimestampUnixMillis`

packages/iscp/requestdata/requestdata.go
```diff
@@ -66,57 +63,48 @@ type RequestData interface {
 }
 
 type TimeData struct {
-	ConfirmingMilestoneIndex uint32
-	ConfirmationTime         time.Time // should better be UnixNano ?
+	MilestoneIndex uint32
+	Timestamp      uint64
```
dessaya, line 67: Same here

dessaya (approved):

packages/iscp/requestdata/request_offledger.go
```diff
@@ -4,18 +4,17 @@ import (
...
+	iotago "github.com/iotaledger/iota.go/v3"
 	"github.com/iotaledger/wasp/packages/hashing"
 	"github.com/iotaledger/wasp/packages/iscp"
 	"github.com/iotaledger/wasp/packages/iscp/requestdata/placeholders"
 	"github.com/iotaledger/wasp/packages/kv/dict"
 )
 
 type OffLedger struct {
-	chainID        *iscp.ChainID
+	chainID        iscp.ChainID
```
dessaya, line 17: Why this change? IIRC we are using `*iscp.ChainID` everywhere.

packages/iscp/requestdata/requestdata.go
jorgemmsilva: is a unix timestamp in seconds I think.
I initially had it as time.Time, but changed to uint64 after feedback from @lunfardo314 , also this way it matches the same iotago type

packages/iscp/requestdata/request_offledger.go
jorgemmsilva: because this was changed :shrug: 
https://github.com/iotaledger/wasp/blob/stardust-vm/packages/iscp/chainid.go#L57

packages/iscp/requestdata/requestdata.go
jorgemmsilva: Agree, this will likely be removed anyway soon

packages/iscp/requestdata/request_metadata.go
jorgemmsilva: requestData is the name of the package and the new interface that we will use in the VM. Request is already used, and there's also UTXOMetadata.

I think RequestMetadata fits because its the request data that is present in the "metadata feature block".

I agree the naming is getting weird, tho

PR #610, opened 2021-11-18
Peer message sending review

Juliusan:
This is a very big PR. My aim was to simplify and unify interfaces and their usage. I concentrated on peer message sending between the wasp nodes. The goals I had in mind were:

- `PeerMessage` structure had fields, which were used during only the part of its lifetime. It wasn't always clear, which fields are set and which are not in which places of the code. There were some duplications with `RecvEvent` structure too. I replaced these two by four other structures: `PeerMessageData` - the actual data, which is passed between the nodes, `PeerMessageNet` - the structure, which has (or knows how to make) a byte array out of `PeerMessageData` structure. This is only needed in `peering` package where such serialisation/deserialisation happens. `PeerMessageIn` - which is used only in the receiver side and in addition to `PeerMessageData` contains information of the peer from which it was returned. And `PeerMessageGroupIn` - also used only in the receiver side but only if the receiver is part of peer group - in addition to `PeerMessageIn` it contains the index of peer in the group.
- More clear way of matching sender and receiver code. Now each different receiver (state manager, consensus, dkg, etc) has its own unique type and then that receiver ensures that the message type for each message it receives is unique. Now while sending message, the sender provides a receiver code (in addition to peering ID) and the receiver attaches to only the messages, that are sent to it. Although this doesn't ensure that each message is received by only one attached function, however IMO it eases the understanding of which code is responsible for handling which received messages. Moreover, the peering ID is now part of group/domain which is used to send messages. This way it is more clear, which peering ID is used in which situations. This IMO eases understanding of internode communication
- Removing a bottleneck of chain. Earlier all the messages that were sent to consensus, state manager, acs and chain itself were first queued in one pipe in their chain. Usually the messages are sent between the same type of parties of different nodes (e.g. state manager of node 1 to state manager of node 2). Now every receiver attaches and listens to the messages it is supposed to receive and chain doesn't queue all the messages simply to forward them to the correct place and queue them in another queue.
- Using pipes more often. In my understanding, the message passing should really be asynchronous (unless, of course, it is used to synchronise something). Pipes ensure that. Sender is not forced to wait for receiver even if the receiver is very congested.

On the way I also did some smaller things:
- tried to improve logging
- tried to separate not related structures into different files for clarity
- tried to unify terminology
- tried to hide (make private) objects that should not be public
- and I also workarounded a bug, which was happening in `develop` as well; you'll notice it by a long comment in lines 624-635 of `packages/chain/consensus/action.go` (next to "NOTE:").

I am not completely satisfied with the result. I think, more could (and should) be done in that direction. But this is a good place to stop and discuss. All the tests are passing, even the heavy ones, except `TestSpamOnledger`. `TestSpamOffledger` passes with logging switched to info level.


packages/chain/consensus/commonsubset/commonsubsetcoordinator.go
```diff
@@ -47,28 +53,30 @@ type CommonSubsetCoordinator struct {
...
 	dkShare *tcrypto.DKShare,
 	log *logger.Logger,
 ) *CommonSubsetCoordinator {
-	return &CommonSubsetCoordinator{
+	ret := &CommonSubsetCoordinator{
 		csInsts:   make(map[uint64]*CommonSubset),
 		csAsked:   make(map[uint64]bool),
 		lock:      sync.RWMutex{},
-		peeringID: peeringID,
+		committee: committee,
```
kape1395: Now the ACS depends on a commiittee and a group. Only a group was a dependency before.

packages/chain/consensus/commonsubset/setup_test.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2020 IOTA Stiftung
```
kape1395: This just to avoid passing peeringID?

packages/chain/consensus/consensus.go
```diff
@@ -52,14 +55,13 @@ type Consensus struct {
...
 	log                              *logger.Logger
-	eventStateTransitionMsgCh        chan *messages.StateTransitionMsg
-	eventSignedResultMsgCh           chan *messages.SignedResultMsg
-	eventSignedResultAckMsgCh        chan *messages.SignedResultAckMsg
-	eventInclusionStateMsgCh         chan *messages.InclusionStateMsg
-	eventACSMsgCh                    chan *messages.AsynchronousCommonSubsetMsg
-	eventVMResultMsgCh               chan *messages.VMResultMsg
-	eventTimerMsgCh                  chan messages.TimerTick
-	closeCh                          chan struct{}
+	eventStateTransitionMsgPipe      pipe.Pipe
```
kape1395, line 58: I'm not sure if internal communication should be done via pipes instead of channels.
Or maybe this part is not for internal communication?

packages/chain/consensus/consensus.go
```diff
@@ -79,113 +81,190 @@ type workflowFlags struct {
...
 	log := chainCore.Log().Named("c")
-	ret := &Consensus{
+	ret := &consensus{
 		chain:                            chainCore,
 		committee:                        committee,
 		mempool:                          mempool,
 		nodeConn:                         nodeConn,
 		vmRunner:                         runvm.NewVMRunner(),
-		resultSignatures:                 make([]*messages.SignedResultMsg, committee.Size()),
+		resultSignatures:                 make([]*messages.SignedResultMsgIn, committee.Size()),
```
kape1395, line 108: Is that OK that for each peering message we have another with SenderID added?

packages/chain/consensus/consensus.go
```diff
@@ -79,113 +81,190 @@ type workflowFlags struct {
...
 	}
+	committee.AttachToPeerMessages(peerMessageReceiverConsensus, ret.receiveCommitteePeerMessages)
 	ret.refreshConsensusInfo()
 	go ret.recvLoop()
 	return ret
 }
 
-func (c *Consensus) IsReady() bool {
+func (c *consensus) receiveCommitteePeerMessages(peerMsg *peering.PeerMessageGroupIn) {
+	if peerMsg.MsgReceiver != peerMessageReceiverConsensus {
```
kape1395: Maybe peerMessageReceiverConsensus can be passed to the attach function and avoid checks in all the handlers?

packages/testutil/peeringNetBehaviour.go
```diff
@@ -137,7 +137,7 @@ func (n *peeringNetUnreliable) sendDelayed(recv *peeringMsg, outCh chan *peering
 	}
 	n.log.Debugf(
 		"Network delivers message %v -%v-> %v (duplicate %v/%v, delay=%vms)",
-		recv.from.netID, recv.msg.MsgType, dstNetID, dupNum, dupCount, delay.Milliseconds(),
+		recv.from, recv.msg.MsgType, dstNetID, dupNum, dupCount, delay.Milliseconds(),
```
kape1395, line 140: before the from field had a peer. You was able to take its NetID, PubKey, send a message to it.
Not it has only a NetID? In most of the communication the netID should be avoided in future, taking PubKeys as the primary identifier.

packages/testutil/peeringNetworkProvider.go
```diff
@@ -108,14 +108,15 @@ type peeringNode struct {
 }
 
 type peeringMsg struct {
-	from *peeringNode
-	msg  peering.PeerMessage
+	from string
```
kape1395: I'm not sure, if its a good idea to replace it with a string. Comment above.

packages/chain/messages/peer_offledger_request_msg.go
```diff
@@ -15,21 +15,12 @@ type OffLedgerRequestMsg struct {
...
 }
 
-func NewOffLedgerRequestMsg(chainID *iscp.ChainID, req *request.OffLedger) *OffLedgerRequestMsg {
-	return &OffLedgerRequestMsg{
-		ChainID: chainID,
-		Req:     req,
-	}
+type OffLedgerRequestMsgIn struct {
+	OffLedgerRequestMsg
+	SenderNetID string
```
kape1395, line 20: Pub key will be needed here, at least in the near future.

packages/dkg/node.go
```diff
@@ -27,19 +27,19 @@ type NodeProvider func() *Node
...
+	identity     *ed25519.KeyPair                 // Keys of the current node.
+	secKey       kyber.Scalar                     // Derived from the identity.
+	pubKey       kyber.Point                      // Derived from the identity.
+	blsSuite     Suite                            // Cryptography to use for the Pairing based operations.
+	edSuite      rabin_dkg.Suite                  // Cryptography to use for the Ed25519 based operations.
+	netProvider  peering.NetworkProvider          // Network to communicate through.
+	registry     registry.DKShareRegistryProvider // Where to store the generated keys.
+	processes    map[string]*proc                 // Only for introspection.
+	procLock     *sync.RWMutex                    // To guard access to the process pool.
+	initMsgQueue chan *initiatorInitMsgIn         // Incoming events processed async.
```
kape1395, line 39: Why have you renamed it so?
That is for incoming messages, not just for init messages. Or maybe I miss something?

kape1395: Or maybe recvStopCh should be renamed in the same way?

packages/peering/lpp/lppPeer.go
```diff
@@ -37,19 +36,22 @@ type peer struct {
...
+var _ peering.PeerSender = &peer{}
+
 func newPeer(remoteNetID string, remotePubKey *ed25519.PublicKey, remoteLppID libp2ppeer.ID, n *netImpl) *peer {
 	log := n.log.Named("peer:" + remoteNetID)
 	messagePriorityFun := func(msg interface{}) bool {
-		peerMsg, ok := msg.(*peering.PeerMessage)
+		peerMsg, ok := msg.(*peering.PeerMessageNet)
 		if ok {
-			switch peerMsg.MsgType {
+			return peerMsg.MsgType > 0
```
kape1395: Why is this needed?
The comment bellow should be removed? Should contain no upper level modules info.

packages/peering/lpp/lppPeer.go
```diff
@@ -135,31 +137,29 @@ func (p *peer) SendMsg(msg *peering.PeerMessage) {
 	p.sendPipe.In() <- msg
 }
 
-func (p *peer) RecvMsg(msg *peering.PeerMessage) {
+func (p *peer) RecvMsg(msg *peering.PeerMessageNet) {
+	// p.log.Debugf("Peer message received from peer %v, peeringID %v, receiver %v, type %v, length %v, first bytes %v",
```
kape1395: Should be removed?

packages/peering/lpp/lppPeer.go
```diff
@@ -179,8 +179,17 @@ func (p *peer) sendMsgDirect(msg *peering.PeerMessage) {
 	p.accessLock.Lock()
 	p.lastMsgSent = time.Now()
 	p.accessLock.Unlock()
+	// p.log.Debugf("Peer message sent to peer %v, peeringID %v, receiver %v, type %v, length %v, first bytes %v",
```
kape1395: cleanup?

packages/peering/peer_message.go
```diff
@@ -0,0 +1,130 @@
...
+	"bytes"
+
+	"github.com/iotaledger/wasp/packages/hashing"
+	"github.com/iotaledger/wasp/packages/util"
+	"github.com/iotaledger/wasp/packages/util/pipe"
+)
+
+// RecvEvent stands for a received message along with
+// the reference to its sender peer.
+/*type RecvEvent struct {
```
kape1395: cleanup?

packages/chain/consensus/commonsubset/commonsubsetcoordinator.go
Juliusan: I aggree. This should be some message sending/attaching interface instead of committee.

packages/chain/consensus/consensus.go
Juliusan: I see this as a way to make code more type safe. You won't be able to retrieve sender ID earlier than it is assigned.

packages/chain/consensus/consensus.go
Juliusan: It is. This check can be removed.

packages/chain/consensus/consensus.go
Juliusan: The reason for using pipes is the same - to make message passing just message passing and not a sender-receiver synchronisation mechanism.

packages/chain/consensus/commonsubset/commonsubsetcoordinator.go
Juliusan: Removed ACS dependency on committee in the last commit. Only the dependency on committee group is remaining.

packages/chain/consensus/consensus.go
Juliusan: Check has been removed in later commit.

packages/peering/lpp/lppPeer.go
Juliusan: This line produces many log entries, which usually are not very helpful. But sometimes it can be beneficial to have such log. This is left to make the testing later easier.

packages/peering/lpp/lppPeer.go
Juliusan: Same as before.

packages/peering/peer_message.go
Juliusan: Removed in later commit.

packages/dkg/node.go
Juliusan: This channel is used only for init message. Other messages have their own channels. I believe, we can do without `recvStopCh` in this case. Implemented these changes in later commit.

packages/peering/lpp/lppPeer.go
Juliusan: I was planning to think of a way to prioritise messages, but didn't have enough time for that. Removed the old code and added TODO for now.

packages/chain/chainimpl/chainimpl.go
```diff
@@ -147,152 +134,77 @@ func NewChain(
 		offledgerBroadcastInterval:       offledgerBroadcastInterval,
 		pullMissingRequestsFromCommittee: pullMissingRequestsFromCommittee,
 		chainMetrics:                     chainMetrics,
+		dismissChainMsgPipe:              pipe.NewLimitInfinitePipe(maxMsgBuffer),
```
jorgemmsilva: this probably only needs `1` instead of `maxMsgBuffer` right?

packages/chain/chain.go
```diff
@@ -26,12 +23,21 @@ import (
...
 	GetCommitteeInfo() *CommitteeInfo
-	ReceiveMessage(interface{})
+	StateCandidateToStateManager(state.VirtualStateAccess, ledgerstate.OutputID)
 	Events() ChainEvents
 	Processors() *processors.Cache
 	GlobalStateSync() coreutil.ChainStateSync
 	GetStateReader() state.OptimisticStateReader
 	Log() *logger.Logger
+
+	// Most of these methods are made publick for mocking in tests
```
kape1395: publick -> public

kape1395 (approved):

jorgemmsilva (approved):

packages/dkg/messages.go
```diff
@@ -55,8 +56,13 @@ const (
 	// in response to duplicated messages from other peers. They should be treated
 	// in a special way to avoid infinite message loops.
 	rabinEcho byte = peering.FirstUserMsgCode + 44
+
+	peerMessageReceiverDkg     = byte(4)
+	peerMessageReceiverDkgInit = byte(5)
```
dessaya: I saw similar constants elsewhere. Are these IDs that need to be unique? Then they should probably be all declared in the same file, to make it clearer that they are related, and reduce the chance of creating a collision.

packages/peering/lpp/lppPeer.go
dessaya: Commented out code is still confusing to the reader. Another option for this case is having a constant `traceMessages = false` or similar, and here:

```
if traceMessages {
  p.log.Debugf(...)
}
```

dessaya (approved):

packages/dkg/messages.go
Juliusan: I was trying to export as little as possible. Most of these variables are private - hence there are no inter-receiver messages there. Only one is public to allow message passing between the consensus and chainimpl. I thought about keeping them unique using some grep command. e.g. `grep "[pP]eerMessageReceiver.*=" -r packages/ | cat`

PR #614, opened 2021-11-19
Acs test by size

kape1395:
ACS tests improved by considering various payload sizes.


Makefile
```diff
@@ -21,6 +21,9 @@ test: install
 test-short:
 	go test -tags $(BUILD_TAGS) --short --count 1 -failfast ./...
 
+test-heavy:
```
jorgemmsilva: we have `test-full` already that does the same thing.
I prefer the name `test-heavy` tho

jorgemmsilva (approved):

Makefile
kape1395: Sorry, that was old uncommitted change. Haven't noticed duplication after merging develop. Removed my version.

PR #615, opened 2021-11-19
Add missing semicolon after get_object_id()

nanderstabel:
Missing semicolon after get_object_id() function causes compile error.

Example:
```
error: expected `;`, found `ImmutableBettingRound`
  --> src/state.rs:29:87
   |
29 |         let sub_id = get_object_id(self.obj_id, Key32(index), TYPE_ARRAY | TYPE_BYTES)
   |                                                                                       ^ help: add `;` here
30 |         ImmutableBettingRound { obj_id: sub_id }
   |         --------------------- unexpected token
```

### DIFF
--- a/tools/schema/generator/generator_rust.go
+++ b/tools/schema/generator/generator_rust.go
@@ -328,7 +328,7 @@ func (g *RustGenerator) generateProxyArrayNewType(field *Field, proxyType string
 			varType = g.generateArrayType(varType)
 		}
 		g.printf("\n    pub fn get_%s(&self, index: i32) -> %s {\n", snake(field.Type), proxyType)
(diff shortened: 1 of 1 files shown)


BugFreeSoftware:
Thank you for addressing this, but it was fixed quite a while ago and already merged into develop.
In addition the schema tool has undergone a serious refactoring (also in develop already).
Always check develop first for the latest changes, and we never do pull-requests into master.
We love your contribution. So don't let this discourage you.

nanderstabel:
I just read on the wiki that pull request should be done on develop and saw the issue was indeed fixed using the nice templates. Thank you for the encouragement, I will checkout develop first from now on.

PR #616, opened 2021-11-21
Change webapi callview endpoint from GET to POST

mrbianchi:
Currently webapi callview endpoint is listening as a GET method but it reads the parameters from the body.
GET method with body doesn't follow the standard and it needs to be changed to POST.

_Yes, GET with body exists and is usable but I'm having problems with that, JS fetch() or Node.js http.request() doesnt allow you to request using GET with body because it isnt standard_

### DIFF
--- a/packages/webapi/state/callview.go
+++ b/packages/webapi/state/callview.go
@@ -39,7 +39,6 @@ func AddEndpoints(server echoswagger.ApiRouter, allChains chains.Provider) {
 		AddParamBody(dictExample, "params", "Parameters", false).
 		AddResponse(http.StatusOK, "Result", dictExample, nil)
 
-	// Deprecated
 	server.GET(routes.CallView(":chainID", ":contractHname", ":fname"), s.handleCallView).
 		SetSummary("Call a view function on a contract").
 		AddParamPath("", "chainID", "ChainID (base58-encoded)").
--- a/packages/webapi/state/callview.go
+++ b/packages/webapi/state/callview.go
@@ -31,6 +31,15 @@ func AddEndpoints(server echoswagger.ApiRouter, allChains chains.Provider) {
 
 	s := &callViewService{allChains}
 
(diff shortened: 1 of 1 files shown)


packages/webapi/state/callview.go
```diff
@@ -31,6 +31,15 @@ func AddEndpoints(server echoswagger.ApiRouter, allChains chains.Provider) {
...
 
+	server.POST(routes.CallView(":chainID", ":contractHname", ":fname"), s.handleCallView).
+		SetSummary("Call a view function on a contract").
+		AddParamPath("", "chainID", "ChainID (base58-encoded)").
+		AddParamPath("", "contractHname", "Contract Hname").
+		AddParamPath("getInfo", "fname", "Function name").
+		AddParamBody(dictExample, "params", "Parameters", false).
+		AddResponse(http.StatusOK, "Result", dictExample, nil)
+
+	// Deprecated
```
dessaya: Thanks for the pull request @mrbianchi !

I understand that having a GET request with a body is controversial, but nevertheless is allowed by the HTTP spec. 
Personally I think that a GET request is semantically better than POST here, because we are retrieving data and not causing any change in the server state. As an example, Elasticsearch also uses GET requests with body: https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html

Having said that, I also understand that it could cause issues with some external tools (e.g. caches, proxies) that do not expect GET+body, so it seems sensible to allow POST as well (that's what Elasticsearch does). I like the idea of accepting both GET and POST, but in that case I would not mark GET as deprecated.

On the other hand, if we are deprecating it, I think we should just remove it, and change all clients to use POST.

To sumarize, in order to merge this change I'd ask to either:
* not mark GET as deprecated, or
* remove support for GET, and update all clients to use POST

Thanks again!

mrbianchi: Thanks for your reply @dessaya ,
I already undid the mark and maybe contribute the last option in the future.

dessaya (approved):

PR #621, opened 2021-11-24
Fix: wasp-cli chain activate/deactivate

ivange94:
`wasp-cli chain activate` fails when run after a `wasp-cli chain deactivate` command. The error was traced down to https://github.com/iotaledger/wasp/blob/develop/packages/chains/chains.go#L184-L195. That method was designed to return only active chains and also has the side effect of removing inactive ones. 

Changing the implementation of that method forced me to update many parts of the code that have nothing to do with this fix and the cli. And aside from the many modifications, changing those implementations still needs to be discussed so I tried my best not to touch them.

I tried to isolate the fix only around the cli without affecting other parts of the application hence leading to some duplicate code in some places.

### DIFF
--- a/client/committee_record.go
+++ b/client/committee_record.go
@@ -27,16 +27,7 @@ func (c *WaspClient) GetCommitteeRecord(addr ledgerstate.Address) (*registry.Com
 // GetCommitteeForChain fetches the CommitteeRecord that manages the given chain
 func (c *WaspClient) GetCommitteeForChain(chainID *iscp.ChainID) (*registry.CommitteeRecord, error) {
 	res := &model.CommitteeRecord{}
-	if err := c.do(http.MethodGet, routes.GetCommitteeForChain(chainID.Base58()), nil, res); err != nil {
-		return nil, err
-	}
-	return res.Record(), nil
-}
-
-func (c *WaspClient) GetCommitteeForChainIncludeInActive(chainID *iscp.ChainID) (*registry.CommitteeRecord, error) {
-	res := &model.CommitteeRecord{}
-	path := routes.GetCommitteeForChain(chainID.Base58()) + "?includeInActive=true"
-	if err := c.do(http.MethodGet, path, nil, res); err != nil {
+	if err := c.do(http.MethodGet, routes.GetCommitteeForChain(chainID.Base58())+"?includeDeactivated=true", nil, res); err != nil {
 		return nil, err
 	}
 	return res.Record(), nil
--- a/client/committee_record.go
+++ b/client/committee_record.go
@@ -32,3 +32,12 @@ func (c *WaspClient) GetCommitteeForChain(chainID *iscp.ChainID) (*registry.Comm
 	}
 	return res.Record(), nil
 }
+
+func (c *WaspClient) GetCommitteeForChainIncludeInActive(chainID *iscp.ChainID) (*registry.CommitteeRecord, error) {
+	res := &model.CommitteeRecord{}
+	path := routes.GetCommitteeForChain(chainID.Base58()) + "?includeInActive=true"
+	if err := c.do(http.MethodGet, path, nil, res); err != nil {
+		return nil, err
+	}
+	return res.Record(), nil
+}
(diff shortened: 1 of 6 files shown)


tools/cluster/tests/wasp-cli_test.go
```diff
@@ -309,3 +309,31 @@ func TestWaspCLIBalance(t *testing.T) {
...
+	require.NotEmpty(t, chainID)
+	t.Logf("Chain ID: %s", chainID)
+
+	// test chain list command
+	out = w.Run("chain", "list")
+	require.Contains(t, out[0], "Total 1 chain(s)")
+	require.Contains(t, out[4], chainID)
+
+	w.Run("chain", "deactivate")
+	w.Run("chain", "activate")
```
jorgemmsilva: we could add extra checks here, to assert that deactivate/activate have the desired effect.

packages/webapi/admapi/committeerecord.go
```diff
@@ -91,7 +93,13 @@ func (s *committeeRecordService) handleGetCommitteeForChain(c echo.Context) erro
 	if err != nil {
 		return httperrors.BadRequest(err.Error())
 	}
-	chain := s.chains().Get(chainID)
+	inCludeInActive, err := strconv.ParseBool(c.QueryParam("includeInActive"))
```
jorgemmsilva:
```suggestion
	inCludeInActive, err := strconv.ParseBool(c.QueryParam("includeInActive"))
```

there is some funky capitalization going on... `inCludeInActive` is just a typo I think
`includeInActive` - means "include in active"? I think we wanted "include inactive" or "includeDeactivated"?

jorgemmsilva (commented):
This change feels a bit confusing tbh. I feel that we should remove the side effect from the existing `Get` function that you pointed out. Also maybe extend that function with a parameter (to include deactivated chains or not).

client/committee_record.go
```diff
@@ -32,3 +32,12 @@ func (c *WaspClient) GetCommitteeForChain(chainID *iscp.ChainID) (*registry.Comm
...
 }
+
+func (c *WaspClient) GetCommitteeForChainIncludeInActive(chainID *iscp.ChainID) (*registry.CommitteeRecord, error) {
+	res := &model.CommitteeRecord{}
+	path := routes.GetCommitteeForChain(chainID.Base58()) + "?includeInActive=true"
+	if err := c.do(http.MethodGet, path, nil, res); err != nil {
+		return nil, err
+	}
+	return res.Record(), nil
+}
```
dessaya: `InActive` is confusing because it reads like "in active". Can we rename to "Inactive"?

packages/chains/chains.go
```diff
@@ -194,3 +194,17 @@ func (c *Chains) Get(chainID *iscp.ChainID) chain.Chain {
 	}
 	return ret
 }
+
+// This is a workaround. Chains.Get is designed to exclude deactivated chains
+// and also remove them from the registry. This side effect makes it impossible to
+// reactivatew a deactivated chain. But Chains.Get is used in a lot of other places
```
dessaya: typo: `reactivatew`

packages/chains/chains.go
```diff
@@ -194,3 +194,17 @@ func (c *Chains) Get(chainID *iscp.ChainID) chain.Chain {
...
+// and any modifications like adding a new parameter or updating the return arguments
+// will require chainging how it's used in the other places. Looked at the code in these
+func (c *Chains) GetIncludeInActive(chainID *iscp.ChainID) chain.Chain {
```
dessaya: You could also modify the `Get` function to receive an "optional" parameter:

```go
func Get(chainID *iscp.ChainID, includeInactive ...bool) chain.Chain {
  if len(includeInactive) > 0 && includeInactive[0] {
    // include inactive
  } else {
    // default behavior
  }
}
```

dessaya (changes requested):

jorgemmsilva (approved):

dessaya (approved):

PR #625, opened 2021-11-25
Blocklog refactoring

lmoe:
* Wrap address serializer into function (BytesFromAddress)
* General refactoring to new structure
* Change Request to RequestData to use Bytes/String functions

### DIFF
--- a/packages/iscp/requestimpl.go
+++ b/packages/iscp/requestimpl.go
@@ -13,9 +13,14 @@ import (
 	"github.com/iotaledger/wasp/packages/kv/dict"
 )
 
-// RequestDataFromBytes read RequestData from byte stream. First byte is interpreted as boolean flag if its an off-ledger request data
 func RequestDataFromBytes(data []byte) (RequestData, error) {
 	mu := marshalutil.New(data)
+
+	return RequestDataFromMarshalUtil(mu)
+}
+
+// RequestDataFromMarshalUtil read RequestData from byte stream. First byte is interpreted as boolean flag if its an off-ledger request data
+func RequestDataFromMarshalUtil(mu *marshalutil.MarshalUtil) (RequestData, error) {
 	isOffLedger, err := mu.ReadBool()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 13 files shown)


packages/vm/core/blocklog/interface.go
```diff
@@ -357,19 +358,20 @@ func ControlAddressesFromMarshalUtil(mu *marshalutil.MarshalUtil) (*ControlAddre
...
 func (ca *ControlAddresses) String() string {
 	var ret string
-	if ca.StateAddress.Equals(ca.GoverningAddress) {
-		ret = fmt.Sprintf("ControlAddresses(%s), block: %d", ca.StateAddress.Base58(), ca.SinceBlockIndex)
+	if ca.StateAddress.Equal(ca.GoverningAddress) {
+		ret = fmt.Sprintf("ControlAddresses(%s), block: %d", ca.StateAddress.String(), ca.SinceBlockIndex)
 	} else {
 		ret = fmt.Sprintf("ControlAddresses(%s, %s), block: %d",
-			ca.StateAddress.Base58(), ca.GoverningAddress.Base58(), ca.SinceBlockIndex)
+			ca.StateAddress.String(), ca.GoverningAddress.String(), ca.SinceBlockIndex)
```
jorgemmsilva: we probably want `bech32` instead of `string` for addresses

packages/iscp/requestimpl.go
```diff
@@ -47,6 +52,15 @@ type OffLedgerRequestData struct {
 	gasBudget      uint64
 }
 
+func NewOffLedgerRequest(chainID ChainID, contract, entryPoint Hname) *OffLedgerRequestData {
+	return &OffLedgerRequestData{
+		chainID:    chainID,
+		contract:   contract,
+		entryPoint: entryPoint,
+		nonce:      uint64(time.Now().UnixNano()),
```
jorgemmsilva: having the nonce default to unix timestamp seems dangerous, we established that it should be monotonically increasing (1,2,3,4,5). maybe we can keep the timestamp for testing purposes, but there should be a parameter to override it

jorgemmsilva (approved):

PR #627, opened 2021-11-25
txbuilder: dust deposit support and tests 

lunfardo314:


packages/vm/vmcontext/runreq.go
```diff
@@ -80,6 +84,24 @@ func (vmctx *VMContext) creditAssetsToChain() {
...
+	// vmtxbuilder.ErrNotEnoughFundsForInternalDustDeposit and the request will be skipped
+	switch {
+	case adjustmentOfTheCommonAccount > 0:
+		vmctx.creditToAccount(commonaccount.Get(vmctx.ChainID()), &iscp.Assets{
+			Iotas: uint64(adjustmentOfTheCommonAccount),
+		})
+	case adjustmentOfTheCommonAccount < 0:
+		err := util.CatchPanicReturnError(func() {
+			vmctx.debitFromAccount(commonaccount.Get(vmctx.ChainID()), &iscp.Assets{
+				Iotas: uint64(-adjustmentOfTheCommonAccount),
```
jorgemmsilva, line 98: right now I'm considering a positive number will be passed to `debitFromAccount`, but I can use absolute values or reconsider this

packages/vm/vmcontext/vmtxbuilder/totals.go
```diff
@@ -0,0 +1,113 @@
...
+			s, ok := ret.TokenBalances[nt.ID]
+			if !ok {
+				s = new(big.Int)
+			}
+			s.Add(s, nt.Amount)
+			ret.TokenBalances[nt.ID] = s
+		}
+	}
+	return
+}
```
jorgemmsilva: it seems a good chunk of code is repeated here, would it make sense to extract to a single function?

packages/vm/vmcontext/runreq.go
lunfardo314: I think it is fine what we have now, always have `uint64`. The adjustment can be negative and this is the only place in the code where we have to consider it.

packages/vm/vmcontext/vmtxbuilder/totals.go
lunfardo314: optimized the code according to your suggestion. I think readability suffers now. Because actually the codes chunks are not identical, there are subtle differences in 2 places. Anyway, tests pass :)

jorgemmsilva (approved):

PR #629, opened 2021-11-25
Wasmvm

BugFreeSoftware:
Implemented Bool/Int8/Uint8/Uint16/Uint32/Uint64 type support in all Wasm VM languages


contracts/wasm/fairroulette/client/events.ts
```diff
@@ -0,0 +1,114 @@
...
+
+type Address = string;
+type AgentID = string;
+type ChainID = string;
+type Color = string;
+type Hash = string;
+type Hname = string;
+type Int16 = number;
+type Int32 = number;
+type Int64 = bigint;
```
lunfardo314: Bigint probably has to be big.Int. It is coming with new VM and native tokens so maybe it makes sense  to reserve the name for the real thing

lunfardo314 (approved):

contracts/wasm/fairroulette/client/events.ts
BugFreeSoftware: Yeah I only just started on the client side code. Haven't even set up a debugging system for it yet.
But this is the JS bigint type. It's the only type that can represent int64 since number in JS is always stored as float. If we can do bigint with the VM in the future then it will probably be named Bigint or BigInt in WasmLib to be consistent with the rest of the naming.

PR #630, opened 2021-11-25
Update docker.md

brunnerpascal:
Added solution for not being able to access the dashboard via docker


lmoe:
Thanks for your contribution!

Unfortunately, I would strongly advise against publishing all ports with the --host=net argument as you open up all ports - also the private ones to the public. 

This requires you to close them down again in your firewall and it could add extra holes to the system. 

The best way to expose the ports is by using `--publish` (https://docs.docker.com/config/containers/container-networking/#published-ports) or by creating a separate docker network which you can connect with other services. For example a http proxy or GoShimmer.

PR #634, opened 2021-11-26
refactor: (startdust-vm) accounts 

jorgemmsilva:


packages/iscp/agentid.go
```diff
@@ -41,12 +41,9 @@ func NewAgentID(addr iotago.Address, hname Hname) *AgentID {
 
 func BytesFromAddress(address iotago.Address) []byte {
 	addressInBytes, err := address.Serialize(serializer.DeSeriModeNoValidation, nil)
-
 	if err != nil {
-		// TODO: Ignoring errors here, as no validation takes place anyway. Maybe needs further discussion?
 		return nil
```
dessaya, line 45: I know it's not technically part of the PR, but: why not panic here? Or just do `bytes, _  := address.Serialize(...)`.

Also, I think `address.Key()` returns the same data as a `string`, so maybe this function is not really necessary?

packages/iscp/assets.go
```diff
@@ -15,13 +17,45 @@ type Assets struct {
 	Tokens iotago.NativeTokens
 }
 
+var IotaTokenID = []byte{0}
```
dessaya, line 21: Mixing iotas with native tokens in a single bag might be prone to bugs. For example, is it guaranteed that this cannot collide with a valid native token id? Or, someone might call `ToDict()` and then forget that the dict includes some iotas, and iterate through it thinking that it only contains native tokens.

Instead of: 

```
NewAssetsFromDict(d dict.Dict)
ToDict() dict.Dict
```

why not:

```
NewAssetsFromDict(iotas uint64, native dict.Dict)
ToDict() (uint64, dict.Dict)
```

packages/iscp/assets.go
```diff
@@ -15,13 +17,45 @@ type Assets struct {
...
 }
 
+func NewAssetsFromDict(d dict.Dict) (*Assets, error) {
+	ret := NewEmptyAssets()
+	var err error
+	ret.Iotas = new(big.Int).SetBytes(d.MustGet(kv.Key(IotaTokenID))).Uint64()
+	if err != nil {
+		return ret, err
+	}
+	d.Del(kv.Key(IotaTokenID))
```
dessaya: Modifying the received dict might be surprising behavior for the caller

packages/vm/core/accounts/impl.go
```diff
@@ -116,38 +138,51 @@ func withdraw(ctx iscp.Sandbox) (dict.Dict, error) {
...
 		return nil, nil
 	}
-	tokensToSend := bals
-	if !harvestAll {
-		balCol := bals[col]
-		a.Require(balCol >= amount, "accounts.harvest.error: not enough tokens")
-		tokensToSend = colored.NewBalancesForColor(col, amount)
+
+	tokensToSend := iscp.NewEmptyAssets()
+	if bytes.Compare(assetID, iscp.IotaTokenID) == 0 {
```
dessaya: bytes.Equal?

Also, maybe this could be a function like `iscp.IsIota(tokenID)`

dessaya (approved):

packages/iscp/agentid.go
jorgemmsilva: > why not panic here? Or just do bytes, _ := address.Serialize(...).

if we return nil its always possible to verify if the serialization was successful

> Also, I think address.Key() returns the same data as a string, so maybe this function is not really necessary?

this function is to return a serialized version (byte slice), not a string - using the standard serialization stuff from iotago seems reasonable

packages/iscp/assets.go
jorgemmsilva: iota is mapped to `0` - there will never be a token with this ID.

I opted to keep iotas in the dictionary because its easy to map to the internal accounts implementation, also everything that used the old `colored` stuff is still compatible.

Using a single dict is easy to extend when we need to add NFTs, just more entries to the same dict, will be possible to differentiate by checking the length of the key (38 bytes for tokens, 20 for NFTs), 

this seems fine to me, but I'm happy to discuss further

packages/iscp/assets.go
jorgemmsilva: nice catch, thanks! :+1:

packages/iscp/agentid.go
dessaya: >> why not panic here? Or just do bytes, _ := address.Serialize(...).

> if we return nil its always possible to verify if the serialization was successful

But it's using `DeSeriModeNoValidation`, then it should never fail, right? In fact, looking at the implementation of `Serialize` I think it never fails, it even ignores the serialization mode :stuck_out_tongue: 

>> Also, I think address.Key() returns the same data as a string, so maybe this function is not really necessary?

> this function is to return a serialized version (byte slice), not a string - using the standard serialization stuff from iotago seems reasonable

If I understand correctly, the string returned by `Key()` is not actually a printable string; it's just the bytes converted to `string`, so it's the same data just in a different type.

Also, seeing that this function is used to store the address as a map key... `address.Key()` is designed for this use case I think.

PR #646, opened 2021-12-01
Governance access nodes

kape1395:


packages/vm/core/governance/governanceimpl/accessnodesImpl.go
```diff
@@ -0,0 +1,119 @@
...
+		accessNodes.MustDelAt(paramsPubKey)
+	}
+	return nil, nil
+}
+
+// SC Command Function handler.
+// Can only be invoked by the chain owner.
+//
+//  changeAccessNodes(
+//    actions: map(pubKey => {0:remove, 1:accept, 2:drop})
```
dessaya: make these constants / enum?

```
type Action byte
const (
  actionRemove = Action(iota)
  actionAccept
  actionDrop
)
```

packages/vm/core/governance/governanceimpl/impl.go
```diff
@@ -55,5 +64,9 @@ func initialize(ctx iscp.Sandbox) (dict.Dict, error) {
 	if feeColorSet {
 		state.Set(governance.VarFeeColor, codec.EncodeColor(feeColor))
 	}
+
+	collections.NewMap(state, governance.VarAccessNodeCandidates).Erase()
+	collections.NewMap(state, governance.VarAccessNodes).Erase()
+	collections.NewArray16(state, governance.VarValidatorNodes).MustErase()
```
dessaya: why call Erase here?

dessaya (approved):

packages/vm/core/governance/governanceimpl/impl.go
kape1395: I wanted to have the state variables and their types mentioned here explicitly.
Such call will actually do nothing. Maybe there is a better way to declare variables and their intended data types?

packages/vm/core/governance/governanceimpl/accessnodesImpl.go
kape1395: Good point. Will change.

packages/vm/core/governance/governanceimpl/impl.go
dessaya: We don't have a standard way of doing that. What I like to do is having functions like:

```
func getAccessNodeCandidatesMap(state kv.KVStore) *collections.Map {
  return collections.NewMap(state(), governance.VarAccessNodeCandidates)
}

func getAccessNodeCandidatesMapR(state kv.KVStoreReader) *collections.ImmutableMap {
  return collections.NewMapReadOnly(state(), governance.VarAccessNodeCandidates)
}

func getAccessNodesMap(state kv.KVStore) *collections.Map {
  return collections.NewMap(state(), governance.VarAccessNodes)
}

func getAccessNodesMapR(state kv.KVStoreReader) *collections.ImmutableMap {
  return collections.NewMapReadOnly(state(), governance.VarAccessNodes)
}

func getValidatorNodesArray(state kv.KVStore) *collections.Array16 {
  return collections.NewArray16(state(), governance.VarValidatorNodes)
}

func getValidatorNodesArrayR(state kv.KVStoreReader) *collections.ImmutableArray16 {
  return collections.NewArray16ReadOnly(state(), governance.VarValidatorNodes)
}
```

kape1395: ok, removed those `Erase` here.

packages/vm/core/governance/governanceimpl/accessnodesImpl.go
```diff
@@ -0,0 +1,117 @@
...
+	params := kvdecoder.New(ctx.Params(), ctx.Log())
+	paramCandidate := params.MustGetBool(governance.ParamCandidateNodeCandidate, true)
+
+	ani := governance.AccessNodeInfo{
+		NodePubKey:    params.MustGetBytes(governance.ParamCandidateNodePubKey),
+		ValidatorAddr: ctx.Request().SenderAddress().Bytes(), // Not from params, to have it validated.
+		Certificate:   params.MustGetBytes(governance.ParamCandidateNodeCertificate),
+		ForCommittee:  params.MustGetBool(governance.ParamCandidateNodeForCommittee, false),
+		AccessAPI:     params.MustGetString(governance.ParamCandidateNodeAccessAPI, ""),
+	}
```
jorgemmsilva: suggestion: we could receive the entire `AccessNodeInfo` as a parameter, right? no need to split each property into a separate param.

packages/vm/core/governance/governanceimpl/impl.go
```diff
@@ -29,6 +32,11 @@ var Processor = governance.Contract.Processor(initialize,
 	governance.FuncGetChainInfo.WithHandler(getChainInfo),
 	governance.FuncSetChainInfo.WithHandler(setChainInfo),
 	governance.FuncGetMaxBlobSize.WithHandler(getMaxBlobSize),
+
+	// access nodes.
+	governance.FuncGetChainNodes.WithHandler(getChainNodesFuncHandler),
+	governance.FuncCandidateNode.WithHandler(candidateNodeFuncHandler),
```
jorgemmsilva: its kinda hard to tell what this function does by name: `candidateNodeFuncHandler`
I propose changing the name to something more descriptive like: `addCandidateNode`

jorgemmsilva (approved):

packages/vm/core/governance/governanceimpl/accessnodesImpl.go
kape1395: Yes, I was thinking about that option. The structure is not exactly the same. The parameters don't include the ValidatorAddr field.

Another possible problem is that the parameter names are then hidden from other parameters, thus making it easier to make them coinciding. E.g. `ParamCandidateNodeCandidate` would be defined in one place (function params), and other parameters in the AccessNodeInfo structure. In the dictionary they would be on the same level. Not sure, which way is better.

packages/vm/core/governance/governanceimpl/impl.go
kape1395: Renamed.

packages/vm/core/governance/governanceimpl/accessnodesImpl.go
kape1395: Updated to use the AccessNodeInfo struct for params, revoke function extracted to a separate one.

PR #647, opened 2021-12-01
Implement storage of anchor ids into their respective blocks

lmoe:


packages/vm/core/blocklog/internal.go
```diff
@@ -23,6 +24,29 @@ func SaveNextBlockInfo(partition kv.KVStore, blockInfo *BlockInfo) uint32 {
 	return ret
 }
 
+func SetAnchorTransactionIdOfLastBlock(partition kv.KVStore, transactionId ledgerstate.TransactionID) {
```
dessaya: Nit: `Id` -> `ID` (standard idiomatic Go naming; I believe the linter should raise a warning)

Nit: `Last` -> `Latest` (last could mean that there will be no more)

packages/vm/core/blocklog/internal.go
```diff
@@ -23,6 +24,29 @@ func SaveNextBlockInfo(partition kv.KVStore, blockInfo *BlockInfo) uint32 {
...
+	blockInfo, _ := BlockInfoFromBytes(lastBlockIndex, blockInfoBuffer)
+
+	blockInfo.AnchorTransactionID = transactionId
+
+	registry.MustSetAt(lastBlockIndex, blockInfo.Bytes())
+}
+
+func GetAnchorTransactionIdByBlockIndex(partition kv.KVStore, blockIndex uint32) (iotago.TransactionID, error) {
+	registry := collections.NewArray32(partition, StateVarBlockRegistry)
+	blockInfoBuffer := registry.MustGetAt(blockIndex)
```
dessaya, line 40: Maybe there should be a bounds check? What if blockIndex is >= array size?

packages/vm/core/blocklog/internal.go
```diff
@@ -23,6 +24,29 @@ func SaveNextBlockInfo(partition kv.KVStore, blockInfo *BlockInfo) uint32 {
...
+	blockInfo, err := BlockInfoFromBytes(blockIndex, blockInfoBuffer)
+
+	if err != nil {
+		return iotago.TransactionID{}, err
```
dessaya: I think it's safe to panic here; decoding the block info bytes should succeed, unless the data is corrupt or something.

packages/vm/core/blocklog/internal.go
jorgemmsilva: it will panic if that happens, should be fine - we never should pass an invalid blockIndex here

packages/vm/vmcontext/vmcontext.go
```diff
@@ -177,6 +177,9 @@ func (vmctx *VMContext) mustSaveBlockInfo(numRequests, numSuccess, numOffLedger
 		PreviousStateHash:     prevStateData.Commitment,
 	}
 
+	// TODO: Make sure that AnchorOutputID.TransactionID == vmctx.chainInput.ID().TransactionID()
```
jorgemmsilva: confused about this TODO here :thinking: what is `vmctx.chainInput`?

packages/vm/vmcontext/txbuilder.go
```diff
@@ -58,11 +58,16 @@ func (vmctx *VMContext) getNativeTokenUtxoIndex(id iotago.NativeTokenID) (stateI
...
 
-	// TODO get correct TXID for the given state index.
-	panic("implement me")
-	return nil
+	transactionId, err := blocklog.GetAnchorTransactionIdByBlockIndex(vmctx.State(), stateIndex)
+
+	if err != nil {
+		// TODO: Panic, or rather return error?
+		panic("?")
+	}
```
jorgemmsilva: could remove the error check here and just panic inside `GetAnchorTransactionIdByBlockIndex`. That func should never fail. if it does that means our logic has a bug :p

jorgemmsilva (approved):

packages/vm/vmcontext/vmcontext.go
lmoe: I was doing that implementation on the develop branch first to be able to test it properly. 
On dev the property is `chainInput` that holds the TxID, now I guess it's `AnchorOutputID.TransactionID`. Just wanted to make sure that I'm right on this one. :)

PR #652, opened 2021-12-02
Add docker dev net setup

lmoe:
This PR introduces a simple/lite devnet setup based on Docker Compose that enables users to spawn a Wasp+GoShimmer combination with one simple `docker-compose up`. 

It is expected to not require any further configuration to make the first steps as easy as possible.

One addition to the setup is the possibility to disable the admin whitelist as it didn't support Docker internal networking so well if the admin port is published to the outside.


packages/webapi/admapi/endpoints.go
```diff
@@ -36,7 +37,11 @@ func AddEndpoints(
 ) {
 	initLogger()
 
-	adm.EchoGroup().Use(protected(adminWhitelist))
+	echoGroup := adm.EchoGroup()
+
+	if !parameters.GetBool(parameters.WebAPIAdminWhitelistDisabled) {
```
jorgemmsilva: this is hard to read, usually is more understandable to make bool the "enabled" and do `!somethingEnalbed`

lmoe: Yeah I would generally agree. I just wanted to keep it in the same way as the other config params like:

* LoggerDisableEvents
* LoggerDisableStacktrace
* LoggerDisableCaller

How about something like putting it into a variable to read it easier?

```
isWhitelistEnabled := !parameters.GetBool(parameters.WebAPIAdminWhitelistDisabled)

echoGroup := adm.EchoGroup()

if isWhitelistEnabled {
	echoGroup.Use(protected(adminWhitelist))
}
```

tools/devnet/goshimmer.config.json
```diff
@@ -0,0 +1,87 @@
...
+    "webapi": {
+      "basic_auth": {
+        "enabled": false,
+        "username": "goshimmer",
+        "password": "goshimmer"
+      },
+      "bindAddress": "0.0.0.0:8080"
+    },
+    "faucet": {
+      "powDifficulty": 12,
```
jorgemmsilva, line 83: shouldn't this be -1 or 0?

jorgemmsilva (approved):

tools/devnet/goshimmer.config.json
lmoe: This is the config for GoShimmer. -1 was set on the waspcli side to use the difficulty of GoShimmer.

PR #653, opened 2021-12-03
Node connection metrics collection and retrieval

Juliusan:
The aim of this PR is to collect some statistics about communication with L1 and provide it to the user through Prometheus, wasp web service, cli and dashboard. Now for every chain and every type of message passed to and received from L1 the following data are collected:
* Total count of the messages transmitted
* Last time the message was transmitted
* Full last transmitted message

Additionally the chain addresses, that are subscribed to the L1, are stored. Only total count and last time of message transmission are passed to Prometheus. However all of the data are available through web service at addresses:
* `/adm/chain/nodeconn/metrics` - the cumulative statistics for the whole nodeconn
* `/adm/chain/<chainID>/nodeconn/metrics` - the statistics for given chain ID.

In order to reach this goal, Goshimmer's `txstream` object is no longer used directly, but it is wrapped in wasp object implementing `NodeConnection` interface. Each chain has it's own `ChainNodeConnection` implementation, which internally stores the chainID and may be used to send messages to L1 or to attach to/detach from receiving messages sent by L1. Each component is now responsible to start AND stop listening for L1 messages it is interested in.

In addition to the main change, the following smaller corrections were made:
* each component, which attaches to listening to any external message (from L1 or other peer) now detaches from it on close.
* `chainimpl` methods got separated to different files by interface they implement. I think, this increases code readability.
* less inner structures are exported from `chainimpl` object: there is no need to export event objects, it is enough to export several methods, which uses the objects inside `chainimpl`.


kape1395 (approved):

packages/chain/chain.go
```diff
@@ -54,18 +53,19 @@ type ChainEntry interface {
...
-	EventRequestProcessed() *events.Event
+	AttachToRequestProcessed(func(iscp.RequestID)) (attachID *events.Closure)
+	DetachFromRequestProcessed(attachID *events.Closure)
 }
 
-type ChainEvents interface {
-	RequestProcessed() *events.Event
-	ChainTransition() *events.Event
+type ChainMetrics interface {
+	GetNodeConnectionMetrics() nodeconnmetrics.NodeConnectionMessagesMetrics
```
ivange94, line 61: Any reason why the existing `ChainMetrics` interface in the `metrics` package could not be reused here? With the current architecture of how we gather metrics in `wasp`, what I would have expected was a 

`type NodeConnectionMetrics interface {}` defined in the `metrics` package and then this added to the `ChainMetrics` interface like it's done here with other ChainMetrics subtypes 

```
type ChainMetrics interface {
	CountMessages()
	CountRequestAckMessages()
	MempoolMetrics
	ConsensusMetrics
	StateManagerMetrics
}
```

 https://github.com/iotaledger/wasp/blob/develop/packages/metrics/chain.go#L15-L21. The idea was to have all metrics collection code in one place making it easy to maintain.

ivange94 (commented):
Why do we need separate endpoints to expose nodecon metrics? Why not expose everything under `/metrics`? I think the `Vec` types in prometheus would have served to add additional information that could be used in filtering, for example 

`c.metrics.processedRequestCounter.With(prometheus.Labels{"chain": c.chainID.String()}).Inc()`

This will generate a metrics that has a label `chain: ID_OF_CHAIN`. This way for a single server we give prometheus only one endpoint instead of 3. With `promql`, you can write queries that retrieve the exact information you need.

packages/chain/chain.go
Juliusan: This is just my suggestion. I saw what you did and I agree that there should be only one way of doing it. However, I am afraid, that if we continue that way, we will have a huge collection of everything in one place and it would be hard to sort everything out and understand the code later. Currently each of the subtypes collects at most 5 things (export 5 methods). I wanted to add 8*3=24 methods, which would be much more than currently are exported in total. `nodeconnmetrics` package currently contains 333 lines of code, whereas both `metrics.go` and `chain.go` in `metrics` package only 158. I think we might get lost if we put everything in several huge files.

Juliusan:
> Why do we need separate endpoints to expose nodecon metrics? Why not expose everything under `/metrics`? I think the `Vec` types in prometheus would have served to add additional information that could be used in filtering, for example
> 
> `c.metrics.processedRequestCounter.With(prometheus.Labels{"chain": c.chainID.String()}).Inc()`
> 
> This will generate a metrics that has a label `chain: ID_OF_CHAIN`. This way for a single server we give prometheus only one endpoint instead of 3. With `promql`, you can write queries that retrieve the exact information you need.

We don't. I didn't know we have such an endpoint. I'll check that.

packages/chain/chain.go
kape1395: I think a struct with a huge number of functions will cause misunderstanding, what those metrics mean.

With sub-objects we could have a better structure, which can represent each component in the system and also various arrays of sub-components as well. E.g. each chain can have its metrics, and each SC in a chain can have own metrics as well. I'm not sure it having a flat structure for that is really adds to the clarity.

ivange94 (approved):

dessaya (approved):

kape1395 (approved):

PR #670, opened 2021-12-09
Update docker documentation

lmoe:


documentation/sidebars.js
```diff
@@ -216,8 +216,13 @@ module.exports = {
...
                 {
                     type: 'doc',
-                    label: 'Using Docker',
-                    id: 'guide/chains_and_nodes/docker',
+                    label: 'Using Docker (Standalone)',
+                    id: 'guide/chains_and_nodes/docker_standalone',
+                },
+                {
+                    type: 'doc',
+                    label: 'Using Docker (Preconfigured)',
```
jorgemmsilva: We should probably make it obvious that this is only for development purposes. something like:
```suggestion
                    label: 'Using Docker Development-Setup (Preconfigured)',
```

Also this should probably be outside "ISCP Chains and Nodes" section, and moved to a separate (new) "Development Tools" section.

we could then link it on the main docs: "for development purposes we recommend using this pre-configured docker setup"

jorgemmsilva (approved):
LGTM, just think this should be moved to a dedicated "development" section

PR #677, opened 2021-12-09
implement codec/address functions

dessaya:


packages/kv/codec/address.go
```diff
@@ -1,22 +1,37 @@
...
+		return nil, err
+	}
+	return addr, nil
 }
 
-func EncodeAddress(value iotago.Address) []byte {
-	panic("refactor me")
-	//return value.Bytes()
+func EncodeAddress(addr iotago.Address) []byte {
+	addressInBytes, err := addr.Serialize(serializer.DeSeriModeNoValidation, nil)
```
lmoe, line 32: We have a function "BytesFromAddress" that does exactly this for you. It's currently inside `agentid.go`. Might be worth to move it or to refactor all other usages of "BytesFromAddress" to your EncodeAddress method.

lmoe (approved):

packages/kv/codec/address.go
lunfardo314: yes, let's get rid of that "BytesFromAddress"

lunfardo314 (approved):

PR #681, opened 2021-12-10
draft VM gas implementation

jorgemmsilva:


packages/iscp/gas/constants.go
```diff
@@ -1,20 +1,46 @@
 package gas
 
-const (
+var (
```
dessaya: Why var? Most, if not all things here look like constants, and the functions can be declared as regular functions :thinking:

lmoe (approved):

packages/kv/subrealm/subrealm.go
```diff
@@ -1,79 +1,107 @@
...
 )
 
+type BurnGasFn = func(gas uint64)
+
 type subrealm struct {
-	kv     kv.KVStore
-	prefix kv.Key
+	kv      kv.KVStore
+	prefix  kv.Key
+	burnGas BurnGasFn
```
dessaya: This looks like adding an unrelated responsibility to this very simple object. In general, the original intention of the `kv` package is to be as independent from iscp as possible. It is "just" a key value store.

Could the burnt gas be calculated from the list of mutations instead? I mean, after the contract gets executed it should be possible to examine the list of set/del mutations.

Another possibility is to create a wrapper (much like this one):

```
type KVStoreWithGas struct {
  kv kv.KVStore
  burnGas BurnGasFn
}
```

in which the wrapped kvstore can be a `subrealm` or a `Dict` or whatever.

jorgemmsilva: I think the gas needs to be deducted on each operation, so we can stop the execution as soon as the request runs out of gas

packages/iscp/gas/constants.go
jorgemmsilva: good point, I just liked the way it  was organized, so I just did this quick change, but you're right 😛 . Changing it

packages/kv/subrealm/subrealm.go
jorgemmsilva: You're right regarding the separation of concerns 👍 I'll create a `KVStoreWithGas` as you suggested

dessaya (approved):

packages/kv/subrealm/subrealm.go
lunfardo314: Diego is right, why would we implement it in the generic kvstore. Good place for it is in the implementation of kvstore in  the VMContext. It implements its own kvstore for additional buggering and catching optimistic read failures anyway

lunfardo314: no need even for additional implementation, it is a responsability of the VM

PR #690, opened 2021-12-16
Fixed wasp-cli offledger nonce always zero

BugFreeSoftware:


tools/wasp-cli/chain/postrequest.go
```diff
@@ -34,6 +35,7 @@ func postRequestCmd() *cobra.Command {
 			scClient := SCClient(iscp.Hn(args[0]))
 
 			if offLedger {
+				params.Nonce = uint64(time.Now().UnixNano())
```
jorgemmsilva, line 38: nonces should be monotonic strictly increasing. Unix timstamp works, but if many requests are sent in a short amount of time, some will likely fail.

BugFreeSoftware: Not in this case. It's a command line utility. Even when sending as batch the unix nano time will be sufficiently different.

lmoe (approved):

PR #693, opened 2021-12-17
Cherry-picked WasmVM updates to 0.2.2

BugFreeSoftware:


luke-thorne:
We should be able to close this now as master is stable again and should include these changes. Any objections @BugFreeSoftware?

PR #697, opened 2021-12-23
Make nodeconn metrics dashboard view more beautiful

Juliusan:
Nodeconn metrics dashboard view does not display full message, but only 100 first symbols. The full message is displayed as a tooltip.


fijter (approved):
lgtm

PR #699, opened 2021-12-25
working on foundries and minting

lunfardo314:


packages/iscp/requestimpl.go
```diff
@@ -412,8 +412,11 @@ func (r *OnLedgerRequestData) SenderAddress() iotago.Address {
 	return senderBlock.Address
 }
 
-func (r *OnLedgerRequestData) CallTarget() CallTarget {
-	return CallTarget{
+func (r *OnLedgerRequestData) CallTarget() *CallTarget {
+	if r.requestMetadata == nil {
+		return nil
+	}
```
dessaya, line 418: Why would `requestMetadata` be nil? If `requestMetadata` is somehow not provided or not found when constructing the `OnLedgerRequestData`, isn't it better to just rais an error at that point instead?

Isn't impractical to return nil here, since every caller of `CallTarget()` now has to check for nil?

packages/kv/codec/int.go
```diff
@@ -120,3 +122,21 @@ func DecodeUint64(b []byte, def ...uint64) (uint64, error) {
...
+		if len(def) == 0 {
+			return nil, xerrors.Errorf("cannot decode nil bytes")
+		}
+		return def[0], nil
+	}
+	ret := big.NewInt(0).SetBytes(b)
+	return ret, nil
+}
+
+func EncodeBigIntAbs(value *big.Int) []byte {
```
dessaya, line 137: Why is it called `EncodeBigIntAbs` and not just `EncodeBigInt`? :thinking:

packages/kv/codec/int.go
```diff
@@ -120,3 +122,21 @@ func DecodeUint64(b []byte, def ...uint64) (uint64, error) {
...
+	if value == nil {
```
dessaya, line 138: Why accept nil here as valid? The semantics of `big.Int` do not accept nil as the zero value.

dessaya (approved):

packages/iscp/requestimpl.go
lunfardo314: that is because request metadata block may be absent in the UTXO and we have to be prepared for this situation.
In particular, the `ExtendedOutput` is recognized as an internal UTXO used for accounting of native assets (not a on-ledger request) when two conditions met:
- the sender is the chain itself
- it does not have metadata

If the sender is itself and with metadata, it means it is a self-posted request. If it does not have metadata and sender is from outside, it is processed as a special request with fake zero target, which means all assets will go to the sender's account and some gas will be charged due to panic

packages/kv/codec/int.go
lunfardo314: because methods of `big.Int` `SetBytes()` and `Bytes()` ignore sign (it must be handled separately by user logic). I was confused, so in order not to forget, I put `Abs`

packages/kv/codec/int.go
lunfardo314: hm, I don't know. I thought it accepts `nil` as zero in some operations at least. I agree, all this umbiguous semantics must be aligned everywhere

PR #704, opened 2021-12-28
Governance access nodes

kape1395:
Maintain the access nodes based on the governance smart contract.
  - Chain nodes are taken from the DKShare.
  - CommitteeRecord removed (covered by the DKShare).
  - ChainRecord has no node list anymore (covered by the DKShare).
  - Access Node information is taken from the Governance SC.
  - Access Nodes are confirmed automatically, if the application sender is the chain owner.
  - PeerNetworkConfig structure removed (covered by the peering.Domain.
  - All the peering is addressed by the node public keys instead of NetIDs.


packages/chain/chain.go
```diff
@@ -222,7 +227,6 @@ type CommitteeInfo struct {
 type PeerStatus struct {
 	Index     int
 	PeeringID string
```
Juliusan: Shouldn't here be a public key too?

packages/chain/statemgr/domain_with_fallback.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/iotaledger/hive.go/crypto/ed25519"
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/wasp/packages/peering"
+	"golang.org/x/xerrors"
+)
+
+// DomainWithFallback acts as a peering domain, but maintains 2 sets of peers,
+// one for the normal operation (matches with the chain peers) and other for
+// the fallback mode (include all the trusted peers).
+type DomainWithFallback struct {
```
Juliusan, line 19: Is `DomainWithFallback` supposed to implement `peering.PeerDomainProvider`? Should mention it here:
```var _ peering.PeerDomainProvider = &DomainWithFallback{}```

File: packages/chain/chainimpl/chainimpl.go [hunk 3]
```diff
@@ -142,12 +142,19 @@ func NewChain(
...
-	ret.chainPeers, err = netProvider.PeerDomain(chainID.Array(), peerNetConfig.Neighbors())
+	chainPeerNodes := []*ed25519.PublicKey{netProvider.Self().PubKey()}
+	ret.chainPeers, err = netProvider.PeerDomain(chainID.Array(), chainPeerNodes)
 	if err != nil {
-		log.Errorf("NewChain: %v", err)
+		log.Errorf("NewChain: unable to create chainPeers domain: %v", err)
 		return nil
 	}
-	ret.stateMgr = statemgr.New(db, ret, ret.chainPeers, ret.nodeConn, chainMetrics)
+	stateMgrDomain, err := statemgr.NewDomainWithFallback(chainID.Array(), netProvider, log.Named("sm"))
```
Juliusan, line 151: Should this code be in `chainimpl`? Can't state manager create it's own domain?
```ret.stateMgr = statemgr.New(db, ret, netProvider, ret.nodeConn, chainMetrics)```

**Review by Juliusan — APPROVED** [review]

File: packages/apilib/deploychain.go [hunk 4]
```diff
@@ -16,38 +16,28 @@ import (
...
+	T                 uint16
+	OriginatorKeyPair *ed25519.KeyPair
+	Description       string
+	Textout           io.Writer
+	Prefix            string
 }
 
 // DeployChainWithDKG performs all actions needed to deploy the chain
 // TODO: [KP] Shouldn't that be in the client packages?
 func DeployChainWithDKG(par CreateChainParams) (*iscp.ChainID, ledgerstate.Address, error) {
```
lunfardo314, line 38: yes, client packages would be a proper places. But it is in the `apilib`, no?

packages/chain/chainimpl/eventproc.go
```diff
@@ -161,53 +165,59 @@ func (c *chainObj) rotateCommitteeIfNeeded(anchorOutput *ledgerstate.AliasOutput
...
 	}
-	// just in case check if I am among committee nodes
-	// should not happen
-	if !util.StringInList(c.peerNetworkConfig.OwnNetID(), rec.Nodes) {
+	if !found {
 		return nil, xerrors.Errorf("createCommitteeIfNeeded: I am not among nodes of the committee record. Inconsistency")
 	}
-	return rec, nil
+	return cmtDKShare, nil
 }
```
jorgemmsilva: nit: the "found" logic could be removed
```go
func (c *chainObj) getChainDKShare(addr ledgerstate.Address) (*tcrypto.DKShare, error) {
	//
	// just in case check if I am among committee nodes
	// should not happen
	selfPubKey := c.netProvider.Self().PubKey()
	cmtDKShare, err := c.dksProvider.LoadDKShare(addr)
	if err != nil {
		return nil, err
	}
	for i := range cmtDKShare.NodePubKeys {
		if *cmtDKShare.NodePubKeys[i] == *selfPubKey {
			return cmtDKShare, nil
		}
	}
	return nil, xerrors.Errorf("createCommitteeIfNeeded: I am not among nodes of the committee record. Inconsistency")
}
```

packages/chain/statemgr/domain_with_fallback.go
```diff
@@ -0,0 +1,135 @@
...
+	return df.fallback
+}
+
+func (df *DomainWithFallback) updateDomainPeers() {
+	var peers []*ed25519.PublicKey
+	if df.fallback {
+		selfPubKey := df.net.Self().PubKey()
+		allTrusted := make([]*ed25519.PublicKey, 0)
+		for _, n := range df.net.PeerStatus() {
+			if n.IsAlive() && *n.PubKey() != *selfPubKey {
```
jorgemmsilva, line 104: why is there the possibility that `selfPubKey` is included in `df.net.PeerStatus`? shouldn't peers be only "other nodes"? :smile: 

I guess it could be the case when the same key owns more than 1 node in a committee (?)

packages/peering/domain/domain.go
```diff
@@ -73,50 +61,71 @@ func (d *DomainImpl) SendMsgByNetID(netID string, msgReceiver, msgType byte, msg
...
+	d.mutex.RUnlock()
+	changed := false
+	//
+	// Add new peers.
+	for _, newPeerPubKey := range newPeerPubKeys {
+		found := false
+		for _, existingPeer := range nodes {
+			if *existingPeer.PubKey() == *newPeerPubKey {
+				found = true
+				break
```
jorgemmsilva: found can be removed, we just `break` every time found would true

packages/peering/domain/domain.go
```diff
@@ -73,50 +61,71 @@ func (d *DomainImpl) SendMsgByNetID(netID string, msgReceiver, msgType byte, msg
...
+			}
+		}
 	}
-	peer, err := d.netProvider.PeerByNetID(netID)
-	if err != nil {
-		return err
+	//
+	// Remove peers that are not needed anymore.
+	for _, oldPeer := range nodes {
+		found := false
```
jorgemmsilva: same as `found` above, could be removed

packages/peering/domain/domain.go
```diff
@@ -73,50 +61,71 @@ func (d *DomainImpl) SendMsgByNetID(netID string, msgReceiver, msgType byte, msg
...
 	}
-	if netID == d.netProvider.Self().NetID() {
-		return nil
+	d.mutex.RUnlock()
+	changed := false
+	//
+	// Add new peers.
+	for _, newPeerPubKey := range newPeerPubKeys {
+		found := false
+		for _, existingPeer := range nodes {
```
jorgemmsilva: `nodes` is a map, couldn't we check directly if it exists? why do we need to iterate it?

packages/peering/domain/domain.go
```diff
@@ -73,50 +61,71 @@ func (d *DomainImpl) SendMsgByNetID(netID string, msgReceiver, msgType byte, msg
...
+				d.log.Infof("Domain peer added, pubKey=%v, netID=%v", newPeerSender.PubKey().String(), newPeerSender.NetID())
+			}
+		}
 	}
-	peer, err := d.netProvider.PeerByNetID(netID)
-	if err != nil {
-		return err
+	//
+	// Remove peers that are not needed anymore.
+	for _, oldPeer := range nodes {
```
jorgemmsilva: this removal logic is a bit convoluted, I think we could probably simplify by starting with an empty list of nodes, and just add the new "valid" peers to it :thinking:

jorgemmsilva (approved):
big one :smile: lgtm overall, just some minor comments

packages/chain/statemgr/domain_with_fallback.go
kape1395: I had no intention to implement PeerDomainProvider.
Some functions are missing for that.
Although it is easy to do, those functions are of no use here, and the PeerDomainProvider is not enough in this context.

packages/chain/chainimpl/chainimpl.go
kape1395: Was thinking about that. Decided to keep it similar to the existing way (the peering domain was created by chainImpl as well). And I was trying to keep the approach of not passing peeringIDs directly, because you have done this in this way previously.

packages/chain/statemgr/domain_with_fallback.go
kape1395: The current node is considered a peer as well. For uniformity.
The same interface is then implemented by the NetProvider instead of dedicated Peer object.

packages/chain/chain.go
kape1395: fixed.

packages/chain/chainimpl/eventproc.go
kape1395: Fixed.

packages/peering/domain/domain.go
kape1395: I was trying to avoid breaking out of the outer loop from the inner.
That `found` would be still needed to log the removal of the node.
Tried to reorganize the code, hope new version looks better.

packages/peering/domain/domain.go
kape1395: Tried to reorganize the code, hope new version looks better.

packages/peering/domain/domain.go
kape1395: Good catch.

packages/peering/domain/domain.go
kape1395: done.

packages/registry/registry_impl.go
```diff
@@ -156,6 +133,9 @@ func (r *Impl) SaveDKShare(dkShare *tcrypto.DKShare) error {
 func (r *Impl) LoadDKShare(sharedAddress ledgerstate.Address) (*tcrypto.DKShare, error) {
 	data, err := r.store.Get(dbKeyForDKShare(sharedAddress))
 	if err != nil {
+		if err.Error() == "key not found" {
```
lunfardo314: better `errors.Is(err, kvstore.ErrKeyNotFound`)

lunfardo314 (approved):

packages/apilib/deploychain.go
kape1395: Maybe will consider that after the merge. To avoid making even more conflicts.

packages/registry/registry_impl.go
kape1395: Done.

PR #722, opened 2022-01-09
multiple renames, improvements and tests

lunfardo314:


packages/iscp/assets.go
```diff
@@ -175,6 +197,47 @@ func (a *Assets) Equals(b *Assets) bool {
...
 
+// FitsTheBudget checks if:
+// - 'a' has all non-negative values. Negative values is error
+// - 'a' all values are <= of corresponding values of 'b'. It means, we can debit 'a' from 'b' without overrun of funds
+func (a *Assets) FitsTheBudget(budget *Assets) (bool, error) {
+	if a == budget {
+		return true, nil
+	}
+	if a.IsEmpty() {
+		return true, nil
```
jorgemmsilva, line 208:
```suggestion
		return false, nil
```

should be false, right?

jorgemmsilva (approved):
lgtm, except 1 comment

PR #730, opened 2022-01-12
Allowance budget mutated

lunfardo314:
Now `Allowance` budget in the `Sandbox` is mutated mutated with each call to `sandbox.TransferAllowedFunds`. 
It is needed to keep track of the allowed budget and prevent overrun by repeating calls to `TransferAllowedFunds`


packages/iscp/assets.go
```diff
@@ -200,6 +200,7 @@ func (a *Assets) Equals(b *Assets) bool {
 // FitsTheBudget checks if:
 // - 'a' has all non-negative values. Negative values is error
 // - 'a' all values are <= of corresponding values of 'b'. It means, we can debit 'a' from 'b' without overrun of funds
+// Deprecated: not needed anymore
```
jorgemmsilva, line 203: can we remove the deprecated methods?

packages/vm/core/accounts/impl.go
```diff
@@ -75,24 +75,29 @@ func withdraw(ctx iscp.Sandbox) (dict.Dict, error) {
...
 	if ctx.Caller().Address().Equal(ctx.ChainID().AsAddress()) {
 		// if the caller is on the same chain, do nothing
 		return nil, nil
 	}
 	// move all allowed funds to the account of the current contract context
+	// before saving the allowance budget because after the transfer it is mutated
+	fundsToWithdraw := ctx.AllowanceAvailable().Clone()
 	ctx.TransferAllowedFunds(ctx.AccountID())
 
+	// por las dudas
```
jorgemmsilva, line 89: :sweat_smile: ?

jorgemmsilva (approved):

packages/vm/vmcontext/runreq.go
```diff
@@ -194,7 +194,7 @@ func (vmctx *VMContext) calculateAffordableGasBudget() {
 	if vmctx.req.SenderAddress() == nil {
 		panic("inconsistency: vmctx.req.SenderAddress() == nil")
 	}
-	// calculate how many tokens for gas fee can be guaranteed after taking into account the allowance
+	// calculate how many tokens for gas fee can be guaranteed after taking into account the allowanceRemains
```
dessaya, line 197: unintended rename?

packages/vm/vmcontext/vmcontext.go
```diff
@@ -69,10 +69,10 @@ type VMContext struct {
...
 
 type callContext struct {
-	caller    *iscp.AgentID // calling agent
-	contract  iscp.Hname    // called contract
-	params    dict.Dict     // params passed
-	allowance *iscp.Assets  // allowance passed
+	caller           *iscp.AgentID // calling agent
+	contract         iscp.Hname    // called contract
+	params           dict.Dict     // params passed
+	allowanceRemains *iscp.Assets  // MUTABLE: allowance budget left after TransferAllowedFunds
```
dessaya, line 75: `allowanceAvailable` better? To make it consistent with the `AllowanceAvailable()` method

packages/vm/vmcontext/call.go
```diff
@@ -67,10 +67,10 @@ const traceStack = false
...
 func (vmctx *VMContext) pushCallContext(contract iscp.Hname, params dict.Dict, allowance *iscp.Assets) {
 	ctx := &callContext{
-		caller:    vmctx.getToBeCaller(),
-		contract:  contract,
-		params:    params.Clone(),
-		allowance: allowance,
+		caller:           vmctx.getToBeCaller(),
+		contract:         contract,
+		params:           params.Clone(),
+		allowanceRemains: allowance.Clone(), // we have to clone it because it will be mutated by TransferAllowedFunds
```
dessaya, line 73: How is a call between contracts handled?

I have an allowance A = 10
I call contract C1. The allowance is cloned, so I have A1 = 10
Contract C1 calls contract C2. The allowance is cloned, so I have A2 = 10
Contract C2 spends 5, so A2 = 5.
When the call context is popped, the current allowance for contract C1 is still A1 = 10?

packages/iscp/assets.go
lunfardo314: I will, I promise

packages/vm/vmcontext/call.go
lunfardo314: yes. Allowance is what C1 (caller) allows for C2 (target) to take from C1's account. Allowance is not passed, it is intended for the target. If target calls another target, it creates another, unrelated, allowance wrt its own account

PR #732, opened 2022-01-12
Add solo.EstimateGas()

dessaya:


packages/vm/vmtask.go
```diff
@@ -34,8 +35,13 @@ type VMTask struct {
 	Entropy                  hashing.HashValue
 	ValidatorFeeTarget       *iscp.AgentID
 	Log                      *logger.Logger
-	OnFinish                 func(callResult dict.Dict, callError error, vmError error)
```
lunfardo314, line 37: why deleted `OnFinish` callback provided by the caller? For Solo it may not be very useful but it is very important for consensus, like here:  https://github.com/iotaledger/wasp/blob/8cd6ecdef31d66c974fb8f9ff3892b10fc135d83/packages/chain/consensus/action.go#L237  We definitely need it. It can be without parameters and default `nil`

lunfardo314 (changes requested):

packages/vm/runvm/runtask.go
```diff
@@ -3,72 +3,63 @@ package runvm
...
+	"github.com/iotaledger/wasp/packages/util"
 	"github.com/iotaledger/wasp/packages/vm"
 	"github.com/iotaledger/wasp/packages/vm/vmcontext"
 )
 
 type VMRunner struct{}
 
-func (r VMRunner) Run(task *vm.VMTask) {
+func (r VMRunner) Run(task *vm.VMTask) (results []*vm.RequestResult, err error) {
 	// optimistic read panic catcher for the whole VM task
```
lunfardo314: the consensus runs the task in goroutine and receives all results asynchronously via calling calback and passing a message with the same `VMTask` to the consensus. All results (state and transaction) are contained in the `VMTask`. Not to complicate things, we better put `[]*vm.RequestResult` into the VMTask structure too, next to the state and transaction essence, not return as a result of the function call.

lunfardo314 (changes requested):

packages/solo/req.go
```diff
@@ -191,12 +181,37 @@ func (ch *Chain) RequestFromParamsToLedger(req *CallParams, keyPair *cryptolib.K
 			},
 			Options: nil,
 		}},
-		RentStructure:                ch.Env.utxoDB.RentStructure(),
-		DisableAutoAdjustDustDeposit: ch.Env.disableAutoAdjustDustDeposit,
```
lunfardo314, line 195: why this deleted?

lunfardo314 (approved):

packages/solo/req.go
dessaya: Sorry, it is a mistake. I'll bring it back in another PR

PR #733, opened 2022-01-13
optional adjustment to dust requirements

lunfardo314:


packages/transaction/makeoutput.go
```diff
@@ -28,20 +29,23 @@ func ExtendedOutputFromPostData(
...
+// If auto adjustment to dust is disabled and not enough iotas, returns an error
 func MakeExtendedOutput(
 	targetAddress iotago.Address,
 	senderAddress iotago.Address,
 	assets *iscp.Assets,
 	metadata *iscp.RequestMetadata,
 	options *iscp.SendOptions,
 	rentStructure *iotago.RentStructure,
-) *iotago.ExtendedOutput {
+	notAutoAdjustToDustRequirement ...bool,
```
jorgemmsilva: nit: feels off to have "negative meaning" boolean values `notAutoAdjustToDustRequirement`

I would normally go for `shouldAutoAdjustDustRequirement` which defaults to true when not specified

jorgemmsilva (approved):

packages/transaction/makeoutput.go
lunfardo314: me too. But (a) currently auto adjust is default, so better to stay it this way and (b) `false` is a default value for booleans . So with `false` and absense it should go for default. It means providing `true` we override default and make NOT to autoadjust

lunfardo314: I will rename it to `supressAutoAdjustment`

packages/transaction/makeoutput.go
```diff
@@ -28,20 +29,23 @@ func ExtendedOutputFromPostData(
...
+// If auto adjustment to dust is disabled and not enough iotas, returns an error
 func MakeExtendedOutput(
 	targetAddress iotago.Address,
 	senderAddress iotago.Address,
 	assets *iscp.Assets,
 	metadata *iscp.RequestMetadata,
 	options *iscp.SendOptions,
 	rentStructure *iotago.RentStructure,
-) *iotago.ExtendedOutput {
+	suppressAutoAdjustToDustRequirement ...bool,
```
dessaya: `disable` instead of `suppress`?

packages/transaction/requesttx.go
```diff
@@ -9,11 +9,12 @@ import (
...
-	UnspentOutputs   []iotago.Output
-	UnspentOutputIDs []*iotago.UTXOInput
-	Requests         []*iscp.RequestParameters
-	RentStructure    *iotago.RentStructure
+	SenderKeyPair                cryptolib.KeyPair
+	UnspentOutputs               []iotago.Output
+	UnspentOutputIDs             []*iotago.UTXOInput
+	Requests                     []*iscp.RequestParameters
+	RentStructure                *iotago.RentStructure
+	SuspendAutoAdjustDustDeposit bool // if true, the minimal dust deposit won't be adjusted automatically
```
dessaya, line 17: `Disable` instead of `Suspend`?

dessaya (approved):

PR #745, opened 2022-01-16
multiple improvements and tests in gas and fee logic

lunfardo314:
- gas fee charging logic rewritten in the core VM
- improvements in Solo: gas estimate, last receipt etc
- additional tests  with `sandbox.Send`
-  additional (less safe) option to move funds in the sandbox, without checking if target exists `TransferAllowedFundsForceCreateTarget`
- etc


packages/solo/solo.go
```diff
@@ -93,18 +95,16 @@ type Chain struct {
...
 	// global processor cache
 	proc *processors.Cache
-
 	// related to asynchronous backlog processing
 	runVMMutex sync.Mutex
-
 	// mempool of the chain is used in Solo to mimic a real node
 	mempool mempool.Mempool
+	// receipt of the last call
+	lastReceipt *blocklog.RequestReceipt
```
dessaya, line 107: Is this really necessary? The `Chain` struct is already bloated, it would be nice if we can avoid adding more state. Also it presents some ambiguities: e.g. is it updated when you call `EstimateGas()`? Should it?
Better to have the receipt be returned by each function IMO.

dessaya (approved):

packages/solo/solo.go
lunfardo314: I think it is necessary for the UX of the Solo user. Otherwise tests are bloated. `ch.LastReceipt()` call It is just a receipt of the last request. Very clear semantics. In general not nice pattern but very useful in a testing environment.
It allowed to delete several Solo functions which existed  only with the purpose to return receipt.
`EstimateGas` should not store it imo. But it also could, no big harm for the logic of the test.

jorgemmsilva: we have `PostRequestSyncReceipt` to receive the receipt in the same call we do the request, right?
couldn't we just use that in solo?

packages/solo/chain.go
```diff
@@ -120,26 +120,32 @@ func (ch *Chain) UploadBlob(keyPair *cryptolib.KeyPair, params ...interface{}) (
 		return expectedHash, nil
 	}
 
-	gasFeePolicy := ch.GetGasFeePolicy()
-	require.Nil(ch.Env.T, gasFeePolicy.GasFeeTokenID)
-	gasEstimate := blob.GasForBlob(blobAsADict) + gasFeePolicy.GasNominalUnit*10
+	userAddr := cryptolib.Ed25519AddressFromPubKey(user.PublicKey)
+	// We have to have some iotas even in estimate mode, otherwise we cannot create transaction
+	const minimumIotasForEstimate = 100_000
```
jorgemmsilva, line 125: why do we need a "minimum of iotas on L1" to estimate?

packages/vm/core/testcore_stardust/sbtests/sbtestsc/impl_send.go
```diff
@@ -0,0 +1,48 @@
...
+		)
+	}
+	return nil, nil
+}
+
+func pingAllowanceBack(ctx iscp.Sandbox) (dict.Dict, error) {
+	// assert caller is L1 address, not a SC
+	ctx.Requiref(!ctx.Caller().Address().Equal(ctx.ChainID().AsAddress()) && ctx.Caller().Hname() == 0,
+		"pingAllowanceBack: caller expected to be a L1 address")
+	// save allowance budget because after transfer it will be modified
```
jorgemmsilva, line 29: > // save allowance budget because after transfer it will be modified

this seems like a confusing design, I feel that we should make `ctx.AllowanceAvailable()` return a cloned object by default

packages/kv/kv_with_gas.go
```diff
@@ -1,93 +1,94 @@
+// Deprecate ???
```
jorgemmsilva, line 1: we should remove this if not in use.

but why do we decided to not use the "kv with gas approach"? 

also: do we not charge for reading the state anymore?

jorgemmsilva (approved):

packages/solo/solo.go
lunfardo314: I deleted `PostRequestSyncReceipt` and ``PostRequestOffLedgerSyncReceipt`` we do not need it anymore. the `LastReceipt`  is more flexible, that is what I found

packages/solo/chain.go
lunfardo314: because a) the mocked logic of `Estimate` checks the L2 account too. It needs to have enough tokens because otherwise it will fail with `not enough gas`  and the gas estimate will be wrong. So, `estimate` creates a request with assets on it to overcome the situation. It is a real transaction with real outputs, so it fails to create one if there are not enough tokens.
It is complicated, but that is the way

packages/vm/core/testcore_stardust/sbtests/sbtestsc/impl_send.go
lunfardo314: returning cloned won't change that. Simply, after `transfer`, the result of `Allowance` changes. We already discussed that

packages/kv/kv_with_gas.go
lunfardo314: we charge here, one line of code https://github.com/iotaledger/wasp/blob/8d6c462a70dabaa5220d8fbf672a647a825d1f69/packages/vm/vmcontext/stateaccess.go#L134
It is the only place, we do not need more.

packages/vm/core/testcore_stardust/sbtests/sbtestsc/impl_send.go
jorgemmsilva: yes but the return value wouldn't change if we clone.

what I mean:
```
a := ctx.AllowanceAvailable()
... // do something that changes the allowance
b := ctx.AllowanceAvailable()

// here a == b, and that feels wrong
```

packages/kv/kv_with_gas.go
jorgemmsilva: for storage only, not for reading the state. What I'm saying is that we should also charge for reading the state (like `SLOAD` opcode for evm)

packages/solo/chain.go
jorgemmsilva: we could potentially clone utxodb, credit enough iotas to the user, do the estimation, then rollback to the previous utxodb. wdyt?

lunfardo314: L1 and L2 ledgers are changing. So estimate is approximate anyway. The Estimate function is exactly the same which will be used in the Wasp's web endpoint. Don't understand how we could snapshot the ledger state

packages/vm/core/testcore_stardust/sbtests/sbtestsc/impl_send.go
lunfardo314: ah, I know what you mean. Yes, we definitely could clone automatically. But then we would clone every time we check it, so probably better force the user to do it explicitely, but only when he/she needs to safe it? Not sure, actually. Safer would be to clone automatically, a bit a waste

packages/kv/kv_with_gas.go
lunfardo314: maybe but I doubt we need it, we are doing it approximately only because sandbox is not a processor instructions. That would be another line of code anyway

packages/solo/chain.go
jorgemmsilva: in solo it should be doable. but maybe overkill - just an idea

packages/vm/core/testcore_stardust/sbtests/sbtestsc/impl_send.go
jorgemmsilva: I'm in favor of cloning it shouldn't be too much of an overhead and it feels good to not change values without the user expecting it. 
It would be nice to have other opinions on this, @dessaya wdyt?

dessaya: +1 for returning a clone, or even making it immutable

PR #748, opened 2022-01-17
fixed 2 bugs!

lunfardo314:


packages/vm/core/accounts/internal.go
```diff
@@ -194,7 +194,7 @@ func DebitFromAccount(state kv.KVStore, agentID *iscp.AgentID, assets *iscp.Asse
 	defer checkLedger(state, "DebitFromAccount OUT")
 
 	if !debitFromAccount(account, assets) {
-		panic(xerrors.Errorf(" agentID: %s. %w", agentID.String(), ErrNotEnoughFunds))
+		panic(xerrors.Errorf(" agentID: %s. %v", agentID.String(), ErrNotEnoughFunds))
```
dessaya, line 197: Is this change intentional? `%w` wraps the given error inside ther newly created one.

dessaya (approved):

packages/vm/core/accounts/internal.go
lunfardo314: yes, it is intentional. I explained once: the wrapped panic messages are too scary for any Solo user. So I removed all %w in the code. We may want to revisit the UX of the error messages in more systematic way

jorgemmsilva: yes, but this means we cannot type check the wrapped error (with `errors.Is`)

jorgemmsilva (approved):

PR #749, opened 2022-01-17
add EstimateGasOffLedger

dessaya:


packages/solo/req.go
```diff
@@ -298,28 +296,43 @@ func (ch *Chain) PostRequestSyncExt(req *CallParams, keyPair *cryptolib.KeyPair)
...
+// any changes in the ledger. It returns the amount of gas consumed.
+func (ch *Chain) EstimateGasOnLedger(req *CallParams, keyPair *cryptolib.KeyPair) (gas uint64, gasFee uint64, err error) {
+	res, err := ch.SimulateRequestOnLedger(req, keyPair)
 	if err != nil {
 		return 0, 0, err
 	}
-	ch.lastReceipt = res.Receipt
-	gasFeePolicy := ch.GetGasFeePolicy()
-	f1, f2 := gasFeePolicy.FeeFromGas(res.Receipt.GasBurned)
-	feeEstimate := f1 + f2 + gasFeePolicy.GasPricePerNominalUnit
```
dessaya, line 321: @lunfardo314 If I understand correctly, this computes the same value as `receipt.GasFeeCharged` plus `GasPricePerNominalUnit`. I hanged it back to simply return the value of `receipt.GasFeeCharged` and tests still pass. Was there a reason to calculate the value by hand and also add the nominal unit instead of just returning the value from the receipt? In that case I can revert it, but I don't see it.

jorgemmsilva (approved):

PR #752, opened 2022-01-19
refactor: rename solo asset getters

jorgemmsilva:


lmoe (approved):
Now that I read through all the tests, I think the naming `L1NativeTokens` implies to return a list of all tokens instead of the balance of one specific token. Sure you add the tokenId as a second parameter which clears that up, but if you simply use intelli sense to go through all methods to find a 'getBalance' function, it might be a bit confusing. 

Except that, it looks great.

jorgemmsilva:
yeah, I considered that, but couldn't find better wording (`L1NativeToken` didn't feel right)- its obvious what does once you see the function signature. There is `L1Assets` which returns everything owned by a given address

lmoe:
Yeah I see. The only option I see is to revert back to L1NativeTokenBalance, but this breaks consistency with the other functions. Forget what I said. :)

dessaya (approved):

PR #753, opened 2022-01-19
Metrics: Chain now has current_state_index metric.

kape1395:


dessaya (approved):

ivange94 (approved):

packages/metrics/chain.go
```diff
@@ -83,6 +89,14 @@ func (c *chainMetricsObj) RecordBlockSize(blockIndex uint32, blockSize float64)
 	c.metrics.blockSizes.With(prometheus.Labels{"chain": c.chainID.String(), "block_index": fmt.Sprintf("%d", blockIndex)}).Set(blockSize)
 }
 
+func (c *chainMetricsObj) LastSeenStateIndex(stateIndex uint32) {
```
ivange94, line 92: Without looking at the code, `LastSeenStateIndex` reads a lot like `CurrentStateIndex`. So from the prometheus interface where we only look at names and values, these two might be confusing. How is this different from CurrentStateIndex? And I feel the naming doesn't cover the logic in the method. This ignores a state index values that are lower than the last seen value. Why?

ivange94: Never mind. I get it.

ivange94 (approved):

packages/metrics/chain.go
kape1395: `CurrentStateIndex` is the index for which next batch is being collected and agreed.
If a node gets stuck, it is possible that it will receive newer `LastSeenStateIndex` from the network while still working on the `CurrentStateIndex`.

PR #755, opened 2022-01-20
Improvements to gas estimation

dessaya:
* It is now possible to call `EstimateGas` without having to provide the gas fee in advance.
* If the gas budget is set to 0, the `PostRequest*` functions in Solo now automatically estimate gas and check that the sender can afford the fee.
* Removed the "unverified account" check for off-ledger requests; not necessary with the gas mechanism

### DIFF
--- a/packages/solo/req.go
+++ b/packages/solo/req.go
@@ -19,6 +19,7 @@ import (
 	"github.com/iotaledger/wasp/packages/vm/core/blocklog"
 	"github.com/iotaledger/wasp/packages/vm/viewcontext"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/xerrors"
 )
 
 type CallParams struct {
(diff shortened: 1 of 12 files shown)


packages/solo/req.go
```diff
@@ -261,11 +263,28 @@ func (ch *Chain) PostRequestSync(req *CallParams, keyPair *cryptolib.KeyPair) (d
...
+		if err != nil {
+			return nil, xerrors.Errorf("cannot estimate gas: %w", err)
+		}
+		if fee > 0 {
+			err := ch.checkCanAffordFee(fee, req, keyPair)
+			if err != nil {
+				return nil, err
+			}
+		}
+		req.WithGasBudget(gas)
```
jorgemmsilva, line 277: don't we also need to ensure that there are enough iotas in `assets - allowance` to cover for the gas budget?

dessaya: checkCanAffordFee does that, or is it something I missed?

packages/solo/req.go
```diff
@@ -283,9 +302,60 @@ func (ch *Chain) LastReceipt() *blocklog.RequestReceipt {
...
+	policy := ch.GetGasFeePolicy()
+	available := uint64(0)
+	if policy.GasFeeTokenID == nil {
+		available = ch.L2Iotas(agentID)
+		if req.assets != nil {
+			available += req.assets.Iotas
+		}
+		if req.allowance != nil {
+			available -= req.allowance.Iotas
+		}
```
jorgemmsilva, line 319: this is tricky, I think we don't take from allowance to pay for gas

packages/solo/req.go
jorgemmsilva: you're absolutely right, forget it :+1:

jorgemmsilva (approved):

PR #756, opened 2022-01-20
impr: simplify gas details, change some costs, enable db compression

jorgemmsilva:


packages/solo/ledgerl1l2.go
```diff
@@ -156,11 +156,11 @@ type foundryParams struct {
 
 // CreateFoundryGasBudgetIotas always takes 1000 iotas as gas budget and assets for the call
```
dessaya:
```suggestion
// CreateFoundryGasBudgetIotas always takes 100000 iotas as gas budget and assets for the call
```

packages/vm/gas/policy.go
```diff
@@ -67,14 +66,13 @@ func (p *GasFeePolicy) AffordableGasBudgetFromAvailableTokens(availableTokens ui
...
 	return nominalUnitsOfGas * p.GasNominalUnit
-
 }
 
 func DefaultGasFeePolicy() *GasFeePolicy {
 	return &GasFeePolicy{
 		GasFeeTokenID:          nil, // default is iotas
-		GasNominalUnit:         100, // gas is burned in 100-s and not less than 100
-		GasPricePerNominalUnit: 100, // default is 1 gas == 1 iota
+		GasNominalUnit:         100, // gas is burned in 1000-s and not less than 1000
```
dessaya: the value of the constant and the comment are not consistent?

dessaya (approved):

packages/vm/gas/policy.go
lunfardo314: comment is not consistent with the value

jorgemmsilva: yes, thanks for pointing that out. updated

PR #759, opened 2022-01-24
Fix goreleaser not found in release pipeline

ernoaapa:
The release GitHub actions release pipeline was failing to error
"goreleaser: command not found".
The root cause was that the goreleaser tool was not installed,
because at some stage the "goreleaser/goreleaser-action" were
dropped off. Now back and the release should work again.


jorgemmsilva:
Thanks @ernoaapa !

PR #761, opened 2022-01-24
impr: (VM) gas 

jorgemmsilva:
- min gas per request
- deprecated `GasNominalUnit` and `GasPricePerNominalUnit`, gas price now set by: `GasPerToken`
- improved gas estimation
- added test for deposits/withdrawals edge cases

### DIFF
--- a/packages/vm/gas/policy.go
+++ b/packages/vm/gas/policy.go
@@ -26,11 +26,6 @@ func calcFee(gasUnits, gasPerToken uint64) uint64 {
 // FeeFromGas return ownerFee and validatorFee
 func (p *GasFeePolicy) FeeFromGas(gasUnits, availableTokens uint64) (sendToOwner, sendToValidator uint64) {
 	var fee uint64
-	// ensure at least the minimum amount of gas is charged
-	minimumGas := BurnCodeMinimumGasPerRequest.Cost()
-	if gasUnits < minimumGas {
-		gasUnits = minimumGas
-	}
 
 	// round up
 	fee = calcFee(gasUnits, p.GasPerToken)
@@ -50,7 +45,7 @@ func (p *GasFeePolicy) FeeFromGas(gasUnits, availableTokens uint64) (sendToOwner
 }
 
 func (p *GasFeePolicy) IsEnoughForMinimumFee(availableTokens uint64) bool {
-	minFee := calcFee(BurnCodeMinimumGasPerRequest.Cost(), p.GasPerToken)
+	minFee := calcFee(BurnCodeMinimumGasPerRequest1P.Cost(), p.GasPerToken)
 	return availableTokens >= minFee
 }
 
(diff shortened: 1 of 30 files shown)


packages/vm/gas/table.go
```diff
@@ -59,6 +60,7 @@ var burnTable = BurnTable{
 	BurnCodeUtilsBLSValidSignature:     {"bls valid", constValue(2000)},
 	BurnCodeUtilsBLSAddrFromPubKey:     {"bls addr", constValue(50)},
 	BurnCodeUtilsBLSAggregateBLS1P:     {"bls aggregate", linear(CoefBLSAggregate)},
+	BurnCodeMinimumGasPerRequest:       {"minimum gas per request", constValue(100)}, // TODO maby make it configurable (gov contract?)
```
dessaya:
```suggestion
	BurnCodeMinimumGasPerRequest:       {"minimum gas per request", constValue(100)}, // TODO maybe make it configurable (gov contract?)
```

packages/vm/gas/policy.go
```diff
@@ -11,69 +11,58 @@ import (
...
 }
 
 func DefaultGasFeePolicy() *GasFeePolicy {
 	return &GasFeePolicy{
-		GasFeeTokenID:          nil, // default is iotas
-		GasNominalUnit:         100, // gas is burned in 100-s and not less than 100
-		GasPricePerNominalUnit: 1,   // default is 1 iota = 100 gas
-		ValidatorFeeShare:      0,   // by default all goes to the governor
+		GasFeeTokenID:     nil, // default is iotas
+		GasPerToken:       100, // gas is burned in 100-s and not less than 100
```
dessaya: `gas is burned in 100-s and not less than 100`
Is this still true?

dessaya (approved):

packages/vm/gas/policy.go
jorgemmsilva: I mean with `GasPerToken = 100` you would need 2 tokens to burn `101` gas, so each tokens would entitle you to 100 gas units. But the wording is kinda confusing, will edit 👍

packages/vm/gas/policy.go
```diff
@@ -11,68 +11,48 @@ import (
 type GasFeePolicy struct {
 	// GasFeeTokenID contains iotago.NativeTokenID used to pay for gas, or nil if iotas are used for gas fee
 	GasFeeTokenID *iotago.NativeTokenID
-	// GasNominalUnit price is specified per how many gas units
-	GasNominalUnit uint64
-	// GasPricePer1000Units how many gas tokens you pay for 1000 gas
-	GasPricePerNominalUnit uint64
+	// GasPerToken specifies how many gas units are paid for each token ( 100 means 1 tokens pays for 100 gas)
+	GasPerToken uint64
```
lunfardo314, line 15: we are losing option to specify for example `1 token per 1000 gas`? That may be a big limiting factor if we want gas fee to be paid for example by ASM token which has big market value. 
The point of nominals is about:
- being able to choose a token for paying a fee
- being independent from the price of the token

lunfardo314 (approved):

packages/vm/gas/policy.go
jorgemmsilva: You're absolutely correct, but as we talked about in the team call - let's keep it simple for now (accept the current limitation that its not possible to increase the price to more that 1 token = 1 gas unit) and we shall implement a more sophisticated gas pricing when needed :+1:

PR #762, opened 2022-01-24
Improve goreleaser pipeline

ernoaapa:
While testing out the release pipeline in a personal fork, found out that the `goreleaser` step didn't work as expected.
Fixed the wrong golang version and also disabled windows and darwin(mac) build because the depending library (grocksdb) does not support that (yet).


jorgemmsilva (approved):

jorgemmsilva:
Awesome, thanks once again for your contribution @ernoaapa !

PR #774, opened 2022-02-02
chore: Support Apple M1

howjmay:
On M1 Apple Silicon, replacing wasmtime-go with a local libray help
us support the macOS arm64 machines.

### DIFF
--- a/documentation/docs/guide/chains_and_nodes/running-a-node.md
+++ b/documentation/docs/guide/chains_and_nodes/running-a-node.md
@@ -64,6 +64,16 @@ You can build and install both `wasp` and `wasp-cli` by running:
 make install
 ```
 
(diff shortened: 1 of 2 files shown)


howjmay:
If this is an acceptable solution, I will modify the command in Makefile, too. I will add an extra install tag `install-macos-arm64`, and similar tag will be added to other tags

jorgemmsilva (approved):
cool :rocket: 

I guess the only downside is that we need to keep the 2 go.mod files in sync, but if it works it works :man_shrugging:

jorgemmsilva (approved):

dessaya (approved):

tuddman:
@HowJMay  Were you able to get it working successfully on an M1?  If so can you post the steps you needed to take?

PR #781, opened 2022-02-02
feat(mempool): timelocked + expired requests

dessaya:


packages/chain/mempool/mempool.go
```diff
@@ -222,32 +239,33 @@ func (m *mempool) traceOut(reqid iscp.RequestID) {
...
+		return false
 	}
-	windowFrom := currentTime.Add(-FallbackDeadlineMinAllowedInterval)
-	windowTo := currentTime.Add(FallbackDeadlineMinAllowedInterval)
+	return !expiry.Time.IsZero() && currentTime.Time.After(expiry.Time.Add(-FallbackDeadlineMinAllowedInterval))
+}
 
-	if expiry.Time.After(windowFrom) && expiry.Time.Before(windowTo) {
+func (m *mempool) isUnlockable(ref *requestRef, currentTime iscp.TimeData) bool {
+	if isExpired(ref, currentTime) {
```
jorgemmsilva, line 255: I feel that this check should be part of iotago, so that we are 100% sure its the same logic.

packages/chain/mempool/mempool.go
```diff
@@ -222,32 +239,33 @@ func (m *mempool) traceOut(reqid iscp.RequestID) {
...
 		return false
 	}
 
 	output, _ := ref.req.AsOnLedger().Output().(iotago.TransIndepIdentOutput)
```
jorgemmsilva, line 259: I know this is not new, but what is a `TransIndepIdentOutput` exactly? :sweat_smile: 
are all requests guaranteed to be of that type?

jorgemmsilva (approved):
LGTM, just have the feeling that expiry logic should be reused on different projects, so that we don't risk having slightly different checks

packages/chain/mempool/mempool.go
dessaya: Yes, although we use a slightly different logic because of the 1 minute window. (Not sure why we need that, but I tried to keep that as is).

packages/chain/mempool/mempool.go
dessaya: Haha no idea :man_shrugging:

PR #783, opened 2022-02-04
Newwasmvm

BugFreeSoftware:
# Description of change

Complete refactoring of Wasm VM in preparation of Stardust integration.
Removed all state from the host side and simplified the design.
Kept it mostly backward compatible.

## Type of change

All smart contracts will have to be regenerated from schema tool and rebuilt.
Old Wasm code will no longer work.

Tested all demo smart contracts in all 3 languages. They pass all tests

## Change checklist


packages/wasmvm/wasmvmhost/host.go
```diff
@@ -0,0 +1,138 @@
...
+func (w *WasmVMHost) StateExists(key []byte) bool {
+	// value size -1 means only test for existence
+	// returned size -1 indicates keyID not found (or error)
+	// this removes the need for a separate hostStateExists function
+	return hostStateGet(&key[0], int32(len(key)), nil, -1) >= 0
+}
+
+func (w *WasmVMHost) StateGet(key []byte) []byte {
+	//TODO optimize when type size is known in advance?
+	// or maybe pass in a larger buffer that will fit most rsults?
```
howjmay: typo
```suggestion
	// or maybe pass in a larger buffer that will fit most results?
```

BugFreeSoftware: Fixed

packages/wasmvm/wasmvmhost/host.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// go:build wasm
```
howjmay:
```suggestion
//go:build wasm
```

packages/wasmvm/wasmvmhost/host.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2020 IOTA Stiftung
```
howjmay, line 1: BTW should it be 2022 here?

packages/wasmvm/wasmvmhost/host.go
BugFreeSoftware: Fixed

packages/wasmvm/wasmvmhost/host.go
BugFreeSoftware: Probably should be 2020-2022, but that means we touch all source files. 
We better coordinate that and do it once for the entire code base. 
Also need to add missing copyright headers to a shitload of files.

packages/wasmvm/wasmlib/go/wasmlib/exports.go
```diff
@@ -0,0 +1,43 @@
...
+func ScExportsExport(exportMap *ScExportMap) {
+	ExportName(-1, "WASM::GO")
+
+	for i := range exportMap.Funcs {
+		ExportName(int32(i), exportMap.Names[i])
+	}
+
+	offset := len(exportMap.Funcs)
+	for i := range exportMap.Views {
+		ExportName(int32(i)|0x8000, exportMap.Names[offset+i])
```
howjmay, line 31: for the const `0x8000` should we declare a const var for it and explain the constraint? I am curious why do we set it as `0x8000`

BugFreeSoftware: It's the flag that determines whether this is a func or a view. I think it's a good idea to give this flag a proper name. Since the rest of the function id is an index, I just took a reasonably high bit. This gives us the ability to have 16K funcs and 16K views.

jorgemmsilva (approved):

contracts/wasm/core_build.cmd
```diff
@@ -1,8 +1,9 @@
 @echo off
-cd ..\..\packages\vm\wasmlib
+go install ../../tools/schema
+cd ..\..\packages\wasmvm\wasmlib
 schema -core -go -rust -ts -client -force
 del /s /q d:\work\node_modules\wasmlib\*.* >nul:
 del /s /q d:\work\node_modules\wasmclient\*.* >nul:
-xcopy /s /q d:\Work\go\github.com\iotaledger\wasp\packages\vm\wasmlib\ts\wasmclient d:\work\node_modules\wasmclient
-xcopy /s /q d:\Work\go\github.com\iotaledger\wasp\packages\vm\wasmlib\ts\wasmlib d:\work\node_modules\wasmlib
+xcopy /s /q d:\Work\go\github.com\iotaledger\wasp\packages\wasmvm\wasmlib\ts\wasmclient d:\work\node_modules\wasmclient
```
lmoe, line 7: Should these lines go into the repo as well? Maybe make it relative?

lmoe (approved):

contracts/wasm/core_build.cmd
BugFreeSoftware: We definitely need to set up some node module for the typescript libraries at some point. Then this can go away.

PR #789, opened 2022-02-07
State manager code and unit tests

Juliusan:
# Description of change

State manager package compiles and all its unit tests pass. I tried to leave the code unchanged as much as possible. The biggest change happened to state manager testing environment. It is now a lot simpler.

## Type of change

Unit tests in following folders were run and they pass:
- `packages/chain/statemgr`
- `packages/chain/mempool`
- `packages/vm/core/testcore_stardust`
- `packages/vm/core/testcore_stardust/sbtests`
- `packages/vm/core/testcore_stardust/sbtests/sbtestsc`
Other unit tests should not be affected by these changes, many packages still do not compile.

## Change checklist

- [ ] I have followed the [contribution guidelines](https://wiki.iota.org/smart-contracts/contribute) for this project: _cannot do this as the whole project does not compile_
- [ ] I have selected the `develop` branch as the target branch: _these changes must go tu `stardust-vm` branch_
- [ ] I have checked that new and existing unit tests pass locally with my changes: _I have checked all the tests that I know are supposed to pass in `stardust-vm`_


packages/dkg/messages.go
```diff
@@ -16,7 +16,8 @@ import (
 	"io"
 	"time"
 
-	"github.com/iotaledger/wasp/packages/cryptolib"
+	"github.com/iotaledger/hive.go/crypto/ed25519"
```
jorgemmsilva, line 19: we should use cryptolib everywhere, remove all references from hive.

Juliusan: I wholeheartedly agree that we should use a wrapper in `cryptolib` for everything. However, we had a lengthy discussion [here](https://iotafoundation.slack.com/archives/G01F91K2CRX/p1642606163179400) and several single post threads before, and I was directed towards "change as little as possible" attitude. Hive package is all over dkg, state manager, consensus and other parts of the code. This would mean many changes and polluted diff. Moreover, Hive keys are arrays, and cryptolib ones are slices. This introduces a delicate changes in e.g. comparing them.
To sum up, I am ready to go any way we decide, but I need a decision to stop going back and fourth about this. @lunfardo314, @kape1395.

jorgemmsilva: I feel that it's worth it to do that now, otherwise we will get a mismatch of crypto libraries over the code (hive/iota.go) :confused: .


also we probably won't get many "sneaky" bugs as we think.
example, by changing this:
https://github.com/iotaledger/wasp/blob/stardust-vm/packages/peering/domain/domain.go#L18
to :

```golang
	nodes       map[*cryptolib.PublicKey]peering.PeerSender
```
all the comparisons will start failing with:
```
(operator == not defined for "crypto/ed25519".PublicKey)
```
so we just need to change them to `bytes.Equal`

jorgemmsilva (approved):

packages/dkg/messages.go
Juliusan: There is another problem: dkg uses public key as a key of the map. This cannot be done if public key is implemented by a slice. It must be workarounded. Why the public key must be a slice instead of fixed size array is beyond me, but yes, I agree with you. I just want to hear it from more people :-)

packages/cryptolib/keypair.go
```diff
@@ -1,8 +1,10 @@
 package cryptolib
 
 import (
+	//	cr "crypto"
```
howjmay: why don't we just remove them?

Juliusan: Thanks, good catch.

PR #798, opened 2022-02-08
dust deposit of internal outputs are financed by the sender

lunfardo314:
- dust deposit for the first appearance of native token on the chain now is taken from the caller account (not from the common account)
- (bug fix, uncovered by tests): `modify supply` dust deposit adjustment now is taken from the allowance, not directly debited from the caller's account
- multiple adjustments/code optimizations


packages/vm/vmcontext/runreq.go
```diff
@@ -87,21 +87,34 @@ func (vmctx *VMContext) creditAssetsToChain() {
...
+	if account == nil {
+		account = commonaccount.Get(vmctx.ChainID())
+	}
+	vmctx.creditToAccount(account, vmctx.req.Assets())
+
+	// adjust the sender's account with the dust consumed or returned by internal UTXOs
+	// if iotas in the sender's account is not enough for the dust deposit of newly created TNT outputs
+	// it will panic with exceptions.ErrNotEnoughFundsForInternalDustDeposit
+	// TNT outputs will use dust deposit from the caller
+	// TODO remove attack vector when iotas for dust deposit is not enough and the request keeps being skipped
```
jorgemmsilva, line 107: what will happen in the case where a user sends N native tokens with the minimum dust deposit for that output. The VM will try to split the tokens to each separate output but there aren't enough iotas to cover many dust deposits?

jorgemmsilva (approved):

packages/vm/vmcontext/runreq.go
lunfardo314: It will be taking dust deposit from the sender’s account, not just the output. If not enough iotas, it will panic with `not enough iotas for internal dust deposit` and request will be skipped. Next time it will come again and so on it is an attack, hence TODO. 
Returning outputs problem is heavier with big number of tokens in the output (currently max 64) because it will be occupyimg anchors’s space
 So, if no return conditions, funds will be lost, ugly but there are not much option.
To somehow mitigate the problem, I have an idea of introducing `ignored request` category: not consumed input but in blocklog saved receipt that os is processed

jorgemmsilva: I think `ignored requests` is a good idea.
I was pondering if sending users into "negative balance" could be a solution: "we process your deposit and cover your dust, fees, but you can only send requests when the debt is re-payed". 
"ignored requests" seems simpler, and funds would still be safe if "expiry" feature is used.

dessaya (approved):

PR #818, opened 2022-02-14
Add consensus pipe metrics

ivange94:
# Description of change

Consensus pipe metrics

## Type of change

Run wasp and visit metrics dashboard

## Change checklist

### DIFF
--- a/packages/chain/chain.go
+++ b/packages/chain/chain.go
@@ -64,6 +64,7 @@ type ChainRequests interface {
 type ChainMetrics interface {
 	GetNodeConnectionMetrics() nodeconnmetrics.NodeConnectionMessagesMetrics
 	GetConsensusWorkflowStatus() ConsensusWorkflowStatus
+	GetConsensusPipeMetrics() ConsensusPipeMetrics
 }
 
 type Chain interface {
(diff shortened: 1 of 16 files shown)


packages/chain/chain.go
```diff
@@ -232,6 +234,16 @@ type ConsensusWorkflowStatus interface {
...
 
+type ConsensusPipeMetrics interface {
+	GetEventStateTransitionMsgPipeSize() int
+	GetEventSignedResultMsgPipeSize() int
+	GetEventSignedResultAckMsgPipeSize() int
+	GetEventInclusionStateMsgPipeSize() int
+	GetEventACSMsgPipeSize() int
+	GetEventVMResultMsgPipeSize() int
+	GetEventTimerMsgPipeSize() int
+}
```
dessaya, line 245: Nit: if there is only one implementation of this interface, and it is just a plain container with no logic, I would just drop the interface and put the struct here.

This would also allow to drop the `model.ConsensusPipeMetrics`, and just use this one, since they are identical.

dessaya (approved):

packages/chain/chain.go
ivange94: I followed the pattern I found with other objects like `ConsensusWorkflowStatus` where an interface is passed around and a private implementation object event when there's only one implementation of this interface. Do you think we need to refactor those too?

dessaya: If they fall into the same category (only a single implementation of the interface, and it is just a plain container with no logic), then yes, I think we could simplify all of it. But it is not really important, and it can be made in another PR.

Juliusan (approved):

PR #819, opened 2022-02-14
feat: views charge gas, refactored external views calls

jorgemmsilva:
# Description of change

- viewcontext deprecated
- external views now use `vmcontext.CreateVMContextForViewCall(chain)`
- external views now have gas limits

## Change checklist


packages/chain/chain.go
```diff
@@ -36,7 +35,13 @@ type ChainCore interface {
...
 	GetCandidateNodes() []*governance.AccessNodeInfo // All the current candidates.
+	VirtualStateAccess() state.VirtualStateAccess
+	GetAnchorOutput() (*iotago.AliasOutput, *iotago.UTXOInput)
 	Log() *logger.Logger
+}
+
+type ChainCoreMock interface {
+	ChainCore
 
 	// FIXME these methods should not be part of the chain interface just for the need of mocking
```
dessaya, line 46: This comment can be removed?

packages/vm/vmcontext/call.go
```diff
@@ -35,16 +31,23 @@ func (vmctx *VMContext) callByProgramHash(targetContract, epCode iscp.Hname, par
...
 	vmctx.pushCallContext(targetContract, params, allowance)
 	defer vmctx.popCallContext()
 
 	// distinguishing between two types of entry points. Passing different types of sandboxes
 	if ep.IsView() {
-		if epCode == iscp.EntryPointInit {
-			panic(xerrors.Errorf("'%v': target=(%s, %s)",
-				ErrEntryPointCantBeAView, vmctx.req.CallTarget().Contract, epCode))
-		}
+		// I don't think this can happen because iscp.EntryPointInit is not a view...
```
dessaya: It may happen if some "forges" the request to call init as if it was a view call. Anyway, I don't know why this check is a special case. IMO it should be: If the called entry point is not a view (init or not), abort.

packages/vm/vmcontext/external_view_calls.go
```diff
@@ -0,0 +1,83 @@
...
+	return &VMContext{
+		blockContextCloseSeq: make([]iscp.Hname, 0),
+		callStack:            make([]*callContext, 0),
+		task:                 task,
+		currentStateUpdate:   state.NewStateUpdate(),
+		virtualState:         virtualState,
+	}
+}
+
+// callViewExternal is onlu for internal usage by CallViewExternal
```
dessaya:
```suggestion
// callViewExternal is only for internal usage by CallViewExternal
```

dessaya (approved):

packages/chain/chain.go
jorgemmsilva: I wasn't sure of the reason for this comment, so I opted to leave it in, just in case it needs to be revistited

packages/vm/vmcontext/call.go
jorgemmsilva: >It may happen if some "forges" the request to call init as if it was a view call. 

is that even possible? the entrypoint is specified as a "func", so it shouldn't be possible to call "as a view", I think
https://github.com/iotaledger/wasp/blob/master/packages/iscp/coreutil/contract.go#L113
I left the comment just in case I was missing something

>IMO it should be: If the called entry point is not a view (init or not), abort.

that won't work because it can be a "view call from inside a SC"

packages/vm/sandbox/sandboxbase.go
```diff
@@ -0,0 +1,108 @@
...
+	"math/big"
+
+	iotago "github.com/iotaledger/iota.go/v3"
+	"github.com/iotaledger/wasp/packages/iscp"
+	"github.com/iotaledger/wasp/packages/iscp/assert"
+	"github.com/iotaledger/wasp/packages/vm/execution"
+	"github.com/iotaledger/wasp/packages/vm/gas"
+)
+
+type Sandboxbase struct {
```
dessaya: `SandboxBase`?

packages/vm/sandbox/sandboxview.go
```diff
@@ -4,117 +4,31 @@
...
-	return s.vmctx.GetIotaBalance(s.vmctx.AccountID())
+func NewSandboxView(ctx execution.WaspContext) iscp.SandboxView {
+	ret := &sandboxView{}
+	ret.Ctx = ctx
+	return ret
 }
 
-func (s *sandboxView) BalanceNativeToken(id *iotago.NativeTokenID) *big.Int {
-	return s.vmctx.GetNativeTokenBalance(s.vmctx.AccountID(), id)
+func (s *Sandboxbase) State() kv.KVStoreReader {
```
dessaya: shouldn't the receiver be `sandboxView`? Or move the method to `sandboxbase.go`?

packages/vm/viewcontext/viewcontext.go
```diff
@@ -1,47 +1,188 @@
...
+	"github.com/iotaledger/wasp/packages/vm/execution"
+	"github.com/iotaledger/wasp/packages/vm/gas"
 	"github.com/iotaledger/wasp/packages/vm/processors"
+	"github.com/iotaledger/wasp/packages/vm/sandbox"
+	"github.com/iotaledger/wasp/packages/vm/vmcontext"
 	"golang.org/x/xerrors"
 )
 
+// viewcontext implements the needed infrastucture to run external view calls, its more lightweight than vmcontext
 type Viewcontext struct {
```
dessaya: `ViewContext`? I know this is not new, but since we are already refactoring...

packages/vm/vmcontext/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+	iotago "github.com/iotaledger/iota.go/v3"
+	"github.com/iotaledger/wasp/packages/hashing"
+	"github.com/iotaledger/wasp/packages/iscp"
+	"github.com/iotaledger/wasp/packages/kv"
+	"github.com/iotaledger/wasp/packages/kv/dict"
+	"github.com/iotaledger/wasp/packages/vm/gas"
+	"github.com/iotaledger/wasp/packages/vm/sandbox"
+)
+
+type contractSandbox struct {
```
dessaya, line 18: just `sandbox`?

dessaya (approved):

packages/vm/sandbox/sandboxview.go
jorgemmsilva: nice catch! thanks :+1:

packages/vm/vmcontext/sandbox.go
jorgemmsilva: that was the name initially, but it collided with the `sandbox` package :(

PR #820, opened 2022-02-14
Stardust error handling

lmoe:
# Description of change

This PR introduces an additional erroring layer which allows core/native contracts to add a typing to errors that happen during the execution of the smart contract. These errors will eventually land inside the receipt after the execution roundtrip is done. 

These errors are designed with low storage space in mind. The created receipt itself does not contain the full error but only an error id and supplied params. Errors are defined once per contract, are stored in the state, and are retrievable afterwords for detailed inspection. 

## Type of change

A new testfile was added and contains a few success/error cases 

Run `packages/vm/core/testcore_stardust/errors_test.go`

## Change checklist


packages/util/panicutil.go
```diff
@@ -1,6 +1,9 @@
 package util
 
-import "golang.org/x/xerrors"
+import (
+	errorlib "errors"
```
dessaya: Why the rename?

packages/vm/core/blocklog/interface.go
```diff
@@ -306,7 +306,7 @@ func EventLookupKeyFromBytes(r io.Reader) (*EventLookupKey, error) {
 // RequestReceipt represents log record of processed request on the chain
 type RequestReceipt struct {
 	Request       iscp.Request // TODO request may be big (blobs). Do we want to store it all?
-	ErrorStr      string
+	Error         error
```
dessaya: I think it is confusing to mix the `error` type here, because `error` is intended for Go errors, and here we are storing an error produced by a SC. Can this member just be a `*vmerrors.Error`? And also ditch the `DetailedError` method?

(It would probably mean that we need a constructor like `NewRequestReceipt(..., err error, ...)` that converts the `error` to a `*vmerrors.Error` if necessary)

packages/vm/core/errors/commonerrors/error_collection_interface.go
```diff
@@ -0,0 +1,10 @@
+package commonerrors
+
+import (
+	"github.com/iotaledger/wasp/packages/vm/vmerrors"
+)
+
+type IErrorCollection interface {
```
dessaya: Nit: This is not Java :rofl: 

interface `IErrorCollection` -> `ErrorCollection`
implementation `ErrorCollection` -> `errorCollection`

(Or just drop the interface... I don't see the need for the interface if there is a single implementation?)

packages/vm/core/errors/commonerrors/error_collection.go
```diff
@@ -0,0 +1,39 @@
...
+	errors map[uint16]*errors.ErrorDefinition
+}
+
+func NewErrorCollection() IErrorCollection {
+	return &ErrorCollection{
+		errors: map[uint16]*errors.ErrorDefinition{},
+	}
+}
+
+func (e *ErrorCollection) Get(errorId uint16) (*errors.ErrorDefinition, error) {
```
dessaya: why not just return `*errors.ErrorDefinition`?

packages/vm/core/errors/commonerrors/error_collection.go
```diff
@@ -0,0 +1,39 @@
+package commonerrors
+
+import (
+	errors "github.com/iotaledger/wasp/packages/vm/vmerrors"
+	"math"
+)
+
+// ErrorCollection implements IErrorCollection. Is used for global errors. Does not require vm context state.
+type ErrorCollection struct {
```
dessaya: Why not `type ErrorCollection map[uint16]*errors.ErrorDefinition`?

packages/vm/core/errors/commonerrors/external.go
```diff
@@ -0,0 +1,22 @@
...
+	errorDefinition, err := globalErrorCollection.Register(errorId, messageFormat)
+
+	if err != nil {
+		panic(err)
+	}
+
+	return errorDefinition
+}
+
+func GetGlobalErrorCollection() IErrorCollection {
```
dessaya: Propose calling this the "core errors" collection instead, and:

`RegisterGlobalError()` -> `RegisterCoreError()`
`GetGlobalErrorCollection()` -> `CoreErrors()`

Maybe rename `commonerrors` -> `coreerrors`? (In that case `commonerrors.GetGlobalErrorCollection()` -> `coreerrors.All()`)

(Or maybe move everything to `vm/core`, then `commonerrors.GetGlobalErrorCollection()` -> `core.AllErrors()`. Flat is better than nested :stuck_out_tongue: )

packages/vm/core/errors/commonerrors/error_collection.go
```diff
@@ -0,0 +1,39 @@
+package commonerrors
```
dessaya: I'm not sure this package is the right place for all the new stuff. Why not in `vmerrors`?

(IMO we currently have too many "errors" packages, like `core/errors`, `core/errors/commonerrors`, `vmerrors`, etc. and I find it difficult to understand the purpose of each one).

packages/vm/core/errors/commonerrors/external.go
```diff
@@ -0,0 +1,22 @@
...
+	return globalErrorCollection
+}
+
+// SandboxErrorMessageResolver has the signature of ErrorMessageResolver to provide a way to resolve the error format
```
dessaya: This comment should go next to the implementation of `SandboxErrorMessageResolver`?

packages/vm/core/errors/external.go
```diff
@@ -0,0 +1,32 @@
+package errors
+
+import (
+	"github.com/iotaledger/wasp/packages/iscp"
+	"github.com/iotaledger/wasp/packages/kv/codec"
+	"github.com/iotaledger/wasp/packages/kv/dict"
+	"github.com/iotaledger/wasp/packages/vm/vmerrors"
+)
+
+func SandboxErrorMessageResolver(ctx iscp.SandboxView) func(*vmerrors.Error) (string, error) {
```
dessaya: This is not called from anywhere? :thinking:

packages/vm/core/errors/impl.go
```diff
@@ -0,0 +1,74 @@
...
+		panic(commonerrors.ErrMessageFormatEmpty)
+	}
+
+	errorId := vmerrors.GetErrorIdFromMessageFormat(errorMessageFormat)
+
+	errorDefinition, err := e.Register(errorId, errorMessageFormat)
+
+	if err != nil {
+		panic(err)
+	}
```
dessaya:
```suggestion
	ctx.RequireNoError(err)
```

packages/vm/core/errors/impl.go
```diff
@@ -0,0 +1,74 @@
...
+func funcRegisterError(ctx iscp.Sandbox) dict.Dict {
+	ctx.Log().Debugf("Registering error")
+	e := NewStateErrorCollectionWriter(ctx.State(), ctx.Caller().Hname())
+
+	params := kvdecoder.New(ctx.Params())
+	errorMessageFormat := params.MustGetString(ParamErrorMessageFormat)
+
+	if len(errorMessageFormat) == 0 {
+		panic(commonerrors.ErrMessageFormatEmpty)
+	}
```
dessaya:
```suggestion
	ctx.Require(len(errorMessageFormat) > 0, commonerrors.ErrMessageFormatEmpty))
```

although I'm not sure the types will match :stuck_out_tongue:

packages/vm/core/errors/impl.go
```diff
@@ -0,0 +1,74 @@
...
+		e = commonerrors.GetGlobalErrorCollection()
+	} else {
+		e = NewStateErrorCollectionReader(ctx.State(), iscp.Hname(contract))
+	}
+
+	errorDefinition, err := e.Get(errorId)
+
+	if err != nil {
+		panic(err)
+	}
```
dessaya:
```suggestion
	ctx.RequireNoError(err)
```

packages/vm/core/errors/commonerrors/error_collection.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+}
+
+func (e *ErrorCollection) Get(errorId uint16) (*vmerrors.ErrorDefinition, error) {
+	if errorDefinition, ok := e.errors[errorId]; ok {
+		return errorDefinition, nil
+	}
+
+	return nil, nil
+}
```
dessaya: It never returns `error`?

packages/vm/vmerrors/error.go
```diff
@@ -0,0 +1,266 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/iotaledger/hive.go/marshalutil"
+	"github.com/iotaledger/wasp/packages/hashing"
+	"golang.org/x/xerrors"
+	"hash/crc32"
+)
+
+type ErrorDefinition struct {
```
dessaya: Maybe `ErrorClass`? `ErrorKind`? `ErroFamily`?

packages/vm/vmerrors/error.go
```diff
@@ -0,0 +1,266 @@
...
+	prefixId      uint32
```
dessaya: Why not make this `hname iscp.Hname?`? Is it used for anything else than identifying the contract hname?

packages/vm/vmerrors/error.go
```diff
@@ -0,0 +1,266 @@
...
+	id            uint16
+	messageFormat string
+}
+
+func NewErrorDefinition(prefixId uint32, errorId uint16, messageFormat string) *ErrorDefinition {
```
dessaya: Isn't the `errorId` always derived from the `messageFormat`? Then why receive both as parameters? Could we even not store the `id` in the struct?

packages/vm/core/errors/commonerrors/error_collection_interface.go
lmoe: Yeah :D

It is even implemented thrice: 
https://github.com/iotaledger/wasp/blob/stardust-error-handling/packages/vm/core/errors/state_error_collection.go

One is the general error collection which just uses a map, the others are using the KVWriter & KVReader. Usage is mostly here: https://github.com/iotaledger/wasp/blob/stardust-error-handling/packages/vm/core/errors/impl.go#L56

packages/vm/core/errors/commonerrors/error_collection.go
lmoe: Because the KV* implementations return errors while the general error implementation does not. I've figured that a lot more can go wrong with the KV* stores, therefore I added the error there and ignored the error in the general implementation.

packages/vm/core/errors/commonerrors/error_collection.go
lmoe: To keep the structure the same as this here: https://github.com/iotaledger/wasp/blob/stardust-error-handling/packages/vm/core/errors/state_error_collection.go#L17

packages/vm/core/errors/commonerrors/error_collection.go
lmoe: I will think about something.

My idea was that the whole error core contract has access to either the general errors but also the contract based ones. That's why it contains both. The problem is that as soon as we move the general parts into the same core/errors package we get import loops because a lot of things are now depending on it. I wasnt sure how to fix this elegantly.

packages/vm/core/errors/commonerrors/error_collection.go
lmoe: No, either the error exists, or it doesn't. I don't see a requirement to add an artifial error for this case here. The KV implementations can produce errors though and this implementation has do abide the interface.

packages/vm/vmerrors/error.go
lmoe: Because parts of iscp include the error parts and when I use iscp.HName there, I will produce import loops again. I changed everything to uint32 because that's basically the hname. 


I might need a bit of help if we want to switch back to iscp.hname

packages/vm/vmerrors/error.go
lmoe: The id is currently calculated based on the message format, yes. While I would keep the storage of the id - to not have to calculate the id every time you need to access it - I can include the calculation in that function and only require prefix and message.

packages/iscp/coreutil/globalindex.go
```diff
@@ -1,11 +1,11 @@
 package coreutil
 
 import (
+	"github.com/iotaledger/wasp/packages/vm/core/errors/commonerrors"
```
jorgemmsilva: feels weird that we are importing stuff from the VM into coreutil. should the "common errors" be part of the coreutil, or some self-contained package?

packages/vm/core/errors/commonerrors/error_collection.go
jorgemmsilva: but why do we need the struct?

packages/vm/vmerrors/error.go
```diff
@@ -0,0 +1,266 @@
...
+	if err = json.Unmarshal(params, &e.params); err != nil {
+		return err
+	}
+
+	return err
+}
+
+type ErrorMessageResolver func(*Error) (string, error)
+
+func ErrorFromBytes(mu *marshalutil.MarshalUtil, errorMessageResolver ErrorMessageResolver) (*Error, error) {
```
jorgemmsilva:
```suggestion
func ErrorFromMarshalUtil(mu *marshalutil.MarshalUtil, errorMessageResolver ErrorMessageResolver) (*Error, error) {
```

packages/vm/core/blocklog/interface.go
```diff
@@ -344,17 +337,48 @@ func RequestReceiptFromMarshalUtil(mu *marshalutil.MarshalUtil) (*RequestReceipt
 	if ret.Request, err = iscp.RequestDataFromMarshalUtil(mu); err != nil {
 		return nil, err
 	}
+
+	if isError, err := mu.ReadBool(); err != nil {
+		return nil, err
+	} else if !isError {
+		return ret, nil
+	}
```
jorgemmsilva, line 343: nitpick: else not needed
```suggestion
	}
	if !isError {
		return ret, nil
	}
```

packages/vm/core/blocklog/interface.go
```diff
@@ -344,17 +337,48 @@ func RequestReceiptFromMarshalUtil(mu *marshalutil.MarshalUtil) (*RequestReceipt
...
 		WriteUint64(r.GasBurned).
 		WriteUint64(r.GasFeeCharged)
+
 	iscp.RequestDataToMarshalUtil(r.Request, mu)
+
+	if r.Error == nil {
+		mu.WriteBool(false)
+	} else {
+		mu.WriteBool(true)
+		r.DetailedError().Serialize(mu)
```
jorgemmsilva: why `.Serialize(mu)`? its not wrong per-se, but its a different pattern from everywhere else in this codebase

packages/vm/core/blocklog/interface.go
```diff
@@ -344,17 +337,48 @@ func RequestReceiptFromMarshalUtil(mu *marshalutil.MarshalUtil) (*RequestReceipt
...
+
 	return ret, nil
 }
 
+func (r *RequestReceipt) DetailedError() *vmerrors.Error {
+	if r.Error == nil {
+		return nil
+	}
+
+	if err, ok := r.Error.(*vmerrors.Error); ok {
```
jorgemmsilva: `vmerrors` feels like the wrong name. This can be an error coming from any SC, not exclusively from the VM, right?

packages/vm/core/errors/commonerrors/error_collection.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (e *ErrorCollection) Register(errorId uint16, messageFormat string) (*vmerrors.ErrorDefinition, error) {
+	if len(messageFormat) > math.MaxUint16 {
+		return nil, ErrErrorMessageTooLong
+	}
```
jorgemmsilva: why is message length limited to `max.Uint16`? is there some reason for this value? seems kind of random :sweat_smile:

lmoe: The idea was simply to save some space. I've figured that we won't be seeing error messages with the length of 65535 bytes, so I've limited it. During the serialization we don't need to store a uint32 for the length of the text and save a few bytes.

packages/vm/core/blocklog/interface.go
lmoe: Yeah true. I wasnt sure about the naming at all. "errors" is too generic and collides with golangs `errors` so I named it vmerrors so it's more distinct. Maybe we can discuss the whole naming today. I'm at the moment pretty much clueless about the proper naming :D

packages/vm/core/blocklog/interface.go
lmoe: That was the issue I was talking about with error.is and require.IsError. It compiles all fine when I use the custom error, but require.IsError and error.is has a weird behavior when I use the custom type. I think it was that those detect an error even when it's nil. 

That's why we had discussed to revert back to error and use a function to return the detailed error.

packages/vm/core/errors/commonerrors/error_collection.go
jorgemmsilva: I think the cost to store the error should scale accordingly (in gas).
Agree that there should be a max cap, but should be defined it as a constant, even if the value will be maxUInt16.

```go
const errorMsgMaxLength = 12345
```

helps with readability, we don't have to second guess why we're using that value

packages/vm/core/blocklog/interface.go
dessaya: But unless I'm missing something, once the error is stored in the receipt we should no longer care what was its original type. We always store it as a `*vmerrors.Error` and never treat it as a Go `error`.

Isn't it like this?

1. Panic with error (of any type, even string) from inside the SC
2. Recover error in the VM.
3. Cast or convert error to `*vmerrors.Error` (whatever its original type) and store in the receipt

packages/vm/vmerrors/error.go
dessaya: I think it would be benefitial to make that change.

Most of the times, when Go complains about import loops it is because we are trying to be too fancy with packages :stuck_out_tongue:. Maybe just put all declarations in the `iscp` package?

packages/vm/vmerrors/error.go
dessaya: Yes, I think it would make sense to calculate the id here instead of requiring every caller to do so.

packages/vm/core/blocklog/interface.go
lmoe: That only works when we create a variable first, like this: 
```	
var isError bool
	
	if isError, err = mu.ReadBool(); err != nil {
		return nil, err
	}

	if !isError {
		return ret, nil
	}
```

Otherwise it complains about isError not existing at the !isError check. I think because the err != nil check next to readBool puts isError into another scope. 

I would keep it how it is, because that makes it just longer. What do you think?

packages/vm/core/errors/coreerrors/error_collection.go
```diff
@@ -0,0 +1,38 @@
+package coreerrors
+
+import (
+	"github.com/iotaledger/wasp/packages/iscp"
+	"math"
+)
+
+// CoreErrorCollection implements IErrorCollection. Is used for global errors. Does not require vm context state.
```
dessaya:
```suggestion
// CoreErrorCollection implements ErrorCollection. Is used for global errors. Does not require vm context state.
```

packages/vm/core/errors/coreerrors/external.go
```diff
@@ -0,0 +1,21 @@
+package coreerrors
+
+import (
+	"github.com/iotaledger/wasp/packages/iscp"
+)
+
+var coreErrorCollection ErrorCollection = NewCoreErrorCollection()
+
+func RegisterGlobalError(messageFormat string) *iscp.VMErrorTemplate {
```
dessaya: Suggest calling it just `Register` (since `coreerrors.RegisterGlobalError` sounds too long and stuttery).

dessaya (approved):

tools/wasp-cli/chain/blocklog.go
```diff
@@ -93,7 +93,7 @@ func logReceipt(receipt *blocklog.RequestReceipt, index ...uint16) {
 		{K: "Contract Hname", V: req.CallTarget().Contract.String()},
 		{K: "Entry point", V: req.CallTarget().EntryPoint.String()},
 		{K: "Arguments", V: argsTree},
-		{K: "Error", V: errMsg},
+		{K: "VMError", V: errMsg},
```
dessaya: this change is unintended, right?

packages/vm/core/blocklog/interface.go
```diff
@@ -306,7 +304,7 @@ func EventLookupKeyFromBytes(r io.Reader) (*EventLookupKey, error) {
 // RequestReceipt represents log record of processed request on the chain
 type RequestReceipt struct {
 	Request       iscp.Request // TODO request may be big (blobs). Do we want to store it all?
-	ErrorStr      string
+	Error         *iscp.VMError
```
jorgemmsilva: shouldn't this just be the standard golang error?

the `AsGoError()` thing seems odd to me.

don't we want to save just the "error ID" (may be hname_prefix+id) in the receipt?

dessaya: Agree, we should probably store just the error ID + args here, and if the client wants to resolve the error message, they may do so manually.

> shouldn't this just be the standard golang error?

The "error" stored in the receipt is related to an error produced inside a SC which may be written in any language; and the receipt may be consumed in other languages too. So IMO it doesn't really make sense to mix it with Go's `error` interface; the important thing is to be able to reproduce the error message.

> the `AsGoError()` thing seems odd to me.

It is odd, because Go's "feature" of making different interfaces nil values different is odd.
But again, maybe we shouldn't even bother implementing the `error` interface...

lmoe: We need the error id, the prefix id and the args to resolve an error. This is basically the complete VMError. I really see no point to separate out the three properties.

Concatting prefixId+errorId will cause issues in the core contract as I need both values separated. 

Lets discuss this later

packages/vm/core/blocklog/interface.go
jorgemmsilva: oh yeah, you're right. let's just keep as-is

packages/solo/req.go
```diff
@@ -396,7 +404,37 @@ func (ch *Chain) EstimateGasOffLedger(req *CallParams, keyPair *cryptolib.KeyPai
 	}
 	r := req.NewRequestOffLedger(ch.ChainID, keyPair)
 	res := ch.estimateGas(r)
-	return res.Receipt.GasBurned, res.Receipt.GasFeeCharged, res.Receipt.Error()
+
+	return res.Receipt.GasBurned, res.Receipt.GasFeeCharged, res.Receipt.Error.AsGoError()
+}
+
+// ErrorMessageResolver has the signature of VMErrorMessageResolver to provide a way to resolve the error format
+func (ch *Chain) ErrorMessageResolver() func(vmError *iscp.UnresolvedVMError) (string, error) {
```
jorgemmsilva:
```suggestion
func (ch *Chain) ErrorMessageResolver(vmError *iscp.UnresolvedVMError) (string, error) {
```
wouldn't this work? why do we need to return a new function?

packages/vm/core/errors/impl.go
```diff
@@ -0,0 +1,67 @@
...
+
+func funcGetErrorMessageFormat(ctx iscp.SandboxView) dict.Dict {
+	params := kvdecoder.New(ctx.Params())
+
+	contract := params.MustGetUint32(ParamContractHname)
+	errorId := params.MustGetUint16(ParamErrorId)
+
+	var e coreerrors.ErrorCollection
+
+	if contract == math.MaxUint32 {
```
jorgemmsilva:
```suggestion
	if contract == math.MaxUint32 {
```
please leave a comment explaining why we are comparing the contract hname to maxUint32

jorgemmsilva (approved):

packages/vm/core/errors/impl.go
lmoe: Because hname with the value of 0 is not allowed and we need a way to differ between core error and contract error types. We could also introduce a boolean which toggles them. I considered maxuint32 to be fine in that regard. 

Any suggestions?

packages/solo/req.go
lmoe: Yep! It's required for the Sandbox resolver implementation but not for this one here.

packages/vm/core/errors/impl.go
jorgemmsilva: just leave a comment, please. So people don't have to guess why we're comparing an hname to maxUint32...

dessaya: Another option is to define a constant somewhere

```
const NilHname = math.MaxUint32
```

then

```
if contract == NilHname
```

PR #824, opened 2022-02-15
Cryptolib

Juliusan:
# Description of change

As discussed in the previous meeting, all the different use cases of various packages for public/private key and related functionality changed to use the single implementation in `cryptolib` package.

## Type of change

The unit tests in the following folders pass:
- `packages/transaction`
- `packages/vm/core/testcore_stardust`
- `packages/vm/core/testcore_stardust/sbtests/`
- `packages/vm/core/testcore_stardust/sbtests/sbtestsc/`
- `packages/peering/`
- `packages/peering/domain/`
- `packages/peering/group/`
- `packages/peering/lpp/`
- `packages/chain/mempool/`
- `packages/chain/statemgr/`

Many parts of the code still do not compile and cannot be checked. I tried to remove all the possible usage cases of `crypto/ed25519` and `github.com/iotaledger/hive.go/crypto/ed25519` from there too, but the correctness cannot be checked.

## Change checklist


packages/cryptolib/private_key.go
```diff
@@ -0,0 +1,63 @@
+package cryptolib
+
+import (
+	cr "crypto"
+	crypto "crypto/ed25519"
```
dessaya: Why the renames?

packages/cryptolib/private_key.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return &PrivateKey{privateKeyBytes}, nil
+}
+
+func NewPrivateKeyFromSeed(seed Seed) *PrivateKey {
+	var seedByte [SeedSize]byte = seed
+	return &PrivateKey{crypto.NewKeyFromSeed(seedByte[:])}
+}
+
+func (pkT *PrivateKey) isValid() bool {
```
dessaya, line 37: Why `pkT` instead of `p` or `pk`? What does the `T` stand for? :thinking:

dessaya (approved):

packages/cryptolib/private_key.go
Juliusan: Throughout all the wasp `crypto/ed25519` is renamed to `crypto`. I wanted to avoid the confusion. I'll drop both  renames in all the places.

packages/cryptolib/private_key.go
Juliusan: `T` stands for forbidden word `this` :-)

PR #826, opened 2022-02-17
feat: Add shell script for building contracts

howjmay:
# Description of change

Please write a summary of your changes and why you made them. 

## Type of change

- Documentation Fix

## Change checklist

- [ ] I have followed the [contribution guidelines](https://wiki.iota.org/smart-contracts/contribute) for this project

### DIFF
--- a/contracts/wasm/all_build.sh
+++ b/contracts/wasm/all_build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+bash core_build.sh
+bash go_all.sh -force
+# bash ts_all.sh -force
+bash rust_all.sh -force
+bash update_hardcoded.sh
--- a/contracts/wasm/all_build.sh
+++ b/contracts/wasm/all_build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 bash core_build.sh
 bash go_all.sh -force
-# bash ts_all.sh -force
+bash ts_all.sh -force
 bash rust_all.sh -force
 bash update_hardcoded.sh
--- a/contracts/wasm/cleanup.sh
+++ b/contracts/wasm/cleanup.sh
@@ -1,22 +1,25 @@
-find . -type f -name "consts.*" -delete
-find . -type f -name "contract.*" -delete
-find . -type f -name "keys.*" -delete
-find . -type f -name "lib.*" -delete
-find . -type f -name "params.*" -delete
-find . -type f -name "results.*" -delete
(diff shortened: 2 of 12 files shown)


contracts/wasm/all_build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+bash core_build.sh
+bash go_all.sh -force
+# bash ts_all.sh -force
+bash rust_all.sh -force
+bash update_hardcoded.sh
```
Juliusan: Would it be much trouble to use `make` to avoid such scripts?

howjmay: I think what you mentioned is pretty reasonable, but now the corresponding build scripts for  windows hasn't merged into makefile yet, therefor, the current implementation may be more symmetric to the windows ones. What do you think?

BugFreeSoftware: Also: we need to really take a good look at our SC build process.
We have all the files generated by schema tool in the repo, which are essentially build artifacts, but they are necessary for the tests to be able to run.
So we need to somehow change our git flow so that git will build schema tool, then generate the go code, and only then run test tests on the contracts/wasm folder.

contracts/wasm/cleanup.sh
```diff
@@ -0,0 +1,22 @@
+find . -type f -name "consts.*" -delete
+find . -type f -name "contract.*" -delete
+find . -type f -name "keys.*" -delete
+find . -type f -name "lib.*" -delete
+find . -type f -name "params.*" -delete
+find . -type f -name "results.*" -delete
+find . -type f -name "state.*" -delete
+find . -type f -name "typedefs.*" -delete
+find . -type f -name "types.*" -delete
+find . -type f -name "main.go" -delete
```
jorgemmsilva: could this delete unintended files when executed from the wrong directory, no? :confused:

jorgemmsilva (approved):
It should be okay to merge this, but I'm a bit out of context, so I might not be the right person to review this thoroughly

contracts/wasm/cleanup.sh
howjmay: I have modified this part to a way with more constraints

lmoe:
LGTM. The concerns seem to be fixed already. 
In general I would not `cp` node_modules but use `npm link` when possible. I think we should finally put wasmlib on npmjs. So we can simply include it into the package.json.

PR #834, opened 2022-02-18
chore: Show the real caller of the logging func

howjmay:
# Description of change

The old implementation will show functions in `packages/vm/vmcontext/log.go` as the logging message caller, which is meaningless and helpless for being debug message.

Here we are skipping the 1 caller in `packages/vm/vmcontext/log.go`, so the logging message can show which function is the caller of these logging functions.

## Type of change

## Change checklist

### DIFF
--- a/packages/chain/consensus/action.go
+++ b/packages/chain/consensus/action.go
@@ -20,6 +20,7 @@ import (
 	"github.com/iotaledger/wasp/packages/peering"
 	"github.com/iotaledger/wasp/packages/util"
 	"github.com/iotaledger/wasp/packages/vm"
+	"go.uber.org/zap"
 )
 
 // takeAction triggers actions whenever relevant
@@ -229,7 +230,7 @@ func (c *consensus) prepareVMTask(reqs []iscp.Calldata) *vm.VMTask {
 		Requests:           reqs,
 		Timestamp:          c.consensusBatch.Timestamp,
 		VirtualStateAccess: c.currentState.Copy(),
-		Log:                c.log,
+		Log:                c.log.Desugar().WithOptions(zap.AddCallerSkip(1)).Sugar(),
 	}
 	task.OnFinish = func(_ dict.Dict, err error, vmError error) {
 		// TODO: OnFinish was dropped; move this block to the goroutine that calls vmRunner.Run()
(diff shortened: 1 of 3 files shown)


packages/vm/vmcontext/log.go
```diff
@@ -2,22 +2,23 @@ package vmcontext
...
 import (
 	"github.com/iotaledger/wasp/packages/iscp"
+	"go.uber.org/zap"
 )
 
 var _ iscp.LogInterface = &VMContext{}
 
 func (vmctx *VMContext) Infof(format string, params ...interface{}) {
-	vmctx.task.Log.Infof(format, params...)
+	vmctx.task.Log.Desugar().WithOptions(zap.AddCallerSkip(1)).Sugar().Infof(format, params...)
```
jorgemmsilva: This is a nice change :+1:

But maybe it would be better to make that change elsewhere, because:
>// WithOptions clones the current Logger

I'm thinking that would make more sense to do this only once, rather than on every log.

We could do that when the logger is created.
here for solo:
https://github.com/iotaledger/wasp/blob/stardust-vm/packages/solo/solo.go#L143
and here for the chain plugin:
https://github.com/iotaledger/wasp/blob/stardust-vm/plugins/chains/plugin.go#L39

It probably also makes sense to extend this to all the plugins with a reusable function (`newWaspLogger() *logger.Logger` or something like this):
Selection_013

howjmay: This file `packages/vm/vmcontext/log.go`, is the only place that would have this problem, since the other places the logger is directly triggered. However, in `packages/vm/vmcontext/log.go`, the meaningful place is the caller. Therefore, if adding the option at the place that logger is created, then we will get the wrong place instead.

jorgemmsilva: Oh, I figured this was some configuration of the logger that would be enough to specify once.
If we have to clone every time so be it :(

howjmay: sorry I am kinda don't understand. I think if we use the `zap.AddCallerSkip(1)` as a systemwide configuration, we will get wrong logging info instead.  Therefore, it maybe not clean to add the option here, and annoying to do it every time we call them, I don't have other idea at this moment

jorgemmsilva (approved):

PR #838, opened 2022-02-18
feat(vm) skip request if it exceeds transaction size

dessaya:


packages/parameters/iotago.go
```diff
@@ -6,16 +6,32 @@ import (
...
+
+func (l1 *L1) RentStructure() *iotago.RentStructure {
+	return l1.DeSerializationParameters.RentStructure
+}
+
+func L1ForTesting() *L1 {
+	return &L1{
+		NetworkID:                 0,
+		MaxTransactionSize:        32000,
+		DeSerializationParameters: DeSerializationParametersForTesting(),
```
jorgemmsilva, line 35: I don't think we need this, its only used for syntactical validation as far as I know

jorgemmsilva (approved):

packages/parameters/iotago.go
dessaya: utxodb needs it, because iota.go transaction builder takes it as a parameter :man_shrugging:

PR #841, opened 2022-02-22
[WIP] Feature/jwt auth

lmoe:
# Description of change
This PR adds a new authentication scheme JWT. It introduces a new account and access system which gives each account the possibility to have its own access rules. This also introduces a new wasp-cli command which gives the option to login and authenticate against the server.

Clean up, testing, tests are still missing. 

## Type of change

- Documentation Fix

## Change checklist

- [ ] I have followed the [contribution guidelines](https://wiki.iota.org/smart-contracts/contribute) for this project


config.json
```diff
@@ -21,16 +21,44 @@
...
+    },
+    "bindAddress": "0.0.0.0:9090"
   },
   "dashboard": {
     "auth": {
-      "scheme": "basic",
-      "username": "wasp",
-      "password": "wasp"
+      "jwt": {
+        "accessClaim": "webapi",
```
dessaya: shouldn't be `dashboard`?

packages/authentication/ip_whitelist.go
```diff
@@ -0,0 +1,53 @@
...
+			parts := strings.Split(c.Request().RemoteAddr, ":")
+			if len(parts) == 2 {
+				ip := net.ParseIP(parts[0])
+				if ip != nil && isAllowed(ip) {
+					authContext.isAuthenticated = true
+					return next(c)
+				}
+			}
+
+			fmt.Printf("Blocking request from %s: %s %s", c.Request().RemoteAddr, c.Request().Method, c.Request().RequestURI)
```
dessaya: This should use a logger instead of printing?

packages/authentication/jwt_auth.go
```diff
@@ -0,0 +1,227 @@
...
+		if !allow(claims) {
+			return false
+		}
+
+		return true
+	}
+	return false
+}
+
+func initJWT(duration int, nodeId string, privateKey []byte, accounts *[]accounts.Account, claimValidator ClaimValidator) (*JWTAuth, func(context echo.Context) bool, MiddlewareValidator, error) {
```
dessaya: why not receive `duration time.Duration` instead of `int`? Or at least rename it to `durationHours` for clarity?

packages/authentication/strategy.go
```diff
@@ -0,0 +1,96 @@
...
+	Scheme    string `koanf:"scheme"`
+	AddRoutes bool   `koanf:"addRoutes"`
+
+	JWTConfig         JWTAuthConfiguration         `koanf:"jwt"`
+	BasicAuthConfig   BasicAuthConfiguration       `koanf:"basic"`
+	IPWhitelistConfig IPWhiteListAuthConfiguration `koanf:"ip"`
+}
+
+type JWTAuthConfiguration struct {
+	Duration    int    `koanf:"duration"`
```
dessaya: Rename to `DurationHours` if `time.Duration` cannot be used here?

packages/authentication/jwt_handler.go
```diff
@@ -0,0 +1,112 @@
...
+			"jwt": token,
+		})
+	}
+
+	if contentType == echo.MIMEApplicationForm {
+		cookie := http.Cookie{
+			Name:     "jwt",
+			Value:    token,
+			HttpOnly: true, // JWT Token will be stored in a http only cookie, this is important to mitigate XSS/XSRF attacks
+			Expires:  time.Now().Add(24 * time.Hour),
```
dessaya: shouldn't this be tied to the duration config? :thinking:

packages/authentication/routes.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+func AuthStatusRoute() string {
+	return "/auth/status"
+}
+
+type AuthStatusModel struct {
+	Scheme  string `swagger:"desc(Authentication scheme (jwt, basic, ip))"`
+	AuthURL string `swagger:"desc(JWT only)"`
+}
```
dessaya: Routes and models usually go somewhere else, so that the client code does not have to import the whole server package in order to use this.

`packages/webapi/routes` for routes
`packages/webapi/model` for models

packages/authentication/strategy.go
```diff
@@ -0,0 +1,96 @@
+package authentication
+
+import (
+	"fmt"
+	"github.com/iotaledger/wasp/packages/parameters"
+	"github.com/iotaledger/wasp/packages/registry"
+	"github.com/iotaledger/wasp/plugins/accounts"
```
dessaya: `packages` should never import code from `plugins`; the import order is always `plugins` -> `packages`

plugins/accounts/plugin.go
```diff
@@ -0,0 +1,58 @@
...
+
+func GetAccountByName(name string) *Account {
+	for _, account := range accounts {
+		if account.Username == name {
+			return &account
+		}
+	}
+
+	return nil
+}
```
dessaya: This should probably go to `packages`

config.json
lmoe: This can (and will) be removed after the clean up. The handling happens inside the respective AddAuthorization call.

packages/authentication/jwt_auth.go
lmoe: I had issues with time.Duration. It turned out to be always zero. At the time of writing I wanted to make some progress and let it be an int. Will take a look before opening up the PR.

packages/authentication/jwt_handler.go
lmoe: Yes definitely the cookie expiry date should be equal to the token one. I was still experimenting with making the cookie secure and kept the hard coded time value. 

For example I'm not yet sure if HttpOnly is enough to secure the token. I want to research this a bit more. But I think it should be fine as long as no scripts can access the cookie.

packages/authentication/routes.go
lmoe: Dashboard and WebAPI both need to add some custom routes. Dashboard for the Login form for example, the WebAPI needs a json body sent. I figured that it makes sense to put them into the authentication package because the routes need to have an equal path. Should I introduce some 'authentication_shared" package?

plugins/accounts/plugin.go
```diff
@@ -0,0 +1,58 @@
+package accounts
```
jorgemmsilva: `accounts` is a loaded word in wasp, I propose calilng it `apiAccounts`, `apiAuth` or something like that

packages/authentication/basic_auth.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/labstack/echo/v4"
+	"github.com/labstack/echo/v4/middleware"
+)
+
+func AddBasicAuth(webAPI WebAPI, accounts *[]accounts.Account) {
+	webAPI.Use(middleware.BasicAuth(func(user, password string, c echo.Context) (bool, error) {
+		authContext := c.Get("auth").(*AuthContext)
+
+		for _, v := range *accounts {
+			if user == v.Username && password == v.Password {
```
jorgemmsilva: - comparison should use `subtle.ConstantTimeCompare` to prevent timing attacks.
- could accounts be a map with Username as the key to prevent having to loop through all the accounts?

plugins/accounts/plugin.go
lmoe: How about `users`? I'm not so sure about apiAccounts as they are also used in the dashboard and it's not really an API. Depending on the usage the users can maybe be used elsewhere as well.

packages/authentication/basic_auth.go
lmoe: * Will do!
* With some changes for sure. Right now we have a plain list of account objects inside the config, using map is possible too, but seems uncommon to do so(?). At least I've never encountered it that way. I don't have a strong opinion on either. It would definitely make sure that no account gets created twice.

jorgemmsilva: having logins in plain text in the config file doesn't sound like a permanent solution anyway.

I'm actually thinking if it could make sense to just use a cryptographic signature as a means to "login" 🤔 - (the server would send a payload that gets signed by the user to confirm the identity). Maybe too cumbersome for authenticating in a website

lmoe: No definitely not. Personally I would like to see a very light user management and maybe a storage of the users inside waspdb with hashed passwords (in the future). We can still add hashes as passwords in the config file, but I think that makes configuration harder to for the end users as they would need to hash those passwords by themselves, or we write a little tool for that?

Maybe that cryptographic signature could be used in combination with some wallet?

packages/authentication/ip_whitelist.go
```diff
@@ -0,0 +1,52 @@
...
+		if ip.IsLoopback() {
+			return true
+		}
+		for _, whitelistedIP := range whitelist {
+			if ip.Equal(whitelistedIP) {
+				return true
+			}
+		}
+		return false
+	}
```
jorgemmsilva: nitpick: to improve readability this could be extracted and defined as a function just a single time:
```go
	func isAllowed(ip net.IP, whitelist []net.IP) bool {
		if ip.IsLoopback() {
			return true
		}
		for _, whitelistedIP := range whitelist {
			if ip.Equal(whitelistedIP) {
				return true
			}
		}
		return false
	}
```

packages/authentication/jwt_auth.go
```diff
@@ -0,0 +1,224 @@
...
+			return false
+		}
+
+		if !allow(claims) {
+			return false
+		}
+
+		return true
+	}
+	return false
```
jorgemmsilva:
```suggestion
	if err != nil || !t.Valid {
		return false
	}
	claims, ok := t.Claims.(*WaspClaims)

	if !ok || !claims.VerifyAudience(j.nodeID, true) {
		return false
	}

	if !allow(claims) {
		return false
	}

	return true
```

packages/authentication/jwt_auth.go
```diff
@@ -0,0 +1,224 @@
...
+
+	return jwtAuth, jwtAuthSkipper, jwtAuthAllow, nil
+}
+
+func AddJWTAuth(webAPI WebAPI, config JWTAuthConfiguration, privateKey []byte, userMap map[string]*users.UserData, claimValidator ClaimValidator) *JWTAuth {
+	durationHours := config.DurationHours
+
+	// If durationHours is 0, we set 24h as the default durationHours.
+	if durationHours <= 0 {
+		durationHours = 24
```
jorgemmsilva: nit: should be a const `defaultJwtDuration`

packages/users/users.go
```diff
@@ -0,0 +1,23 @@
+package users
+
+type UserData struct {
+	Username string
+	Password string
+	Claims   []string
```
dessaya, line 6: Why not `Claims *WaspClaims`? So that decoding of the claims happens only once when creating the `UserData`.

packages/authentication/jwt_handler.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if contentType == echo.MIMEApplicationForm {
+		cookie := http.Cookie{
+			Name:     "jwt",
+			Value:    token,
+			HttpOnly: true, // JWT Token will be stored in a http only cookie, this is important to mitigate XSS/XSRF attacks
+			Expires:  time.Now().Add(a.Jwt.durationHours),
+			Path:     "/",
+			SameSite: http.SameSiteStrictMode,
+		}
```
jorgemmsilva, line 99:
```suggestion
			Secure:     true,
		}
```
we could add `Secure` property to the cookie also, but it might be problematic if we don't have https by default, tho :thinking:

packages/authentication/strategy.go
```diff
@@ -0,0 +1,104 @@
...
+	IPWhiteList []string `koanf:"whitelist"`
+}
+
+type WebAPI interface {
+	GET(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
+	POST(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
+	Use(middleware ...echo.MiddlewareFunc)
+}
+
+var log *logger.Logger
```
dessaya: Why not receive the logger as a parameter instead?
Typically we initialize all loggers in plugins and pass them as parameters tp packages.

dessaya (approved):

packages/users/users.go
lmoe: Because that doesn't work in this place. I need a struct to deserialize the config into, which in our case is just a list of strings which act as claims. This user plugin is a first iteration for now anyway as I don't like storing the passwords inside a config.

packages/authentication/jwt_handler.go
lmoe: I think that would rather happen with a http/s proxy infront. Not sure if a https echo server is helpful in our case..

packages/authentication/strategy.go
lmoe: Authentication is just a lib without a plugin initializer. Should I add it?

/edit: Forget about it, I ditch the log from there alltogether, its just required in the ipwhitelist handler.

lmoe:
I have set the default auth config back to the original config to not break something by accident. (API -> Whitelist, Dashboard -> Basic auth)
Cluster tests pass. 

Will write additional documentation and introduce the experimental JWT auth method in the next PR.

PR #845, opened 2022-02-23
fix(evm): jsonrpc tests + refactor vmerrors

dessaya:
* Fixed EVM json rpc tests, and added to `stardust-test` script
* Some changes to VMErrors

### DIFF
--- a/contracts/native/evm/evminternal/management.go
+++ b/contracts/native/evm/evminternal/management.go
@@ -17,7 +17,6 @@ import (
 	"github.com/iotaledger/wasp/packages/kv/dict"
 	"github.com/iotaledger/wasp/packages/kv/kvdecoder"
 	"github.com/iotaledger/wasp/packages/kv/subrealm"
-	"github.com/iotaledger/wasp/packages/util"
 	"github.com/iotaledger/wasp/packages/vm/gas"
 )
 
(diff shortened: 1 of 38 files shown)


contracts/native/evm/evminternal/management.go
```diff
@@ -150,25 +149,21 @@ func ApplyTransaction(ctx iscp.Sandbox, apply func(tx *types.Transaction, blockT
 
 	gasRatio := codec.MustDecodeRatio32(ctx.State().MustGet(keyGasRatio), evm.DefaultGasRatio)
 
-	// ignore the evm gas budget set in the evm tx, use the remaining ISC gas budget instead
-	gasBudget := ISCGasBudgetToEVM(ctx.Gas().Budget(), gasRatio)
+	gasBudget := evm.ISCGasBudgetToEVM(ctx.Gas().Budget(), gasRatio)
+	if !ignoreEVMGasBudget && gasBudget < tx.Gas() {
```
jorgemmsilva, line 153: when is `ignoreEVMGasBudget` used? 

edit: I see its `true` on `emvlight/impl.go`, why do we need that?

jorgemmsilva (approved):

contracts/native/evm/evminternal/management.go
dessaya: I don't know if it's the right choice, but I figured that specifying a gas budget in the EVM tx and also **another** gas budget in the ISC request is too cumbersome, so in the evmlight version I just ignore the gas limit in EVM: only the ISC gas budget is taken into account. Because of how evmchain works it is impossible to do the same.

Maybe it will bring more issues than it solves... at least it can be easily disabled if needed.

jorgemmsilva: oh, nice - that makes sense! :+1:

PR #850, opened 2022-03-02
Stardust trie julius

Juliusan:
# Description of change

Mempool and state manager compiles after these changes.

## Type of change

Using `stardust-test` script.


packages/chain/statemgr/action.go
```diff
@@ -144,7 +150,7 @@ func (sm *stateManager) addBlockAndCheckStateOutput(block state.Block, nextState
 	if isBlockNew {
 		if sm.stateOutput != nil {
 			sm.log.Debugf("addBlockAndCheckStateOutput: checking if block index %v (local %v, nextStateHash %v, approvingOutputID %v, already approved %v) is approved by current stateOutput",
-				block.BlockIndex(), candidate.isLocal(), candidate.getNextStateCommitment().String(), iscp.OID(candidate.getApprovingOutputID()), candidate.isApproved())
+				block.BlockIndex(), candidate.isLocal(), candidate.getNextStateCommitmentString(), iscp.OID(candidate.getApprovingOutputID()), candidate.isApproved())
```
lunfardo314: if you use %s, the fmt. formatting looks for method `String()string` automatically

packages/chain/statemgr/candidate_block.go
```diff
@@ -15,19 +15,19 @@ type candidateBlock struct {
 	local               bool
 	votes               int
 	approved            bool
-	nextStateCommitment hashing.HashValue
+	nextStateCommitment trie.CommitmentBase
```
lunfardo314: must be `trie.VCommitment` it cant be `TCommitment` anyway

packages/chain/statemgr/candidate_block.go
```diff
@@ -15,19 +15,19 @@ type candidateBlock struct {
...
 	nextState           state.VirtualStateAccess
 }
 
 func newCandidateBlock(block state.Block, nextStateIfProvided state.VirtualStateAccess) *candidateBlock {
 	var local bool
-	var stateCommitment hashing.HashValue
+	var stateCommitment trie.CommitmentBase
 	if nextStateIfProvided == nil {
```
lunfardo314, line 25: see above

packages/chain/statemgr/eventproc.go
```diff
@@ -108,12 +108,13 @@ func (sm *stateManager) handleStateMsg(msg *messages.StateMsg) {
 		"chainOutput", iscp.OID(msg.ChainOutput.ID()),
 	)
 	sm.stateManagerMetrics.LastSeenStateIndex(msg.ChainOutput.GetStateIndex())
-	stateHash, err := msg.ChainOutput.GetStateCommitment()
+	stateL1Commitment, err := state.L1CommitmentFromAliasOutputWithID(msg.ChainOutput)
 	if err != nil {
```
lunfardo314, line 112: why this function with long name?
I'd expect something like  `state.L1CommitmentFromAliasOutput(msg.ChainOutputWithID.Output)`
Otherwise it is confused with the concept what is output from iota.go.
Also in other places

packages/chain/statemgr/eventproc.go
```diff
@@ -108,12 +108,13 @@ func (sm *stateManager) handleStateMsg(msg *messages.StateMsg) {
...
 		sm.log.Errorf("EventStateMsg ignored: failed to parse state hash: %v", err)
 		return
 	}
-	sm.log.Debugf("EventStateMsg state hash is %v", stateHash.String())
+	stateCommitment := stateL1Commitment.Commitment
+	sm.log.Debugf("EventStateMsg state hash is %v", stateCommitment.String())
 	if sm.stateOutputReceived(msg.ChainOutput, msg.Timestamp) {
```
lunfardo314, line 118: use `%s`, no need for `.String()`

packages/chain/statemgr/statemgr.go
```diff
@@ -159,7 +160,7 @@ func (sm *stateManager) initLoadState() {
 		sm.solidState = solidState
 		sm.chain.GlobalStateSync().SetSolidIndex(solidState.BlockIndex())
 		sm.log.Infof("SOLID STATE has been loaded. Block index: #%d, State hash: %s",
```
lunfardo314: rename `state hash` -> `state commitment`. 
No need for `String()` is `%s` is used

packages/chain/statemgr/syncing_block.go
```diff
@@ -154,7 +154,7 @@ func (syncsT *syncingBlocks) approveBlockCandidates(output *iscp.AliasOutputWith
 		for blockHash, candidate := range sync.blockCandidates {
 			alreadyApproved := candidate.isApproved()
 			syncsT.log.Debugf("approveBlockCandidates: checking candidate %v: local %v, nextStateHash %v, approvingOutputID %v, already approved %v",
-				blockHash.String(), candidate.isLocal(), candidate.getNextStateCommitment().String(), iscp.OID(candidate.getApprovingOutputID()), alreadyApproved)
```
lunfardo314, line 157: use '%s'

packages/state/l1commitment.go
```diff
@@ -24,6 +25,14 @@ func L1CommitmentFromBytes(data []byte) (L1Commitment, error) {
 	return ret, nil
 }
 
+func L1CommitmentFromAliasOutputWithID(output *iscp.AliasOutputWithID) (*L1Commitment, error) {
+	l1c, err := L1CommitmentFromBytes(output.GetStateMetadata())
```
lunfardo314: in `iscp.AliasOutputWithID` make `Output` open field in struct and this function is not needed

lunfardo314 (changes requested):

packages/chain/statemgr/action.go
Juliusan: The problem is that I was getting null pointers while using `String` method here.

packages/state/l1commitment.go
Juliusan: Making `Output` public defeats the purpose of the wrapper. Then Output can be assigned to different object without assigning output ID.

packages/chain/statemgr/action.go
Juliusan: I'll check if "%s" doesn't throw null pointer

PR #852, opened 2022-03-02
Allow "0x<hname>" for function name in web api when only hname is known.

BugFreeSoftware:


packages/webapi/state/callview.go
```diff
@@ -65,6 +66,15 @@ func (s *callViewService) handleCallView(c echo.Context) error {
 	}
 
 	fname := c.Param("fname")
```
dessaya, line 68: I think it would be cleaner to accept a `hname` param instead, and the client would have to provide exactly one of `fname` or `hname`. That way you would still be able to call by function name if it actually begins with `"0x"`.

BugFreeSoftware: Function names and contract names cannot start with a digit anyway. They are being used as identifier in code, and most languages do not accept identifiers that start with a digit. Which is why I chose this simple solution instead of providing a different api url.

dessaya: That limitation applies only to wasm contracts? Because for native contract names is possible to begin with "0x".

I'm just being paranoid here because once the API is used by many different clients it becomes very difficult to change.

PR #856, opened 2022-03-04
feat: NFTs

jorgemmsilva:
Initial NFT implementation


packages/iscp/sandbox_interface.go
```diff
@@ -136,6 +136,10 @@ type RequestParameters struct {
 	Metadata *SendMetadata
 	// SendOptions includes options of the output, such as time lock or expiry parameters
 	Options SendOptions
+	// NFTID (optional)
+	NFTID *iotago.NFTID
+	// when sending a request via smartcontract (ctx.Send) the NFT data will be automatically filled if NFTID is present
+	NFT *NFT
```
dessaya, line 142: This looks out of place. I mean, a public member of the struct that is not meant to be filled. No one will read the documentation :rofl: .

Maybe make both `NFTID` and `NFT` private? And add a method `WithNFT(NFTID id)` that fills in the NFT data, and `GetNFT() *NFT` for querying.

Or maybe split this object in two? Like `RequestParameters` and `ResolvedRequestParameters`.

packages/transaction/nfttransaction.go
```diff
@@ -0,0 +1,46 @@
+package transaction
+
+import (
+	iotago "github.com/iotaledger/iota.go/v3"
+	"github.com/iotaledger/wasp/packages/cryptolib"
+	"github.com/iotaledger/wasp/packages/parameters"
+)
+
+type MintNftTransactionParams struct {
```
dessaya, line 9: Nit: `MintNFTTransactionParams`

packages/vm/core/accounts/internal.go
```diff
@@ -373,13 +472,85 @@ func calcL2TotalAssets(state kv.KVStoreReader) *iscp.Assets {
...
+				panic(fmt.Sprintf("inconsistency: NFT %s is owned by more than 1 account", nft.String()))
+			}
+			ret[nft] = true
+		}
+		return true
+	})
+	return ret
+}
+
+func NFTmapEqual(a, b map[iotago.NFTID]bool) bool {
```
dessaya, line 529: Nit: `NFTMapEqual`

dessaya (approved):

packages/iscp/sandbox_interface.go
jorgemmsilva: you're 100% right 😄 
I also felt that this wasn't right, and even experimented creating a `RequestParametersWithNFTData`, but I didn't like that either, so I left it as is just to make it work for now.
(btw, even if the user fills it, it won't have any bad consequenses, it just gets overwritten, but still it will cause confusion for sure lol)

the suggestion to add a private field and getter/setters seems like a nice idea, will do that 👍

lunfardo314: I am not a fan of this design. To make a request, we just have to specify existing NFT output, attach tokens and parameters to it and specify in allowance if target is allowed to take NFT or not. I'd prefer to have a special call for that, like `SendAsNFT` with some additional parameters and same call data struct. Now it looks like chimera

packages/vm/core/accounts/internal.go
```diff
@@ -232,8 +240,86 @@ func debitFromAccount(account *collections.Map, assets *iscp.Assets) bool {
...
+	if err != nil || len(bytes) == 0 {
+		return false
+	}
+	err = account.DelAt(id[:])
+	return err == nil
+}
+
 // MoveBetweenAccounts moves assets between on-chain accounts. Returns if it was a success (= enough funds in the source)
-func MoveBetweenAccounts(state kv.KVStore, fromAgentID, toAgentID *iscp.AgentID, transfer *iscp.Assets) bool {
+func MoveBetweenAccounts(state kv.KVStore, fromAgentID, toAgentID *iscp.AgentID, transfer *iscp.Allowance) bool {
```
lunfardo314, line 322: this semantically not correct. Allowance is used as not allowance. I want to avoid that.
The correct way would be have separate functions to move fungible tokens and NFT as separate functions:
- `MoveBetweenAccounts` should stay without  modifications, probably renamed to `MoveBeteenAccountFungibleTokens`
- NFT should be moved by `MoveBetweenAccountsNFTs` with parameter of list of NFTs

lunfardo314 (changes requested):

packages/iscp/sandbox_interface.go
jorgemmsilva: >To make a request, we just have to specify existing NFT output

why shouldn't the NFTID be enough? no need to "specify the output" IMO

>and specify in allowance if target is allowed to take NFT or not

I opted to leave it this way because it allows to specify many NFTs, instead of a single one.

> I'd prefer to have a special call for that, like SendAsNFT with some additional parameters and same call data struct. Now it looks like chimera

We can change it to have `Send` and `SendAsNFT`, but we should discuss it - its more confusing imo to have 2 Send functions. What is the problem with the following? (I don't se a "chimera" :rofl: )

```golang
	ctx.Send(iscp.RequestParameters{
		...
		NFTID: nftID,
	})
```

packages/vm/core/accounts/internal.go
jorgemmsilva: >this semantically not correct. Allowance is used as not allowance.

I don't understand this statement. What do you mean by this?

also what is the advantage of having `MoveBeteenAccountFungibleTokens` and `MoveBetweenAccountsNFTs`? :confused:

jorgemmsilva:
closed in favor of https://github.com/iotaledger/wasp/pull/859

PR #862, opened 2022-03-08
NodeConn and Hornet PrivTangle helper

kape1395:
PrivTangle:
  - A helper for starting a hornet cluster. It is to be used in the cluster tool and for the tests.
  - Some stability problems still persists, they are worked around with delays for now.

NodeConn:
  - This implementation provides the needed functions, but some internal features are missing.
  - Publishing a TX, receiving unspent outputs and TX inclusion state updates works.
  - TX reattachment, promotion, pipelining, pow coordination is not done yet.


client/chainclient/chainclient.go
```diff
@@ -1,27 +1,31 @@
...
 
 // Client allows to interact with a specific chain in the node, for example to send on-ledger or off-ledger requests
 type Client struct {
-	Layer1Client interface{} // TODO implement
+	Layer1Client nodeconn.L1Client
 	WaspClient   *client.WaspClient
 	ChainID      *iscp.ChainID
 	KeyPair      *cryptolib.KeyPair
-	nonces       map[[cryptolib.PublicKeySize]byte]uint64
+	nonces       map[iotago.Address]uint64
```
dessaya: This should probably be `map[string]uint64`, and use `address.Key()`. Otherwise you can have two identical addresses as separate keys.

PR #864, opened 2022-03-08
Feat: 256+ Merkle trie with proofs of inclusion 

lunfardo314:
# Description of change

- implemented `trie` package, a generic trie with the branching factor of 256 and variable length keys
- implemented `trie_merkle` package, a Merkle trie with hash commitments
- implemented Proofs Of Inclusion (and absence) of arbitrary key/value pairs in the chain's state based on Merkle proofs
- minor adjustments and optimizations in `state` and `kv` package

* [Commitments to the state](https://hackmd.io/@Evaldas/ryQFvhXfF)
* [256+ trie definition](https://hackmd.io/@Evaldas/H13YFOVGt)

## Type of change

Covered with multiple tests and benchmarks


packages/kv/collections/array32.go
```diff
@@ -35,16 +35,16 @@ func NewArray32ReadOnly(kvReader kv.KVStoreReader, name string) *ImmutableArray3
 
 const array32ElemKeyCode = byte('#')
 
-func array32SizeKey(name string) kv.Key {
-	return kv.Key(name)
+func Array32SizeKey(name string) []byte {
+	return []byte(name)
```
dessaya: Why change to `[]byte`? Same for `Array32ElemKey`, `MapSizeKey`, etc. If you need `[]byte` you can always convert after calling, but most callers will expect `kv.Key`.

packages/kv/diff.go
```diff
@@ -0,0 +1,149 @@
+package kv
+
+import (
+	"bytes"
+	"fmt"
+	"sort"
+)
+
+// used mostly for testing
```
dessaya, line 9: Is this being used? I don't find any callers to functions in this file.

If it's used only for testing it should probably go outside the `kv` package, maybe in something like `kvtest` or `kvdiff` or `kvstats`.

packages/kv/kv.go
```diff
@@ -3,12 +3,13 @@ package kv
...
 	"fmt"
-
 	"github.com/iotaledger/wasp/packages/util"
 	"golang.org/x/xerrors"
+	"io"
+	"os"
 )
 
-// Since map cannot have []byte as key, to avoid unnecessary conversions
+// Key since map cannot have []byte as key, to avoid unnecessary conversions
```
dessaya: Is this change intentional? The sentence now does not make sense :stuck_out_tongue: 

```suggestion
// Key represents a key in the KVStore.
// Since map cannot have []byte as key, to avoid unnecessary conversions
```

packages/kv/kv.go
```diff
@@ -155,3 +156,65 @@ func ByteSize(s KVStoreReader) int {
...
+
+func DumpToFile(r KVStoreReader, fname string) (int, error) {
+	file, err := os.Create(fname)
+	if err != nil {
+		return 0, err
+	}
+	defer file.Close()
+
+	var bytesTotal int
+	_ = r.Iterate("", func(k Key, v []byte) bool {
```
dessaya: Either capture the error or use `MustIterate`?

packages/kv/trie/node.go
```diff
@@ -0,0 +1,188 @@
...
+		n.Terminal = nil
+	}
+	if smallFlags&hasChildrenFlag != 0 {
+		var flags [32]byte
+		if _, err := r.Read(flags[:]); err != nil {
+			return err
+		}
+		for i := 0; i < 256; i++ {
+			ib := uint8(i)
+			if flags[i/8]&(0x1<<(i%8)) != 0 {
```
dessaya: :exploding_head:

packages/kv/trie/nodestore.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return tr.Model().CommitToNode(n)
+}
+
+func EqualCommitments(c1, c2 CommitmentBase) bool {
+	if c1 == c2 {
+		return true
+	}
+	c1Nil := c1 == nil || (reflect.ValueOf(c1).Kind() == reflect.Ptr && reflect.ValueOf(c1).IsNil())
+	c2Nil := c2 == nil || (reflect.ValueOf(c2).Kind() == reflect.Ptr && reflect.ValueOf(c2).IsNil())
```
dessaya, line 70: Why is this needed?

packages/state/block_test.go
```diff
@@ -63,28 +65,26 @@ func TestBatches(t *testing.T) {
...
-
-	require.True(t, b.Timestamp().IsZero())
-	require.True(t, b1.Timestamp().IsZero())
-	require.True(t, b2.Timestamp().IsZero())
-
-	require.EqualValues(t, hashing.NilHash, b.PreviousStateHash())
-	require.EqualValues(t, hashing.NilHash, b1.PreviousStateHash())
-	require.EqualValues(t, hashing.NilHash, b2.PreviousStateHash())
-}
+//func TestOriginBlock(t *testing.T) {
```
dessaya: Why is this test commented out?

Suggest using `t.Skip("skipping because this and that")` instead of commenting out. (Or if it doesn't compile just comment out the offending block and skip), so that when running tests we can know how many tests are actually skipped.

packages/state/types.go
```diff
@@ -14,13 +14,16 @@ import (
 type VirtualStateAccess interface {
 	BlockIndex() uint32
 	Timestamp() time.Time
-	PreviousStateHash() hashing.HashValue
-	StateCommitment() hashing.HashValue
+	TrieAccess() trie.NodeStore
```
dessaya, line 17: I don't understand what "Access" means in "VirtualStateAccess", "TrieAccess", etc. Why not call this just `Trie()`?

dessaya (approved):

packages/kv/diff.go
lunfardo314: will delete a bit later. It is need for deep debugging, may be usefull in the future

packages/kv/collections/array32.go
lunfardo314: this is a confusion which one to use where and what is the main use case. Just made it uniform after exposed it as an exported function (before it was internal). Intended use case is composing of keys in proofs of inclusion, we need bytes there

packages/kv/kv.go
lunfardo314: don't remember, what's wrong here?

packages/kv/collections/array32.go
dessaya: I still think that for consistency all functions that deal with keys must use `kv.Key`. If you need bytes, you can do `[]byte(Array32SizeKey("..."))`

packages/kv/kv.go
dessaya: "Key since map cannot have []byte as key, [...]"
The problem is that the sentence does not make sense like this. Godoc comments must be simple English sentences. Starting with the same identifier is a recommendation, but not a requirement. Especially not if it breaks the sentence.

PR #875, opened 2022-03-14
Consensus refactored for stardust

Juliusan:
# Description of change

Consensus code compiles and all the unit tests pass. However, `TestCruelWorld` test occasionally fails, but it appears that this is due to problems in ACS. This problem will be resolved later.

## Type of change

Using `stardust-test` script


jorgemmsilva (approved):
lgtm, I see some TODO/commented code, but assume that is to be tackled later

dessaya (approved):

PR #882, opened 2022-03-17
Implement a text key/value store

ivange94:
# Description of change

Make registryStore configurable and implemented a text key/value store that can be injected into the registry.

## Type of change

In the config file, add the config

```
"registry": {
    "useText": true
}
```

## Change checklist


packages/database/textdb/textkv_impl.go
```diff
@@ -0,0 +1,336 @@
...
+func GetMarshaller() Marshaller {
+	regFile := parameters.GetString(parameters.DatabaseTextFilename)
+	if filepath.Ext(regFile) == "yaml" {
+		return NewYAMLMarshaller()
+	}
+	return NewJSONMarshaller()
+}
+
+func NewTextKV(log *logger.Logger, filename string) kvstore.KVStore {
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_APPEND, 0o666)
```
dessaya: 666 gives write permissions to "others"; is that intended?

packages/database/textdb/textkv_impl.go
```diff
@@ -0,0 +1,336 @@
...
+	defer f.Close()
+	if err != nil {
+		// log error
```
dessaya: Please either log the error or just panic :)

packages/database/textdb/textkv_impl.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	if fd.Size() == 0 {
+		os.WriteFile(f.Name(), []byte("{}"), 0o666)
+	}
+	return &textKV{filename: f.Name(), log: log, Marshaller: GetMarshaller()}
+}
+
+// WithRealm is a factory method for using the same underlying storage with a different realm.
+func (s *textKV) WithRealm(realm kvstore.Realm) kvstore.KVStore {
+	// return &textKV{s.filename, s.log, realm}
```
dessaya: This commented code left intentionally?

packages/database/textdb/textkv_impl.go
```diff
@@ -0,0 +1,336 @@
...
+
+func GetMarshaller() Marshaller {
+	regFile := parameters.GetString(parameters.DatabaseTextFilename)
+	if filepath.Ext(regFile) == "yaml" {
+		return NewYAMLMarshaller()
+	}
+	return NewJSONMarshaller()
+}
+
+func NewTextKV(log *logger.Logger, filename string) kvstore.KVStore {
```
dessaya: Maybe document this function? What is a "TextKV"? What is its purpose?

Just sayng because I see that all other functions are documented, and this one is probably the one that really needs it :sweat_smile:

packages/parameters/parameters.go
```diff
@@ -25,8 +25,10 @@ const (
 	LoggerOutputPaths       = "logger.outputPaths"
 	LoggerDisableEvents     = "logger.disableEvents"
 
-	DatabaseDir      = "database.directory"
-	DatabaseInMemory = "database.inMemory"
+	DatabaseDir          = "database.directory"
+	DatabaseInMemory     = "database.inMemory"
+	DatabaseUseText      = "database.useText"
+	DatabaseTextFilename = "database.TextFilename"
```
dessaya: Nit: all other flags are camelCase and this one is TitleCase

packages/database/textdb/textdb.go
```diff
@@ -0,0 +1,53 @@
...
+func (db *textDB) RequiresGC() bool {
+	return false
+}
+
+func (db *textDB) GC() error {
+	return nil
+}
+
+func Base58Text(in []byte, m Marshaller) []byte {
+	hexStr := hex.EncodeToString(in)
```
dessaya: hex or base58? :rofl:

packages/registry/chainrecord.go
```diff
@@ -18,6 +19,15 @@ type ChainRecord struct {
 	Active  bool
 }
 
+type ChainRecordText struct {
+	ChainID string `json:"chainid" yaml:"chainid"`
+	Active  bool   `json:"active" yaml:"active"`
+}
```
dessaya: If `ChainID` implements `json.Marshaler` and `json.Unmarshaler` (same for yaml), then this struct becomes unnecessary, since `ChainRecord` can be automatically marshalled. Same for `DKShare` and any other type that needs to be marshalled to json/yaml.

ivange94: This struct is used here to create a flat object. Using the default marshaler will cause the resulting json to be very nested which defeats the purpose of readability.

packages/database/textdb/textdb.go
ivange94: Yes that's a bug. I was initially using hex then refactored everywhere to use base58 but forgot to change the implementation 😄. But that's one of the reasons the PR is still in a draft format. I just wanted to shed some light on the challenge of the current approach before cleaning up the code.

packages/database/textdb/textkv_impl.go
ivange94: Yes, that will not be in the final PR.

packages/registry/chainrecord.go
dessaya: What I'm proposing is not using the default marshaler, but implement `json.Marshaler` and `json.Unmarshaler` for `ChainID` instead.

Something like:

```
func (c *ChainID) MarshalJSON() ([]byte, error) {
    // return base58
}

func (c *ChainID) UnmarshalJSON([]byte) error {
    // unmarshal from base58
}
```

Same for Yaml.

This not only gives you a flatter json representation for `ChainRecord`, but also is more reusable since now ChainID can be marshaled to json in any other context.

ivange94: Oh ok. Yeah, that's much cleaner.

packages/database/textdb/textkv_impl.go
ivange94: We need to be able to manually edit the config reason why it's so permissive. Do you think we should restrict it?

packages/database/textdb/textkv.go
```diff
@@ -0,0 +1,334 @@
...
+
+	"github.com/iotaledger/hive.go/byteutils"
+	"github.com/iotaledger/hive.go/kvstore"
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/hive.go/types"
+	"github.com/iotaledger/wasp/packages/parameters"
+	"github.com/mr-tron/base58"
+)
+
+const StorePerm = 0o666
```
jorgemmsilva:
```suggestion
const storePerm = 0o666
```
should be private

packages/database/textdb/textkv_impl.go
dessaya: But why not 664? (user = rw, group = rw, others = r). It's very rare to give write permissions to "others".

packages/database/textdb/yaml_marshaller.go
```diff
@@ -0,0 +1,15 @@
+package textdb
+
+import "gopkg.in/yaml.v2"
+
+type yamlMarshaller struct{}
```
jorgemmsilva, line 5: why the empty structs?

packages/registry/registry_impl.go
```diff
@@ -18,17 +19,28 @@ import (
 	"github.com/iotaledger/wasp/packages/kv/codec"
 	"github.com/iotaledger/wasp/packages/peering"
 	"github.com/iotaledger/wasp/packages/tcrypto"
+	"github.com/mr-tron/base58"
 )
 
-// region Registry /////////////////////////////////////////////////////////
-
 // Impl is just a placeholder to implement all interfaces needed by different components.
```
jorgemmsilva:
```suggestion
// region Registry /////////////////////////////////////////////////////////

// Impl is just a placeholder to implement all interfaces needed by different components.
```

packages/database/textdb/yaml_marshaller.go
ivange94: Just needed an implementation of the marshaler interface. The kvstore still works with json and yaml even though the registry only sends json data.

packages/database/textdb/textkv.go
```diff
@@ -0,0 +1,291 @@
...
+
+	err := s.inMemoryStore.DeletePrefix(byteutils.ConcatBytes(s.realm, prefix))
+	if err != nil {
+		return err
+	}
+	return s.flush()
+}
+
+func (s *textKV) flush() error {
+	var err error
```
jorgemmsilva, line 196: we probably also want to lock on flush
```suggestion
	s.Lock()
	defer s.Unlock()
	var err error
```

ivange94: `flush` is only called within code that already has a handle to the lock. So calling s.Lock() inside flush will may block as the parent code still has not unlocked. Or did I think of this wrongly?

plugins/users/plugin.go
```diff
@@ -45,6 +45,8 @@ func loadUsersFromConfiguration() error {
 		// During the transition phase, create a default user when the config is empty.
 		// This keeps the old authentication working.
 
+		// TODO: just a workaround. Actual fix needed.
+		userMap = make(map[string]*users.UserData)
```
jorgemmsilva: @lmoe :point_up:

jorgemmsilva (approved):
LGTM :heavy_check_mark:  just some minor comments

packages/database/textdb/textkv.go
jorgemmsilva: oh I see, I thought it was a public method, now that I see the public Flush just returns nil. nevermind :+1:

dessaya (approved):

PR #886, opened 2022-03-20
feat: snapshot package and tools

lunfardo314:
- `snapshot` package with related functions
- `snapshot` CLI tool to work with shapshot files
- TODO not finished snapshot validation with the state proofs because it requires WEB APIs for proofs


packages/database/dbmanager/db.go
```diff
@@ -63,8 +63,9 @@ type rocksDB struct {
 
 // NewDB returns a new persisting DB object.
 func NewDB(dirname string) (DB, error) {
-	opt := rocksdb.UseCompression(true)
-	db, err := rocksdb.CreateDB(dirname, opt)
+	//opt := rocksdb.UseCompression(true)
+	//db, err := rocksdb.CreateDB(dirname, opt)
```
dessaya, line 67: This left commented out intentionally?

packages/kv/kvstream.go
```diff
@@ -0,0 +1,114 @@
...
+
+// binStreamWriter writes stream of k/v pairs in binary format.
+// Keys encoding is 'bytes16' and values is 'bytes32'
+type binStreamWriter struct {
+	w         io.Writer
+	kvCount   int
+	byteCount int
+}
+
+func NewBinaryStreamWriter(w io.Writer) binStreamWriter {
```
dessaya: If the function is public, the returned value should also be public: either return `StreamWriter` or make the struct public.

packages/kv/kvstream.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/iotaledger/wasp/packages/util"
+	"io"
+	"os"
+)
+
+// Interfaces for writing/reading persistent streams of key/values
+
+// StreamWriter represents an interface specific to write a sequence of key/value pairs
+type StreamWriter interface {
+	Write(key, value []byte) error
```
dessaya, line 14: Why not have `key Key` instead of `key []byte`? For consistency.

packages/kv/testutil.go
```diff
@@ -0,0 +1,134 @@
+package kv
```
dessaya: Nit: if this is only used for testing I would move it to a `kvtest` package, so that it is only compiled if explicitly imported.

packages/state/oreader.go
```diff
@@ -3,56 +3,63 @@ package state
...
-	db          kvstore.KVStore
+// optimisticStateReaderImpl state reader reads the chain state from db and validates it
+type optimisticStateReaderImpl struct {
 	stateReader *optimism.OptimisticKVStoreReader
 	trie        trie.NodeStore
 }
 
 // NewOptimisticStateReader creates new optimistic read-only access to the database. It contains own read baseline
-func NewOptimisticStateReader(db kvstore.KVStore, glb coreutil.ChainStateSync) *OptimisticStateReaderImpl {
+func NewOptimisticStateReader(db kvstore.KVStore, glb coreutil.ChainStateSync) *optimisticStateReaderImpl {
```
dessaya, line 21: Same here: public function returning private struct is not kosher :rofl:

packages/state/state.go
```diff
@@ -191,8 +201,8 @@ func (vs *virtualStateAccess) Commit() {
 	vs.kvs.Mutations().ResetModified()
 }
 
-// StateCommitment returns the hash of the state, calculated as a hashing of the previous (committed) state hash and the block hash.
-func (vs *virtualStateAccess) TrieAccess() trie.NodeStore {
+// TrieAccess returns the hash of the state, calculated as a hashing of the previous (committed) state hash and the block hash.
```
dessaya:
```suggestion
// TrieNodeStore returns the hash of the state, calculated as a hashing of the previous (committed) state hash and the block hash.
```

packages/vm/core/accounts/impl.go
```diff
@@ -41,6 +41,9 @@ func initialize(ctx iscp.Sandbox) dict.Dict {
 	ctx.Requiref(err == nil && iotasOnAnchor >= dustDepositAssumptions.AnchorOutput,
 		"accounts.initialize.fail: %v", ErrDustDepositAssumptionsWrong)
 	ctx.State().Set(kv.Key(stateVarMinimumDustDepositAssumptionsBin), dustAssumptionsBin)
+	// storing hname as a terminal value of the contract's state root.
+	// This way we will be able to retrieve commitment to the contract's state
+	ctx.State().Set("", ctx.Contract().Bytes())
```
dessaya, line 46: I see that this is duplicated on all core contracts?
Can't this be done automatically by the root contract or something? It looks error prone: we can easily forget to add this when adding a new core contract.

tools/snapshot/i.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+go install -tags rocksdb,builtin_static
```
dessaya, line 3: I suggest defining your own aliases in `~/.bashrc` instead. Like:

`alias i='go install -tags rocksdb,builtin_static'`

dessaya (approved):

tools/snapshot/i.sh
jorgemmsilva: this is literally the behaviour of `make install` :smile:

packages/database/dbmanager/db.go
jorgemmsilva: it seems we are going to just disable compression for the foreseeable future:
https://iotafoundation.slack.com/archives/G01F91K2CRX/p1647525793227049?thread_ts=1647508882.523679&cid=G01F91K2CRX
:shrug:

tools/snapshot/readme.md
```diff
@@ -0,0 +1,118 @@
...
+
+The snapshot-related library code is located in `packages/snapshot`. The `tools/snapshot` directory contains code of the `snapshot` executable.
+
+## What is *snapshot*?
+*Snaphot* is the state of the chain stored separately. Node can start the chain from snapshot, i.e. it assumes current state of the chain is equal to the one provided and then syncronizes the state with other nodes.
+
+## How we create a snapshot of the chain's state?
+
+There are two possibilities:
+- by backing up the database
```
jorgemmsilva: I don't think `backing up` is correct in this context

```suggestion
- by creating a backup of the database
```

tools/snapshot/readme.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Database backup
+
+All the data of the chain is stored in the `rocksdb` key/value database. The name of the directory of the database is hex-encoded `chainID` of the chain. It contains full state of the chain with trie at some state index.
+
+Upon activation of the chain in the node, the `state manager` automatically loads the state in the database and synchronizes it with the chain.
+
+The directory can be backed up and restored with usual backup tools.
+
+Note, that `state manager` updates the database when syncing the state and when state transition occurs, i.e. at least once per 10 sec. It means, for the backed up database to be consistent, all chain activity must be stopped.
```
jorgemmsilva:
```suggestion
Note: the `state manager` updates the database when syncing the state and when state transition occurs, i.e. at least once per 10 sec. It means, for a database backup to be consistent, all chain activity must be stopped.
```

tools/snapshot/readme.md
```diff
@@ -0,0 +1,118 @@
...
+
+For example for chainID = `210fc7bb818639ac48a4c6afa2f1581a8b9525e2` snapshot file for the state index `314` will be `210fc7bb818639ac48a4c6afa2f1581a8b9525e2.314.snapshot`.
+
+#### Create snapshot file
+
+`snapshot -create <chainID>`
+
+The command expect database of the chain in the current directory. The chain must be deactivated or the node must be stopped.
+
+The command iterates the detabase and creates the snapshot file.
```
jorgemmsilva:
```suggestion
The command iterates the database and creates a snapshot file.
```

tools/snapshot/readme.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Important.** The order in which key/value pairs in database are iterated, is **not-deterministic** in general. This means, that content of the snapshot file (and, therefore, digest/hash of it) is not guaranteed to be the same for the same source database.
+
+The above, however, does not violate consisency of the database. It is alway guaranteed, that *root commitment* to the state is fully deterministic.
```
jorgemmsilva:
```suggestion
The above, however, does not violate consistency of the database. It is always guaranteed, that *root commitment* to the state is fully deterministic.
```

tools/snapshot/readme.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Scan snapshot file
+
+`snapshot -scan <chainID>` scans the file, checks it formatting and extracts main parameters of the state stored in the snapshot itself (not in the file name):
```
jorgemmsilva:
```suggestion
`snapshot -scan <chainID>` scans the file, checks the formatting and extracts main parameters of the state stored in the snapshot itself (not in the file name):
```

suggestion: I feel like `scan` is not the right name here, what about `snapshot -info <file>` ?

packages/kv/kvstream.go
lunfardo314: yeah, I know, this is what linter thinks. I do not like its opinion. 
I want to use the returned value but not the type of it

packages/kv/kvstream.go
lunfardo314: the intention is to make most of it independent of wasp specific types, to move to hive.go

packages/vm/core/accounts/impl.go
lunfardo314: That was my intention, to forecefully put it into the init entry point. Two reasons why I didn't do it:
- significant refactoring, because init entry point may be absent
- there's no way to prevent to overwrite that by the SC itself, except in the k/v store implementation, which is ugly and also requires refactoring.

So, currently it is up to the SC developer: if it does  not put hname under nil key in the init, it is not guaranteed to get state commitment for the SC. Not so bad, taking into account that this feature not always needed

tools/snapshot/i.sh
lunfardo314: yes, I just wanted it short. Please refactor to your liking. I do a bash expert

tools/snapshot/readme.md
```diff
@@ -0,0 +1,118 @@
...
+
+It will work the following way:
+* restore the snapshot from file into the database
+* take root commitment of the restored state
+* ask the L2 node for the *proof of the past state* that the state root commitment with the specific state index of the restored state is commited in the chain state as a past state (the commitment should be  contained in the `blocklog` partition).
+* validate the *proof of the past state* with the L1 node.
+
+To perform the `-validate` operation we need the following WEB APIs:
+* L2 (Wasp) `get proof of inclusion` (see `viewcontext.GetMerkleProof` call)
+* L1: `get state commitment` which retrieves `anchor output` of the chain (`AliasOutput`) with the state commitment. The proof must be verified agains the state commitment stored in the anchor.
```
jorgemmsilva, line 98:
```suggestion
* L1: `get state commitment` which retrieves `anchor output` of the chain (`AliasOutput`) with the state commitment. The proof must be verified against the state commitment stored in the anchor.    
```

PR #891, opened 2022-03-22
Stardust vm julius

Juliusan:
New nodeconn implementation is used in chain subpackages


packages/chain/chainimpl/chainimpl.go
```diff
@@ -240,70 +249,69 @@ func (c *chainObj) receiveChainPeerMessages(peerMsg *peering.PeerMessageIn) {
...
 	c.log.Debugf("updateChainNodes, stateIndex=%v", stateIndex)
 	govAccessNodes := make([]*cryptolib.PublicKey, 0)
 	govCandidateNodes := make([]*governance.AccessNodeInfo, 0)
 	if stateIndex > 0 {
-		res, err := viewcontext.New(c).CallView(
+		res := viewcontext.New(c).Call(
 			governance.Contract.Hname(),
 			governance.FuncGetChainNodes.Hname(),
 			governance.GetChainNodesRequest{}.AsDict(),
+			nil, // TODO: *iscp.Allowance???
```
jorgemmsilva:
```suggestion
		res := viewcontext.New(c).CallViewExternal(
			governance.Contract.Hname(),
			governance.FuncGetChainNodes.Hname(),
			governance.GetChainNodesRequest{}.AsDict(),
```

jorgemmsilva (approved):

packages/chain/chainimpl/chainimpl.go
```diff
@@ -11,15 +11,16 @@ import (
 	"github.com/iotaledger/hive.go/kvstore"
 	"github.com/iotaledger/hive.go/logger"
 	"github.com/iotaledger/wasp/packages/chain"
-	"github.com/iotaledger/wasp/packages/chain/mempool"
+	mempool_pkg "github.com/iotaledger/wasp/packages/chain/mempool"
```
dessaya, line 14: why the rename? It should not be necessary since there is no other conflicting `mempool` package here

packages/chain/nodeconnchain/nodeconn_chain.go
```diff
@@ -0,0 +1,237 @@
...
+	result.txInclusionStateHandlerRef, err = result.nc.AttachTxInclusionStateEvents(result.chainAddr, result.txInclusionStateHandler)
+	if err != nil {
+		result.log.Errorf("cannot create chain nodeconnection: %v", err)
+		return nil, err
+	}
+	result.log.Debugf("chain nodeconnection created")
+	return &result, nil
+}
+
+func (nccT *nodeconnChain) outputHandler(outputID iotago.OutputID, output iotago.Output) {
```
dessaya, line 67: Nit: Not a fan of this `nccT` name; following Go coventions it should be `n` or `nc`.

`nccT` is a bit too long, plus it forces you to press shift to type it :rofl:

dessaya (approved):
LGTM, apart from the commented out code, I guess it will be handled later

packages/chain/nodeconnchain/nodeconn_chain.go
Juliusan: I'll have this in mind :-)

packages/chain/chainimpl/chainimpl.go
Juliusan: Linter complains about naming variable the same as package. And I can see why: it is hard to differ package name from variable name in Go. I decided to rename the package and leave the descriptive variable name as is. Would you prefer renaming the variable? O ignoring the linter warning?

dessaya: I see. Probably the most idiomatic thing to do would be renaming the package itself, or even moving all the mempool stuff to the `chain` package. This would also avoid the stuttering, i.e. instead of:

```
mempool := mempool.NewMempool()
```

we would have instead:

```
mempool := chain.NewMempool()
```

But this would probably be a non-trivial refactor (because of import loops etc). So for now I guess renaming the package when importing is ok.

kape1395: I would rename a variable.

PR #898, opened 2022-03-29
Docs/diataxis 2

aleksei-korolev:
# Description of change

Rewrite validators.md and consensus.md. There was much more to fix than just the style, there were some factual errors, and I also restructured it and cut some information out for ease of reading.

## Type of change

- Documentation Fix

> Describe the tests that you ran to verify your changes.

I've read it out loud.

> Make sure to provide instructions for the maintainer as well as any relevant configurations.

~~Read it out loud~~ Probably it doesn't make much sense to compare diffs because I completely overhauled the articles. Just read it as if it is a brand new article.

## Change checklist

- [X] I have followed the [contribution guidelines](https://wiki.iota.org/smart-contracts/contribute) for this project
- [X] I have performed a self-review of my own code
- [X] I have selected the `develop` branch as the target branch
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation

### DIFF
--- a/documentation/docs/guide/core_concepts/consensus.md
+++ b/documentation/docs/guide/core_concepts/consensus.md
@@ -1,54 +1,34 @@
 ---
-description: IOTA Smart Contracts Consensus
+description: IOTA Smart Contracts consensus is how Layer 2 validators agree to change the chain state in the same way.
 image: /img/logo/WASP_logo_dark.png
 keywords:
 - smart contracts
 - consensus
+- validator committee
+- validators
+- validator nodes
 - explanation
 ---
 # Consensus
 
-## Consensus on computations
+To update the chain, its committee needs to reach a consensus, meaning that more than two thirds of its validators have to agree to change the state in the exact same way. This prevents a single malicious node from wreaking havoc over the chain, of course, but there are also more mundane reasons for individual nodes to act up.
 
-Even if all committee nodes are honest (i.e. they have no malicious intent), there are factors which may make each node see things differently. This can lead to different inputs to the same program on different nodes and, consequently, to different results.
+Smart contracts are deterministic, so all honest nodes will produce the same output — but only if they have received same the input. Each validator node has its own point of access to the Tangle, so it may look a bit different to different nodes, as fresh transactions take time to propagate through the network. Validator nodes will receive smart contract requests with random delays in a random order, and, finally, all computers run on their own always slightly skewed clocks.
 
-There are several possible reasons for such an apparently non-deterministic outcome.
+## Batch Proposals
 
-Each committee node has its own access to the UTXO ledger, i.e. committee nodes are usually connected to different IOTA nodes. The reason for this is to not make access to the UTXO ledger a single point of failure, i.e. we also want access to the Tangle to be distributed. This may often lead to a slightly different perception of some aspects of the ledger, for example of the token balance in a particular address. Also, each node has its own local clock and those clocks may be slightly skewed, so there isn’t an objective time for nodes.
+As the first step, each node provides its own vision, a *batch proposal*. It contains a local timestamp, a list of unprocessed requests, and the node's partial signature of the current state's hash.
 
-The requests (UTXOs) may reach Wasp nodes in an arbitrary order and with arbitrary delays (even if these are usually close to the network latency).
+Then the nodes have to agree on which batch proposals they want to work on. In short, nodes A, B, and C have to confirm that they plan to work on proposals from A, B, and C, and from no one else. As long as there are more than two thirds of honest nodes, they will be able to find an *asynchronous common subset* of the batch proposals. From that point nodes have the same input and will produce the same result independently.
 
-Before starting calculations, nodes are required to have consensus on the following:
+## The Batch
 
-- The current state of the chain i.e. on the state output
-- Timestamp to be used for the next state transaction
-- Ordered batch of requests to be processed
-- Address where node fees for processing the request must be sent (if enabled)
-- Mana pledge targets
+The next step is to convert the raw list of batch proposals into an actual batch: for that, all requests from all proposals are counted and filtered to produce the same single list of requests in the same order.
 
-In order to achieve a bigger throughput, the committee picks requests from the on-ledger backlog and processes requests in batches, not one by one. This means the committee has to have a consensus on the batch of the requests and the order of the requests in the batch. After at least a quorum of committee nodes have a consensus on the above, honest committee members will always produce identical results of calculations.
+As all nodes had to sign the same piece of information, the hash of the current state, these partial signatures can be combined into a full, valid signature that is then fed to a pseudo-random function that orders the smart contract requests. Validator nodes can neither affect nor predict the final order of requests in the batch.
 
-## Proof of consensus
+## State Anchor
 
-Suppose a quorum of committee nodes has reached consensus on inputs and produced identical results, these being the block of state updates and the anchor transaction.
+After agreeing on the input, each node executes every smart contract request in order, producing the same new block independently. Each node then crafts a state anchor, a Layer 1 transaction that proves the commitment to this new chain state. The timestamp for this transaction is derived from the timestamps of all batch proposals.
 
-The anchor transaction contains chain state transition, the AliasOutput and token transfers, so it must be signed.
-
-**It is only possible to produce valid signatures of inputs of the anchor transaction by the quorum** of nodes. In this case, a confirmed anchor transaction becomes a cryptographical **proof of consensus** in the committee.  
-
-To achieve this, IOTA Smart Contracts uses **BLS threshold signatures in combination with polynomial (Shamir) secret sharing** to identify the address controlling the chain state. In order for the secret keys to be distributed across the chain validators, a DKG (Distributed Key Generation) procedure is executed when starting a chain (using the Rabin-Gennaro algorithm).
-
-## The Consensus Algorithm
-
-The committee is of fixed size, thus we use a Byzantine Fault Tolerant (BFT) Algorithm, which guarantees consistency and byzantine fault tolerance if less than ⅓ of nodes are malicious.
-
-As a basis for the IOTA Smart Contracts consensus, the Asynchronous Common Subset (ACS) part of the HoneyBadgerBFT algorithm is used, with the exception of how the proposals are combined.
-
-The rest of the consensus algorithm is built on top of the ACS. Each node supplies to the ACS its batch proposal which indicates a set of Request IDs, a timestamp, consensus and access mana pledge addresses, fee destination and a partial signature for generating non-forgeable entropy. Upon termination of the ACS, each honest node gets the same set of such proposals and aggregates them into the final batch in a deterministic way.
-
-It is ensured that all honest nodes have the same input for the VM. After running the selected batch, the VM results are then collectively signed using the threshold signature. The signed transaction can be published by any node at this point. In order to minimize the load on the IOTA network, the nodes calculate a delay for posting the transaction to the network based on a deterministic permutation of the nodes relative to the local perception of time.
-
-:::note
-The comprehensive overview of architectural design decisions of IOTA Smart Contracts can be found in the
-[whitepaper](https://github.com/iotaledger/wasp/raw/master/documentation/ISC_WP_Nov_10_2021.pdf).  
-:::
+All nodes then sign the state anchor with their partial keys and exchange these signatures. This way, every node obtains the same valid combined signature and the same valid anchor transaction, which means that any node can publish this transaction to Layer 1. In theory, nodes could publish these state anchors every time they update the state; in practice, they do it only every ten or so seconds to reduce the load on the Tangle.
--- a/documentation/docs/guide/core_concepts/validators.md
+++ b/documentation/docs/guide/core_concepts/validators.md
@@ -10,6 +10,6 @@ keywords:
 ---
 # Validators
 
-Each chain is run by a network of validator nodes, which run a consensus on the chain state updates. The [Wasp](https://github.com/iotaledger/wasp) node is an implementation of the validator node. The validators of the chain form a committee, a bound together closed set of nodes. The committee of the chain may change, allowing new validators and validator nodes to be added or replaced. This also makes the chain itself agnostic to its validators (the committee).
+Each chain is run by that chain's *committee of validators*. This committee owns a key that is split between all of its validators. Each key share is useless on its own, but a collective signature gives validators full control over the chain.
 
-Only when a supermajority of the validators (the quorum) of a chain reaches [consensus](./consensus.md), a new state update can be signed, which unlocks the AliasOutput for the chain and produces the next state UTXO.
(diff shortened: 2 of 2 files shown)


aleksei-korolev:
@iotaledger/tech-writers

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
-- The current state of the chain i.e. on the state output
-- Timestamp to be used for the next state transaction
-- Ordered batch of requests to be processed
-- Address where node fees for processing the request must be sent (if enabled)
-- Mana pledge targets
-
-In order to achieve a bigger throughput, the committee picks requests from the on-ledger backlog and processes requests in batches, not one by one. This means the committee has to have a consensus on the batch of the requests and the order of the requests in the batch. After at least a quorum of committee nodes have a consensus on the above, honest committee members will always produce identical results of calculations.
-
-## Proof of consensus
+To update the chain, its committee needs to reach a consensus, meaning that more than two thirds of its validators have to agree to change the state in the exact same way. This prevents a single malicious node from wrecking havoc over the chain, of course, but there are also more mundane reasons for individual nodes to act up.
```
lucas-tortora:
```suggestion
To update the chain, its committee needs to reach a consensus, meaning that more than two thirds of its validators have to agree to change the state in the exact same way. This prevents a single malicious node from wreaking havoc over the chain, but there are also more mundane reasons for individual nodes to act up.
```

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
 
-Suppose a quorum of committee nodes has reached consensus on inputs and produced identical results, these being the block of state updates and the anchor transaction.
+Smart contracts are deterministic, so all honest nodes will produce equal results — but only if the input was the same. Each validator node has its own access to The Tangle, so it may look a bit different to different nodes, as fresh transactions take time to propagate through the network. Validator nodes will receive smart contract requests with random delays in a random order, and, finally, all computers run on their own always slightly skewed clocks.
 
-The anchor transaction contains chain state transition, the AliasOutput and token transfers, so it must be signed.
+## Batch Proposals
 
-**It is only possible to produce valid signatures of inputs of the anchor transaction by the quorum** of nodes. In this case, a confirmed anchor transaction becomes a cryptographical **proof of consensus** in the committee.  
+As the first step, each node provides its own vision, a *batch proposal*. It contains a local timestamp, a list of unprocessed requests, and the node's partial signature of the current state's hash.
```
lucas-tortora, line 20:
```suggestion
At first, each node provides its vision, a *batch proposal*. It contains a local timestamp, a list of unprocessed requests, and the node's partial signature of the current state's hash.
```

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
 
-To achieve this, IOTA Smart Contracts uses **BLS threshold signatures in combination with polynomial (Shamir) secret sharing** to identify the address controlling the chain state. In order for the secret keys to be distributed across the chain validators, a DKG (Distributed Key Generation) procedure is executed when starting a chain (using the Rabin-Gennaro algorithm).
+Then nodes have to agree on which batch proposals they want to work on. In short, Alice, Bob, and Charlie have to confirm that they plan to work on proposals from Alice, Bob, and Charlie, and from no one else. As long as there are more than two thirds of honest nodes, they will be able to find an *asynchronous common subset* of the batch proposals. From that point nodes have the same input and will produce the same result independently.
```
lucas-tortora:
```suggestion
Then the nodes have to agree on which batch proposals they want to work on. In short, Alice, Bob, and Charlie have to confirm that they plan to work on proposals from Alice, Bob, and Charlie and no one else. As long as there are more than two thirds of honest nodes, they will be able to find an *asynchronous common subset* of the batch proposals. From that point, nodes have the same input and will produce the same result independently.
```
If you plan on using Alice Bob and Charlie as example it  may be a good idea to introduce them or the scenario. In this case I'd probably go with validator A, B, and C  or something like that.

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
 
-## The Consensus Algorithm
+## The Batch
 
-The committee is of fixed size, thus we use a Byzantine Fault Tolerant (BFT) Algorithm, which guarantees consistency and byzantine fault tolerance if less than ⅓ of nodes are malicious.
+The next step is to convert the raw list of batch proposals into an actual batch: for that, all requests from all proposals are counted and filtered to produce a single list of requests in a certain order.
```
lucas-tortora:
```suggestion
The next step is to convert the raw list of batch proposals into an actual batch: for that, all requests from all proposals are counted and filtered to produce a single list of requests in a particular order.
```

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
 
-As a basis for the IOTA Smart Contracts consensus, the Asynchronous Common Subset (ACS) part of the HoneyBadgerBFT algorithm is used, with the exception of how the proposals are combined.
+As all nodes had to sign the same piece of information, the hash of the current state, these partial signatures can be combined into a full, valid signature that is then fed to a pseudo-random function that orders the smart contract requests. Validator nodes cannot neither affect nor predict the final order of requests in the batch.
```
lucas-tortora:
```suggestion
As all nodes have to sign the same piece of information, the hash of the current state, these partial signatures can be combined into a full, valid signature that is then fed to a pseudo-random function that orders the smart contract requests. Validator nodes can neither affect nor predict the final order of requests in the batch.
```

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
+## State Anchor
 
-It is ensured that all honest nodes have the same input for the VM. After running the selected batch, the VM results are then collectively signed using the threshold signature. The signed transaction can be published by any node at this point. In order to minimize the load on the IOTA network, the nodes calculate a delay for posting the transaction to the network based on a deterministic permutation of the nodes relative to the local perception of time.
+After agreeing on the input, each node executes every smart contract request in order, producing the same new block independently. Each node then crafts a state anchor, a Layer 1 transaction that proves the commitment to this new chain state. The timestamp for this transaction is derived from timestamps of all batch proposals.
```
lucas-tortora:
```suggestion
After agreeing on the input, each node executes every smart contract request in order, producing the same new block independently. Each node then crafts a state anchor, a Layer 1 transaction that proves the commitment to this new chain state. The timestamp for this transaction is derived from the timestamps of all batch proposals.
```

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
 
-:::note
-The comprehensive overview of architectural design decisions of IOTA Smart Contracts can be found in the
-[whitepaper](https://github.com/iotaledger/wasp/raw/master/documentation/ISC_WP_Nov_10_2021.pdf).  
-:::
+All nodes then sign the state anchor with their partial keys and exchange these signatures. Now all nodes have a valid combined signature and any node can send the anchor transaction to Layer 1. In theory, nodes could publish every single crafted transaction, but in practice ISC committees skip over some state anchors to reduce the load on the Tangle.
```
lucas-tortora:
```suggestion
All nodes sign the state anchor with their partial keys and exchange these signatures, so all nodes have a valid combined signature, and any node can send the anchor transaction to Layer 1. In theory, nodes could publish every single transaction. However,  in practice, ISC committees skip over some state anchors to reduce the load on the Tangle.
```

documentation/docs/guide/core_concepts/validators.md
```diff
@@ -1,13 +1,16 @@
 ---
+description: Each chain is run by a network of validator nodes which run a consensus on the chain state update.
+image: /img/logo/WASP_logo_dark.png
 keywords:
-- Validators
+- validators
+- validator nodes
 - consensus
 - state update
```
lucas-tortora:
```suggestion
- state update
- explanation
```

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
 ---
-keywords:
-- Smart Contracts
-- Consensus
 description: IOTA Smart Contracts Consensus
```
lucas-tortora: The description could use a couple more words, this seems more like a title.

documentation/docs/guide/core_concepts/consensus.md
```diff
@@ -1,53 +1,31 @@
...
-
-## Proof of consensus
+To update the chain, its committee needs to reach a consensus, meaning that more than two thirds of its validators have to agree to change the state in the exact same way. This prevents a single malicious node from wrecking havoc over the chain, of course, but there are also more mundane reasons for individual nodes to act up.
 
-Suppose a quorum of committee nodes has reached consensus on inputs and produced identical results, these being the block of state updates and the anchor transaction.
+Smart contracts are deterministic, so all honest nodes will produce equal results — but only if the input was the same. Each validator node has its own access to The Tangle, so it may look a bit different to different nodes, as fresh transactions take time to propagate through the network. Validator nodes will receive smart contract requests with random delays in a random order, and, finally, all computers run on their own always slightly skewed clocks.
```
lucas-tortora:
```suggestion
Smart contracts are deterministic, so all honest nodes will produce identical results if the input is the same. Each validator node has its own connection to The Tangle. This means each validator node may have a slightly different perception of the Tangle, as new transactions take time to propagate through the network. Validator nodes will receive smart contract requests with random delays in a random order, and, finally, all computers run on their own always slightly skewed clocks.
```

documentation/docs/guide/core_concepts/validators.md
```diff
@@ -1,13 +1,16 @@
...
-description: Each chain is run by a network of validator nodes which run a consensus on the chain state update.
-image: /img/logo/WASP_logo_dark.png
 ---
 # Validators
 
-Each chain is run by a network of validator nodes, which run a consensus on the chain state updates. The [Wasp](https://github.com/iotaledger/wasp) node is an implementation of the validator node. The validators of the chain form a committee, a bound together closed set of nodes. The committee of the chain may change, allowing new validators and validator nodes to be added or replaced. This also makes the chain itself agnostic to its validators (the committee).
+Each chain is run by that chain's *committee of validators*. ISC does not define how do you select validators to form a committee: it could be a solitary choice of the chain's owner, or it could be a [public competition](https://wiki.assembly.sc/learn/introduction/) between candidates. ISC does not define how validators are rewarded, either.
```
lucas-tortora:
```suggestion
Each chain is run by that chain's *committee of validators*. ISC does not define how you select validators to form a committee: it could be a choice of the chain's owner or a [public competition](https://wiki.assembly.sc/learn/governance/dynamic_committees) between candidates. ISC does not define how validators are rewarded, either.
```

I'm not sure we have this feature yet.

documentation/docs/guide/core_concepts/validators.md
```diff
@@ -1,13 +1,16 @@
...
+
+Each committee is made of a known finite number of validators, and each validator holds its own part of a common key. On their own, validators cannot meddle with the chain. Together, they have full control over it.
```
lucas-tortora:
```suggestion
Each committee is composed of a known finite number of validators, and each validator holds its part of a common key. On their own, validators cannot meddle with the chain. Together, they have complete control over it.
```
Isn't it a shared key?

documentation/docs/guide/core_concepts/validators.md
```diff
@@ -1,13 +1,16 @@
...
 
-Only when a supermajority of the validators (the quorum) of a chain reaches [consensus](./consensus.md), a new state update can be signed, which unlocks the AliasOutput for the chain and produces the next state UTXO.
+Validators own computers that perform the actual calculations to maintain the chain. We call these computers validator nodes, and our software for running a validator node we called Wasp.
```
lucas-tortora:
```suggestion
Validators own computers that perform the actual calculations to maintain the chain. These computers are validator nodes, and the software running on the validator node is Wasp.
```

This paragraph seems to be a bit out of context here. Perhaps it'd work better at the beginning.

lucas-tortora (commented):
I like the direction the explanations are taking, trying to simplify things to reach a broader audience.  However, I think we may be losing technical content in the process which is probably relevant to the more technical audience.  

What would you think about adding the new article in a `key concepts` sort of section, while keeping the more technical decisions(BLS threshold, DKG procedure, etc) in a `reference` or `advanced concepts` section?

documentation/docs/guide/core_concepts/validators.md
aleksei-korolev: ahh, so I did everything except the thing that I was supposed to do.

documentation/docs/guide/core_concepts/consensus.md
charlesthompson3:
```suggestion
Smart contracts are deterministic, so all honest nodes will produce equal results — but only if the input was the same. Each validator node has its own access to the Tangle, so it may look a bit different to different nodes, as fresh transactions take time to propagate through the network. Validator nodes will receive smart contract requests with random delays in a random order, and, finally, all computers run on their own always slightly skewed clocks.
```

charlesthompson3 (commented):
One suggestion, basically just echoing Lucas' changes.

documentation/docs/guide/core_concepts/consensus.md
aleksei-korolev: I wanted it to sound it trivial on purpose. Like "There is an obvious reason that doesn't worth a discussion, but there is also a not-that-obvious reason that I actually want to talk about." Do you think it makes sense?

documentation/docs/guide/core_concepts/consensus.md
aleksei-korolev: I think I'll stick to my version. I think mine is less ambiguous.

documentation/docs/guide/core_concepts/consensus.md
aleksei-korolev: I have rewritten this sentence.

documentation/docs/guide/core_concepts/consensus.md
aleksei-korolev: Since the signing came from the previous step, I think I'll keep the past tense for the clarity sake.

aleksei-korolev:
I've applied your suggestions, but let's put it on hold until I know a better idea of what I am doing in general, at least until we merge  #904

jorgemmsilva (approved):

aleksei-korolev:
@jorgemmsilva I've rebased this one, too. As Lucas mentioned, I left some information out if you compare it to the original, so it's your call if you want to merge it now or wait until we have separate articles on stuff like BLS and DKG.

PR #910, opened 2022-04-04
Extend evm sandbox bindings

lmoe:
# Description of change

Adds registerError, panics/revert, partial Send (Requires serialization part at function call)

## Type of change

Added tests for each case


packages/vm/core/errors/external.go
```diff
@@ -19,6 +19,20 @@ func GetMessageFormat(code iscp.VMErrorCode, callView ViewCaller) (string, error
...
+	messageFormat, err := GetMessageFormat(*e, callView)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return iscp.NewVMErrorTemplate(*e, messageFormat), nil
+}
+
 func Resolve(e *iscp.UnresolvedVMError, callView ViewCaller) (*iscp.VMError, error) {
```
dessaya: I think `Resolve` now can be simplified using `ResolveErrorCode`

packages/vm/core/evm/emulator/emulator.go
```diff
@@ -199,28 +257,28 @@ func (e *EVMEmulator) applyMessage(msg core.Message, statedb vm.StateDB, header
 	return result, gasUsed, err
 }
 
-func (e *EVMEmulator) SendTransaction(tx *types.Transaction, gasLimit uint64) (*types.Receipt, uint64, error) {
+func (e *EVMEmulator) SendTransaction(tx *types.Transaction, gasLimit uint64) (*types.Receipt, uint64, error, *revertError) {
```
dessaya: I think decoding the error should not be a responsibility of the emulator, since it is agnostic of ABIs and ISC. Can't all this new VMError logic be moved to `evmimpl` or somewhere else?

packages/vm/core/evm/isccontract/ISC.sol
```diff
@@ -50,6 +49,45 @@ struct ISCDict {
...
 
+struct IotaFungibleTokens {
+	uint64 iotas;
+	IotaNativeToken[] tokens;
+}
+
+struct ISCSendMetadata  {
+	ISCHname targetContract;
+	ISCHname entrypoint;
+	//mapping(string => bytes) params;
```
dessaya, line 60: mapping does not work here I think. It has to be an `ISCDict`.

packages/vm/core/evm/isccontract/ISC.sol
```diff
@@ -50,6 +49,45 @@ struct ISCDict {
 	ISCDictItem[] items;
 }
 
+struct IotaFungibleTokens {
```
dessaya, line 52: Should be called `ISCFungibleTokens`?

packages/vm/core/evm/emulator/emulator.go
```diff
@@ -91,11 +90,14 @@ func NewEVMEmulator(store kv.KVStore, timestamp uint64, iscContract vm.ISCContra
...
 	}
+
+	logger := logger.NewJSONLogger(&logger.Config{EnableReturnData: true, EnableMemory: true}, os.Stdout)
+
 	return &EVMEmulator{
 		timestamp:   timestamp,
 		chainConfig: makeConfig(int(bdb.GetChainID())),
 		kv:          store,
-		vmConfig:    vm.Config{ISCContract: iscContract},
+		vmConfig:    vm.Config{ISCContract: iscContract, Debug: true, Tracer: logger},
```
dessaya: Debug is left enabled intentionally? I find it produces an obscene amount of log output and it's difficult to find anything else. Also the tracer might use extra resources and it is not needed in normal operation.

Maybe make this configurable somehow?

packages/vm/core/evm/emulator/error.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/iotaledger/wasp/packages/iscp"
+	"golang.org/x/xerrors"
+)
+
+type RevertError struct {
+	msg    string
+	reason string // revert reason hex encoded
```
dessaya: What is the hex-encoded reason used for? Why is it hex instead of []byte?

dessaya (approved):

packages/vm/core/evm/emulator/emulator.go
lmoe: I needed it for debugging and forgot to remove it! :) 
I can make it configurable in the future.

PR #916, opened 2022-04-06
Add none authentication

lmoe:
# Description of change

Makes it possible to disable the authentication. This might be useful in testing environments especially in combination with Docker. This PR will disable the WebApi authentication in the **DevNet** config  and remove the setting `adminWhitelistDisabled` as this was essentially used to disable the authentication before.

## Type of change


documentation/docs/guide/chains_and_nodes/running-a-node.md
```diff
@@ -145,6 +145,8 @@ By default, Wasp accepts any API request coming from `127.0.0.1`. The Dashboard
 
 Both authentication methods allow any form of request and have therefore 'root' permissions.
 
+You can disable the authentication per endpoint by setting `scheme` to `none` on any `auth` parent block like [this](https://github.com/iotaledger/wasp/blob/6b9aa273917c865b0acc83df9a1935f49498e43d/docker_config.json#L58).
```
jorgemmsilva:
```suggestion
You can disable the authentication per endpoint by setting `scheme` to `none` on `webapi.auth`, like [this](https://github.com/iotaledger/wasp/blob/6b9aa273917c865b0acc83df9a1935f49498e43d/docker_config.json#L58).
```

jorgemmsilva (approved):

documentation/docs/guide/chains_and_nodes/running-a-node.md
lmoe: I wrote any because we have two different endpoints (dashboard, webapi) and you can configure both independently. 

How about rather this: ```You can disable the authentication per endpoint by setting `scheme` to `none` on any `auth` block such as `webapi.auth` or `dashboard.auth`. [Example configuration](https://github.com/iotaledger/wasp/blob/6b9aa273917c865b0acc83df9a1935f49498e43d/docker_config.json#L58).```? @jorgemmsilva

jorgemmsilva: makes sense, yes - I like that we are specific about it :+1: Thanks!

lmoe:
After discussion, we will merge regardless of the failing builds. The build failing is not caused by an build error and is misleading. Fixes happen inside the stardust branch.

PR #935, opened 2022-04-11
Fix: missing inputcommitment

jorgemmsilva:
because of the difference between using:

`(u *TransactionEssence) SigningMessage()`
or
`(u *TransactionEssence) Sign(inputsCommitment []byte, addrKeys ...AddressKeys) ([]Signature, error)`

when using `SigningMessage`, as we are doing in consensus, the inputsCommitment was not being added to the tx essence, now it is included when the essence is generated

### DIFF
--- a/packages/chain/nodeconnchain/nodeconn_chain.go
+++ b/packages/chain/nodeconnchain/nodeconn_chain.go
@@ -129,7 +129,6 @@ func (nccT *nodeconnChain) txInclusionStateHandler(txID iotago.TransactionID, st
 		txID:  txID,
 		state: state,
 	}
-	nccT.log.Debugf("handling inclusion state of tx ID %v: inclusion state %v handled", txIDStr, state)
 }
 
 func (nccT *nodeconnChain) AttachToAliasOutput(handler chain.NodeConnectionAliasOutputHandlerFun) {
--- a/packages/chain/nodeconnchain/nodeconn_chain.go
+++ b/packages/chain/nodeconnchain/nodeconn_chain.go
@@ -129,6 +129,7 @@ func (nccT *nodeconnChain) txInclusionStateHandler(txID iotago.TransactionID, st
 		txID:  txID,
 		state: state,
 	}
+	nccT.log.Debugf("handling inclusion state of tx ID %v finished", txIDStr)
 }
 
 func (nccT *nodeconnChain) AttachToAliasOutput(handler chain.NodeConnectionAliasOutputHandlerFun) {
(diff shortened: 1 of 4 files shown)


packages/chain/nodeconnchain/nodeconn_chain.go
```diff
@@ -129,7 +129,6 @@ func (nccT *nodeconnChain) txInclusionStateHandler(txID iotago.TransactionID, st
 		txID:  txID,
 		state: state,
 	}
-	nccT.log.Debugf("handling inclusion state of tx ID %v: inclusion state %v handled", txIDStr, state)
```
Juliusan, line 132: Why have you removed this? It is usually beneficial to see that the code is not stuck on blocked channel.

packages/vm/vmcontext/vmtxbuilder/txbuilder.go
```diff
@@ -208,7 +208,11 @@ func (txb *AnchorTransactionBuilder) BuildTransactionEssence(stateData *state.L1
 		Outputs:   txb.outputs(stateData),
 		Payload:   nil,
 	}
-	return essence, inputIDs.OrderedSet(inputs).MustCommitment()
+
+	inputsCommitment := inputIDs.OrderedSet(inputs).MustCommitment()
+	copy(essence.InputsCommitment[:], inputsCommitment)
+
```
Juliusan, line 214: Shouldn't this be part of VM code? Transaction essence comes from VM and I think `InputsCommitment` should be set there too. LGTM if this is only a temporar change.

packages/chain/nodeconnchain/nodeconn_chain.go
jorgemmsilva: there is a very similar log above, this was error inducing (it seems like the state message actually is  `inclusion state %v handled`). We both made that wrong judgment today.
we can change the log to something else, tho.

packages/vm/vmcontext/vmtxbuilder/txbuilder.go
jorgemmsilva: `vmtxbuilder` is literally in the VM :stuck_out_tongue:

packages/chain/nodeconnchain/nodeconn_chain.go
Juliusan: Yes... wording might be misleading. I agree.

packages/vm/vmcontext/vmtxbuilder/txbuilder.go
Juliusan: Sorry, my bad. Somehow from our talk I understood that this change is in `SigningMessage` method and haven't checked that.

packages/chain/nodeconnchain/nodeconn_chain.go
jorgemmsilva: updated with a different message

Juliusan (approved):

PR #946, opened 2022-04-13
Stardust VM,  ADKG

kape1395:
Asynchronous DKG.


dessaya (approved):

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,235 @@
...
+	//
+	// Mark the message as received.
+	h := r.ensureValueStored(msg.v)
+	if _, ok := r.echoRecv[h]; !ok {
+		r.echoRecv[h] = map[gpa.NodeID]bool{}
+	}
+	r.echoRecv[h][msg.s] = true
+	//
+	// Send the READY messages and support other ECHO messages, if quorum is reached.
+	if float64(len(r.echoRecv[h])) >= math.Ceil(float64(r.n+r.f)/2.0) {
```
Wollac:
```suggestion
	// Send the READY message, if Byzantine quorum ⌈(n+f+1)/2⌉ of received ECHO messages is reached.
	// As there are only n distinct peers, every two Byzantine quorums overlap in at least one correct peer.
	// |echoRecv| ≥ ⌈(n+f+1)/2⌉ ⟺ |echoRecv| > ⌊(n+f)/2⌋
	if len(r.echoRecv[h]) > (r.n+r.f)/2 {
```
I think there is a "+1" missing in the quorum size and also the float computations could potentially pose problems. I've explained my reasoning the the proposed changes.

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,235 @@
...
+	for i := range r.peers {
+		msgs = append(msgs, &msgBracha{
+			t: msgBrachaTypeInitial,
+			r: r.peers[i],
+			v: inputVal,
+		})
+	}
+	r.initialSent = true
+	return msgs
+}
```
Wollac:
```suggestion
	msgs := r.broadcast(msgBrachaTypeInitial, inputVal)
	r.initialSent = true
	return msgs
}

func (r *rbc) broadcast(t msgBrachaType, v []byte) []gpa.Message {
	var msgs []gpa.Message
	for i := range r.peers {
		msgs = append(msgs, &msgBracha{
			t: t,
			r: r.peers[i],
			v: v,
		})
	}
	return msgs
}
```
The exact same loop is executed several times, so I'd suggest to create a helper function for this.

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,235 @@
...
+	r.readyRecv[h][msg.s] = true
+	count := len(r.readyRecv[h])
+	//
+	// Decide, if quorum is enough.
+	if count >= 2*r.f+1 && r.output == hashing.NilHash {
+		r.output = h
+	}
+	//
+	// Support others, if quorum is enough.
+	if count >= r.f+1 {
```
Wollac:
```suggestion
	// Send the READY message, when a READY message was received from at least one honest peer.
	// This amplification assures totality.
	if count > r.f {
```
Only slight beautification

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,235 @@
...
+func (r *rbc) handlePredicateUpdate(msg msgPredicateUpdate) []gpa.Message {
+	r.predicate = msg.predicate
+	resendMsgs := r.pendingMsgs
+	r.pendingMsgs = []gpa.Message{}
+	return resendMsgs // The OwnHandler will resend them back.
+}
+
+func (r *rbc) maybeSendEchoReady(v []byte) []gpa.Message {
+	msgs := []gpa.Message{}
+	if !r.echoSent {
```
Wollac: What is the reasoning behind also sending an ECHO (if not done yet) together with the READY?
For correctness it is definitely not needed  and I don't think I have ever seen this version in the literature...
Is it supposed to improve performance? I would intuitively just say it leads to more unnecessary messages. I would thus remove the ECHO from here to be closer to the algorithm.

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	case *msgPredicateUpdate:
+		return r.handlePredicateUpdate(*msgT)
+	default:
+		panic(xerrors.Errorf("unexpected message: %+v", msg))
+	}
+}
+
+func (r *rbc) handleInitial(msg *msgBracha) []gpa.Message {
+	if r.echoSent {
```
Wollac: It is crucial here to check that `msg` is coming from the broadcaster! Only they are allowed to issues an INITIAL message.

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,235 @@
...
+		})
+	}
+	r.echoSent = true
+	return msgs
+}
+
+func (r *rbc) handleEcho(msg *msgBracha) []gpa.Message {
+	//
+	// Mark the message as received.
+	h := r.ensureValueStored(msg.v)
```
Wollac: I see an amplification attack here:
- there are no limits on the size of `v`
- it could be huge and exceed to peer's memory (either directly or when hashed)
- or cause problems when the peer broadcasts this huge message to other peers
- too large messages could be introduced by a malicious party at any stage of the algorithm.

As a simple fix, I would suggest to add an additional (global) config parameter to the constructor and check _all_ incoming messages against this size. For the actual program we can then either use the max message size, or come up with a reasonable max (maybe like 1MB)

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,235 @@
...
+	echoRecv    map[hashing.HashValue]map[gpa.NodeID]bool // Quorum counter for the ECHO messages.
+	readySent   bool                                      // Have we sent the READY messages?
+	readyRecv   map[hashing.HashValue]map[gpa.NodeID]bool // Quorum counter for the READY messages.
+	output      hashing.HashValue
+}
+
+var _ gpa.GPA = &rbc{}
+
+// Update predicate for the RBC instance.
+func SendPredicateUpdate(rbc gpa.GPA, me gpa.NodeID, predicate func([]byte) bool) []gpa.Message {
```
Wollac: This predicate updating seems to be rather error-prone. Is it even still used somewhere?
Otherwise, I would strongly suggest to just remove it complete (and maybe only keep a static global predicate function in the constructor)

Wollac (changes requested):

Wollac:
Only reviewed Bracha for now

packages/gpa/rbc/bracha/bracha.go
kape1395: Good catch. Was probably lost between f and t.

packages/gpa/rbc/bracha/bracha.go
kape1395: I have no strong opinion here. I personally prefer the `count > r.f`, as you suggested. But in a lot of papers and implementations the quorums are expressed as `count >= r.f+1`. Let's keep the simpler version. The same will be in a lot of other places.

packages/gpa/rbc/bracha/bracha.go
kape1395: Indeed, will add
```
	if msg.s != r.broadcaster {
		return gpa.NoMessages()
	}
```

packages/gpa/rbc/bracha/bracha.go
kape1395: I was trying to be close to the algorithm in the paper:
```
//      Broadcast(v)
//      step 0. (By process p)
//          Send (initial, v) to all the processes.
//      step 1. Wait until the receipt of,
//              one (initial, v) message
//              or (n+t)/2 (echo, v) messages
//              or (t+l) (ready, v) messages
//              for some v.
//          Send (echo, v) to all the processes.
//      step 2. Wait until the receipt of,
//              (n+t)/2 (echo, v) messages
//              or (t+1) (ready, v) messages
//              (including messages received in step 1)
//              for some v.
//          Send (ready, v) to all the processes.
//      step 3. Wait until the receipt of,
//              2t+1 (ready, v) messages
//              (including messages received in step 1 and step 2) for some v.
//          Accept v.
```
Here, e.g. if a process don't receive the `(initial, v)` message but gets $(n+t)/2$ `(echo, v)` messages it suddenly satisfies conditions for steps 2 and 3, thus has to send `echo` and `ready` messages. This function does that.

Wollac: OK, fair enough. Seems like I've never really read the original paper :see_no_evil: 
Turns you don't need to do that though. See for example here for an excellent write-up and proof: https://dcl.epfl.ch/site/_media/education/sdc_byzconsensus.pdf
I am mean the current version is obviously also fine. What would you intuitively say? You think the version without the ECHO could be preferable as it involves less messages?

kape1395: Intuitively -- I agree, it should be enough to send the READY messages.

Not sure, if that's true under partial/asymmetric partitions. Or we don't consider them. I guess a mis-configuration in the peering can easily cause those.

packages/gpa/rbc/bracha/bracha.go
kape1395: I simplified the predicate handling just now, only a single message can be postponed, to I made that explicit.
It is used in `packages/gpa/adkg/das`, but the latter is not used. Nevertheless, I would like to keep this function.
Update of the predicate is well defined, while the predicate sequence is monotonic IMO (i.e. if a message was accepted by a predicate, it has to be accepted by the following predicates as well).

kape1395: Added  a TLA spec for this [BrachaRBC.tla](https://github.com/iotaledger/wasp/blob/stardust-vm-adkg/packages/gpa/rbc/bracha/BrachaRBC.tla).

packages/gpa/rbc/bracha/bracha.go
kape1395: Aligned the implementation (and the TLA spec) with the algorithm presentation in Das et al.
The ECHO messages are not sent upon receiving READY anymore.

packages/gpa/rbc/bracha/bracha.go
kape1395: Agree. Despite its hard to estimate the limit, it should be used here. Will add.

Also, the current code can store multiple values sent by a a single peer.
A correct node only send a message of particular type once, thus we should probably check, if a message of that was already received from the peer before storing the value. I will ignore the remaining messages, they can be caused by message re-delivery, not just a byzantine node.
This invariant holds, according to my spec:
```
SingleValueFromPeerPerMsgType ==
    \A m1, m2 \in msgs : (
        /\ m1.src \in CN
        /\ m1.src = m2.src
        /\ m1.t = m2.t
    ) => m1.v = m2.v
```

Wollac: That is indeed a very good point!
Every (honest) peer sends each message type at most once. There should be a check for that very early in the handling and just reject any duplicates!

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,289 @@
...
+//	10:     send ⟨READY, 𝑀⟩ to all
+func (r *rbc) handleEcho(msg *msgBracha) []gpa.Message {
+	//
+	// Mark the message as received.
+	h := r.ensureValueStored(msg.v)
+	r.markEchoRecv(h, msg)
+	//
+	// Send the READY message, if Byzantine quorum ⌈(n+f+1)/2⌉ of received ECHO messages is reached.
+	// As there are only n distinct peers, every two Byzantine quorums overlap in at least one correct peer.
+	if len(r.echoRecv[h]) > 2*r.f {
```
Wollac:
```suggestion
	// |echoRecv| ≥ ⌈(n+f+1)/2⌉ ⟺ |echoRecv| > ⌊(n+f)/2⌋
	if len(r.echoRecv[h]) > (r.n+r.f)/2 {
```
$|E| > 2f$ is _only_ equivalent to the Byzantine quorum, if $n=3f+1$. If we happen to know that the number of faulty peers is lower than that, it is no longer correct.
Maybe you have strong assumptions and checks in WASP that $n=3f+1$ is indeed correct, but I would still strongly suggest to use the formula for the proper Byzantine quorum to be closer to the actual algorithm and be safe in all eventualities. 
Also, maybe it is just be but I don't find these different representations with $n$ and $f$ not trivial and hard to understand. That's why I proposed to add an explanation again why `⌊(n+f)/2⌋` is indeed equivalent. But feel free to ignore this additional comment if it is just me :sweat_smile:

packages/gpa/rbc/bracha/bracha.go
Wollac: OK got it now. That single message is already making it much better (although I think it needs to be combined with the check from https://github.com/iotaledger/wasp/pull/946/files#r955842752)

Having a static predicate for the RBC is a very common trick, that you find in various algorithms and it makes definitely sense to integrate that. However, as I've said, usually it is assumed to be global and _static_.
As mentioned by you, Das et al. do it a bit differently by "switching" the RBC on when a certain condition is met.
In their paper they write "Participate in $j^{th}$ RBC when ...". They technically don't use the _Predicate_ for that. 
So an alternative implementation of Das Algorithms would be to cache all messages of the $j^{th}$ RBC in the DKG and only forward them to the RBC instance (without any predicate). Maybe this would even lead to a cleaner implementation as you automatically get the guarantee that you also know all the elements once the RBC finishes...

Anyways, I would suggest:

1. Remove the unused `packages/gpa/adkg/das` from this PR and into a separate branch.
2. Shortly consider whether we still need such a dynamic predicate. The monotonicity assumption is a bit unintuitive, but if you like to keep this functionality that is also fine. However, in that case we should make it a public method like `ChangePredicate`. Having it changed by a message is just dangerous and the wrong concept as no "outsider" must be allowed to change the predicate remotely.

packages/gpa/rbc/bracha/bracha.go
```diff
@@ -0,0 +1,289 @@
...
+// Handle the PROPOSE messages.
+//
+//	06: upon receiving ⟨PROPOSE, 𝑀⟩ from the broadcaster do
+//	07:     if 𝑃(𝑀) then
+//	08:         send ⟨ECHO, 𝑀⟩ to all
+func (r *rbc) handlePropose(msg *msgBracha) []gpa.Message {
+	if msg.s != r.broadcaster {
+		// PROPOSE messages can only be sent by the broadcaster process.
+		// Ignore all the rest.
+		return gpa.NoMessages()
```
Wollac: Why not return `nil` instead. That would be the more Go idiomatic way and have exactly the same properties.

Wollac (changes requested):

packages/gpa/rbc/bracha/bracha.go
kape1395: - Added checks for duplicates `checkMsgRecv(...)`.
  - It appears we don't need to store the values in memory at all.
  - Added a limit on message size.

Wollac: Yep, those changes look good!

packages/gpa/rbc/bracha/bracha.go
kape1395: Good point. Will change that in other branch. I have started to reorganize things there a bit.

main.go
```diff
@@ -25,6 +30,13 @@ import (
 )
 
 func main() {
+	// TODO: Remove before merging to develop
+	go func() {
+		log.Println(http.ListenAndServe("0.0.0.0:6060", nil))
+	}()
+	runtime.SetMutexProfileFraction(1)
+	runtime.SetBlockProfileRate(1)
+
```
jorgemmsilva: this should be removed.
it is enough to turn on profiling via the config to enable pprof:
https://github.com/iotaledger/wasp/blob/develop/packages/parameters/parameters.go#L112-L113

kape1395: Maybe that could be made a plugin, that is disabled by default?

jorgemmsilva: Yup, that is the case already 😄

kape1395: @lmoe , maybe we can remove this part already?

lmoe: Yea I just didn't know that we have that already and I wanted to fix the bugs quickly. :D 
Of course we can remove that there.

packages/gpa/rbc/bracha/bracha.go
kape1395: OK, I'll remove that predicate update along with the adkg/das.
The ChangePredicate cannot be sent form outside, because it has no implementation for serialization.

main.go
kape1395: OK, I removed it.

Wollac (commented):
Bracha looks good to me now!
How do we wanna proceed? Merge everything for now and continue review in a different branch or continue in this PR?

kape1395:
@Wollac , I continue improving stuff on the `gpa` branch (https://github.com/iotaledger/wasp/pull/1259) . Maybe you can proceed with a review there, or we can create a separate branch for it..

Wollac (approved):
A more thorough review will happen in a different branch

PR #949, opened 2022-04-14
Feat: Request Receipt endpoint

jorgemmsilva:
Currently we have only the `/status` endpoint which is used across all tests to check whether a request was processed.
This doesn't actually tells us if the request was handled *successfully*.
This PR changes the webapi to expose a `/receipt` endpoint instead, and changes everything to expect a receipt.
This receipt should have the VMError translated to improve user experience.
Also fixes minor bugs found along the way


dessaya (approved):

packages/chain/chain.go
```diff
@@ -50,7 +51,8 @@ type ChainEntry interface {
 
 // ChainRequests is an interface to query status of the request
 type ChainRequests interface {
-	GetRequestProcessingStatus(id iscp.RequestID) RequestProcessingStatus
+	GetRequestReceipt(id iscp.RequestID) (*blocklog.RequestReceipt, error)
+	TranslateError(e *iscp.UnresolvedVMError) (string, error)
```
dessaya: IMO this should be `ResolveError(e *iscp.UnresolvedVMError) (*iscp.VMError, error)` to make it more generic. If the caller needs the formatted string, they can call `.Error()` on the returned `VMError`.

jorgemmsilva: sounds good, will refactor :ok_hand:

PR #957, opened 2022-04-20
Fix: misc fixes for access nodes

jorgemmsilva:


packages/chain/chain.go
```diff
@@ -96,17 +96,17 @@ type (
 type NodeConnection interface {
 	RegisterChain(chainID *iscp.ChainID, stateOutputHandler, outputHandler func(iotago.OutputID, iotago.Output))
 	UnregisterChain(chainID *iscp.ChainID)
-	//----------delimeter to appease linter
+
```
kape1395, line 99: Interesting. The linter was complaining on these empty lines. Was that changed with some lib update?

packages/nodeconn/nc_chain.go
```diff
@@ -201,16 +217,7 @@ func (ncc *ncChain) subscribeToChainOwnedUTXOs() {
...
-	//
-	// Subscribe to the new outputs first.
-	eventsCh, subInfo := ncc.nc.mqttClient.AliasOutputsByID(*ncc.chainID.AsAliasID())
-	if subInfo.Error() != nil {
-		ncc.log.Panicf("failed to subscribe: %w", subInfo.Error())
-	}
-
-	//
-	// Then fetch all the existing unspent outputs owned by the chain.
+func (ncc *ncChain) queryLatestChainStateUTXO() {
```
kape1395, line 220: I was hoping to avoid having such functions. If we have all the stream of data, we don't have to poke the node for the stuff anymore. Maybe I don't know some context.

packages/nodeconn/nodeconn.go
```diff
@@ -193,15 +193,26 @@ func (nc *nodeConn) Close() {
 }
 
 func (nc *nodeConn) PullLatestOutput(chainID *iscp.ChainID) {
```
kape1395, line 195: Was hoping to drop all those Pull... and ...Latest functions. Maybe we will reconsider it later.

tools/cluster/cluster.go
```diff
@@ -164,7 +176,7 @@ func (clu *Cluster) DeployChain(description string, allPeers, committeeNodes []i
 		return nil, xerrors.Errorf("DeployChain: %w", err)
 	}
 
-	committeePubKeys := make([]string, 0)
+	committeePubKeys := make([]string, len(committeeNodes))
```
kape1395: Is that correct? You create a non-empty array, and then append to it. E.g.
```
func TestXxx(t *testing.T) {
	arr := make([]int, 3)
	arr = append(arr, 2)
	arr = append(arr, 3)
	arr = append(arr, 5)
	t.Logf("XXXXXXXXXX: %v+", arr)
}
```
outputs
```
== RUN   TestXxx
    adkg_test.go:97: XXXXXXXXXX: [0 0 0 2 3 5]+
--- PASS: TestXxx (0.00s)
PASS
```
Consider the zeros.

kape1395 (approved):
Looks good to me. Left some minor comments.
A exception is maybe the question regarding `make(..., 0)` -> `make(..., len(...))`.

packages/nodeconn/nc_chain.go
jorgemmsilva: yes, this was only used when we start the subscription. now its extracted to be re-usable because of this function that is used by the syncmgr:
https://github.com/jorgemmsilva/wasp/blob/stardust-vm-jorge/packages/nodeconn/nodeconn.go#L195

packages/nodeconn/nodeconn.go
jorgemmsilva: see https://github.com/iotaledger/wasp/pull/957#discussion_r856000661
used by the syncmgr, maybe we can refactor it so it doesn't need it, but I didn't want to make such breaking changes.

tools/cluster/cluster.go
jorgemmsilva: well spotted, thank you! :+1: 
fixed.

packages/chain/chain.go
jorgemmsilva: :man_shrugging: I don't see any linting error, there was an update to the golangci-lint version some time ago, maybe that fixed it

PR #971, opened 2022-04-27
Fix: timelocked requests cluster tests

jorgemmsilva:
fixes:
 - `http` was hardcoded in nodeconn, now the entire URL (including prefix) is taken from the config
 - time data coming from L1 is in seconds, no t nanoseconds
 - sandbox `Timestamp()` now states that it returns time in seconds, so it matches the real precision
 - use consensus time assumption to run the batch selected from ACS, previously no timedata was being used. Its possible we want nodes to agree on this time assumption during the batch selection instead, so that the time assumptions of nodes always match - @lunfardo314 @kape1395  what do you think?


packages/vm/core/evm/evmtest/evm_test.go
```diff
@@ -264,7 +264,7 @@ func TestISCTimestamp(t *testing.T) {
 	var ret int64
 	evmChain.ISCContract(evmChain.faucetKey).callView(nil, "getTimestampUnixNano", nil, &ret)
```
dessaya: `getTimestampUnixNano` should be renamed?

packages/iscp/sandbox_interface.go
```diff
@@ -32,7 +32,7 @@ type SandboxBase interface {
 	ContractAgentID() *AgentID
 	// ContractCreator returns the agentID that deployed the contract
 	ContractCreator() *AgentID
-	// Timestamp returns the UnixNano timestamp of the current state
+	// Timestamp returns the Unix timestamp of the current state in seconds
 	Timestamp() int64
```
dessaya: Did you adapt all usages of `Timestamp`?
E.g. if I search the repo for `UnixNano` I get like 20 matches.

Also, while we are at it, could we improve the typing? The caller should not have to read the documentation to know if the timestamp is seconds or nanos. Some alternatives:
* return `time.Time` instead of `uint64`
* return `UnixSeconds` and create `type UnixSeconds uint64`
* or at least rename to `TimestampUnixSeconds()` or just `UnixSeconds()`

jorgemmsilva: you're right, I thought about that but figured out we would still need to have it in int64 because it gets passed to the wasm sandbox. Its probably fine, tho - will make that change, I agree its much better. Thanks for the review :+1:

dessaya (approved):

kape1395 (approved):

PR #973, opened 2022-04-28
chore: add Ethereum address AgentID

dessaya:
As a first step to making EVM a first-class citizen, this commit adds a
subtype of AgentID that is composed of an alias address + an Ethereum
address, representing an Ethereum address inside the ISC chain.

The change is backwards compatible: the serialized bytes of non-ethereum
AgentID are exactly the same as before (this is mainly to avoid breaking
wasm code).


packages/iscp/agentid.go
```diff
@@ -4,97 +4,96 @@
...
+// * an ISC contract inside an ISC chain (serialized as: alias address + hname)
+// * an Ethereum address inside an ISC chain (serialized as:
+//   tagEthAddressAgentID + alias address + eth address)
 type AgentID struct {
+	// if a is nil => this is NilAgentID
 	a iotago.Address
+	// h can be != 0 only if a is an alias address, and eth == nil
 	h Hname
+	// eth can be != nil only if a is an alias address, and h == 0
+	eth *common.Address
```
lunfardo314: - implementation of the Ethereum address type on L1 may be a better way. This will be transparent for the code and no need to add eth address into the agent id as a field
- Meanwhile `hname` field should fit 2 types of hnames: the current one for ISC SCs and Ethereum SC identified (address?) for the Eth SCs. The follwowing combinations of AgentID would makes sense
-- (Address, 0) - for L1 addresses including Ethereum address
-- (ChainID, hname) - ISC SC
-- (ChainID, Eth SC ID) - Eth chain SC

lunfardo314: so, I suggest before going this way we first try to implement Eth address as 1st class citizen on L1

dessaya:
I accidentally replaced my branch with some other code. I'm closing this PR and eventually opening another one when needed,

PR #974, opened 2022-04-29
State manager

Juliusan:
Refactored state manager synchronisation algorithm. As block essence hash is included in alias output, the catching up algorithm may be changed to deterministically obtain block sequence from the last block to the first one. There is no need to try choosing every possible block candidate from the first one to the last one and check if the obtained sequence is correct.

Unit tests in `stardust-test` script passes as well as deploy cluster tests. `TestMissingRequests` was passing until I merged recent stardust-vm changes.


lunfardo314 (approved):

packages/chain/statemgr/action.go
```diff
@@ -109,7 +109,9 @@ func (sm *stateManager) addStateCandidateFromConsensus(nextState state.VirtualSt
 		return false
 	}
 	block.SetApprovingOutputID(approvingOutputID)
-	sm.addBlockAndCheckStateOutput(block, nextState)
+	sm.syncingBlocks.startSyncingIfNeeded(nextState.BlockIndex())
+	sm.syncingBlocks.addBlockCandidate(block, nextState)
```
jorgemmsilva, line 113: I don't fully grok how this process works, but shouldn't we `addBlockCandidate` before `startSyncingIfNeeded`?

packages/util/permute_test.go
```diff
@@ -4,16 +4,93 @@ import (
...
+	// There is p = 1 - 3628800!/(3628795!*3628800^5) ~ 0.00028% probability of obtaining at least two equal permutations while picking 5 random ones.
+	// There is q = 1 - (1-p)^10 ~ 0.0028% probability of obtaining at least two equal permutations while picking 5 random ones if you have 10 tries to do that.
+	// q is a probability of test returning false positive on initial permutation length of 10. ~ 1 false positive in 36k runs.
+	// For 9 q ~ 0.028% ~ 1 false positive in 3.6k runs.
+	// For 8 q ~ 0.25% ~ 1 false positive in 400 runs.
+	// For 7 q ~ 2.0% ~ 1 false positive in 50 runs.
+	// For 6 q ~ 13.0% ~ 1 false positive in 8 runs.
+	// For 5 q ~ 57.0% ~  false positive every second run.
+	// That is why `n` starts at 10.
+	// NOTE: the exact values might be a little bit of due to accuracy of float arythmetics in LibreOffice Calc. They are provided just to get an impression.
```
jorgemmsilva, line 66: do we just accept that we have a test  with a small probability of failing? :/

stardust-test
```diff
@@ -26,6 +26,8 @@ PATHS=(\
     "packages/dkg/..." \
     "packages/iscp" \
     "packages/webapi/..." \
+    "packages/util" \
+    "packages/util/pipe" \
```
jorgemmsilva:
```suggestion
    "packages/util/..." \
```
would this work?

jorgemmsilva (commented):
it looks fine to me overall, but the logic around the sync algorigthm is a bit hard to follow. Would you mind adding a small .md file with an high level description of how it should work? :pray:

jorgemmsilva (approved):

stardust-test
Juliusan: Not sure about all the packages. Will check.

packages/util/permute_test.go
Juliusan: IMO you cannot do better. Either you remove the test or you accept that it might theoretically fail. If you try to test that each time a random algorithm picks two different values (or permutations in this case), you have to accept that sometimes randomly same values are picked consecutively. Or you have to improve the random algorithm to ignore the previously picked value.

packages/chain/statemgr/action.go
Juliusan: I'll do the MD file. No. `startSyncingIfNeeded` simply creates structures to store blocks in. To avoid memory leaks these structures are later destroyed. `addBlockCandidate` adds a block to the structure, which must be created in advance.

PR #975, opened 2022-04-29
Cluster tests improvements

jorgemmsilva:
many updates to cluster tests, also includes a small fix for L1 message promotion

cluster tests status:

PASSING:
  deploy_test.go
  nodeconn_test.go
  privtangle_test.go
  post_test.go
  wal_test.go
  blob_test.go
  nanopublisher_test.go
  missing_requests_test.go
  account_test.go

HAS ERRORS:
  offledger_requests_test.go // access nodes not syncing - Julius looking into this
  transfer_test.go // missing a way to withdrawing EVERYTHING - we need to discuss better UX for deposit/withdrawals
  spam_test.go - onledger spam currently fails because the indexer returns older (already spent) outputs, offledger spam fails because requests start to be skipped based on nonces. - to be revisited (update L1 deps and review nonce logic)

TO BE LOOKED INTO:
  advanced_inccounter_test.go
  cluster_stability_test.go
  inccounter_solo_test.go
  inccounter_test.go
  jsonrpc_test.go
  wasp-cli-evm_test.go
  wasp-cli_test.go

### DIFF
--- a/packages/testutil/testmisc/misc.go
+++ b/packages/testutil/testmisc/misc.go
@@ -13,7 +13,7 @@ import (
 func RequireErrorToBe(t *testing.T, err error, target interface{}) {
 	if err == nil {
 		assert.Fail(t, "error expected, found nil")
-		t.FailNow()
+		t.Fatal()
 		return
 	}
 	if target, ok := target.(iscp.VMErrorBase); ok {
@@ -39,5 +39,5 @@ func RequireErrorToBe(t *testing.T, err error, target interface{}) {
 		return
 	}
 	assert.Fail(t, fmt.Sprintf("error does not contain '%s' but instead is '%v'", targ, err))
-	t.FailNow()
+	t.Fatal()
 }
--- a/tools/cluster/templates/waspconfig.go
+++ b/tools/cluster/templates/waspconfig.go
@@ -19,11 +19,11 @@ type WaspConfigParams struct {
 const WaspConfig = `
 {
   "database": {
-    "inMemory": false,
+    "inMemory": true,
     "directory": "waspdb"
   },
   "logger": {
-    "level": "debug",
+    "level": "info",
     "disableCaller": false,
     "disableStacktrace": true,
     "encoding": "console",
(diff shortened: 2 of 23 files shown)


packages/testutil/testmisc/misc.go
```diff
@@ -13,7 +13,7 @@ import (
 func RequireErrorToBe(t *testing.T, err error, target interface{}) {
 	if err == nil {
 		assert.Fail(t, "error expected, found nil")
-		t.FailNow()
+		t.Fatal()
```
dessaya, line 16: What's the difference? Why is Fatal better than FailNow? Just curious.

tools/cluster/templates/waspconfig.go
```diff
@@ -19,11 +19,11 @@ type WaspConfigParams struct {
 const WaspConfig = `
 {
   "database": {
-    "inMemory": true,
+    "inMemory": false,
```
dessaya: It was set to `true` because cluster instances are usually disposable, and to avoid thrashing the SSD with unneeded stuff every time you run the tests.
But maybe the correct way is to run the tests in a RAM partition or similar :thinking: 
Anyway, is there a reason you needed to change this? If inMemory is not working properly maybe we should remove the feature.

tools/cluster/templates/waspconfig.go
```diff
@@ -19,11 +19,11 @@ type WaspConfigParams struct {
...
     "directory": "waspdb"
   },
   "logger": {
-    "level": "warn",
+    "level": "debug",
```
dessaya: Same here, I think `warn` is a better default, thinking of the `wasp-cluster` tool. Maybe we should make this configurable and set it to debug only when running tests?

dessaya (approved):

packages/testutil/testmisc/misc.go
jorgemmsilva: Fatal tells us the line where it broke :hammer_and_wrench:

tools/cluster/templates/waspconfig.go
jorgemmsilva: Thank you, the changes to this file were just for testing. will revert them! :+1:

jorgemmsilva: `inMemory` needs to be false to run the offledger spam test properly, that's why I changed it. (the in-memory DB has a memory leak from what I remember, something about the `Commit` func - don't remember in particular, but it works for most tests)

PR #986, opened 2022-05-05
fix: `wasp-cli chain deploy` & dashboard

dessaya:
The following sequence now works:

1. Start a disposable privtangle + wasp cluster: `wasp-cluster start -d`

2. in some other console:

        wasp-cli init
        wasp-cli request-funds
        wasp-cli chain deploy --chain=chain1

3. Visit the dashboard at http://localhost:7000

### DIFF
--- a/tools/wasp-cli/config/config.go
+++ b/tools/wasp-cli/config/config.go
@@ -54,15 +54,27 @@ func Read() {
 }
 
 func L1Host() string {
-	return viper.GetString("l1.host")
+	host := viper.GetString("l1.host")
+	if host != "" {
+		return host
+	}
+	return "http://localhost"
 }
 
 func L1APIPort() int {
-	return viper.GetInt("l1.apiport")
+	port := viper.GetInt("l1.apiport")
+	if port != 0 {
+		return port
+	}
+	return 16500 // privtangle
 }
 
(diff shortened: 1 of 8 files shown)


BugFreeSoftware (approved):

tools/wasp-cli/config/config.go
```diff
@@ -54,15 +54,27 @@ func Read() {
...
+	return "http://localhost"
 }
 
 func L1APIPort() int {
-	return viper.GetInt("l1.apiport")
+	port := viper.GetInt("l1.apiport")
+	if port != 0 {
+		return port
+	}
+	return 16500 // privtangle
```
jorgemmsilva, line 69: this can get broken easily if we change the ports in privtangle (I kinda did that today when updating to the new L1 versions 😅). Would be nice if we got the ports from the privatangle package

jorgemmsilva (approved):

tools/wasp-cli/config/config.go
dessaya: Agree. Will fix later

dessaya: I just didn't want to add the privtangle dependency to wasp-cli, but maybe it is not so bad

PR #993, opened 2022-05-06
Fixed 2 bugs in Allowance.SpendFromBudget()

BugFreeSoftware:


packages/iscp/allowance.go
```diff
@@ -53,7 +53,12 @@ func (a *Allowance) Clone() *Allowance {
 }
 
 func (a *Allowance) SpendFromBudget(toSpend *Allowance) bool {
-	a.Assets.SpendFromFungibleTokenBudget(toSpend.Assets)
+	if a.IsEmpty() {
+		return toSpend.IsEmpty()
+	}
+	if !a.Assets.SpendFromFungibleTokenBudget(toSpend.Assets) {
+		return false
+	}
```
jorgemmsilva, line 61: suggestion: we could eliminate the empty check (that is duplicated from the code inside `Assets.SpendFromFungibleTokenBudget`) and write it like this:
```
	if ok := a.Assets.SpendFromFungibleTokenBudget(toSpend.Assets); !ok {
		return false
	}
```
in cases like this we should also add a test that reproduces the bug

PR #1000, opened 2022-05-11
Fix: stateaccess / kv len / harvest

jorgemmsilva:


packages/kv/collections/map.go
```diff
@@ -95,7 +95,11 @@ func (m *Map) SetAt(key, value []byte) error {
 	if err != nil {
 		return err
 	}
-	if !ok {
+	length, err := m.Len()
+	if err != nil {
+		return err
+	}
+	if !ok || length == 0 { // if the map is empty, setting some value to the `nil` key will return ok = true, but that is not correct
```
dessaya, line 102: I don't understand this. Who returns ok = true? `HasAt(nil)`? Why? :(

dessaya (approved):

packages/kv/collections/map.go
jorgemmsilva: yes. I don't know. These changes handle that edge case but I have no idea why that happens, I'm not too familiar with the design philosophy behind the kv package. Maybe @lunfardo314 knows why?

jorgemmsilva:
closing in favor of https://github.com/iotaledger/wasp/pull/1016

PR #1012, opened 2022-05-16
Refactor: L1 update

jorgemmsilva:
updates to support all the new versions of iotago/hornet & inx plugins

many breaking changes were made in iotago addressed in this PR, multiple fixes related to the way storage deposit is calculated.


contracts/wasm/testcore/src/testcore.rs
```diff
@@ -277,11 +277,11 @@ pub fn func_send_nf_ts_back(ctx: &ScFuncContext, _f: &SendNFTsBackContext) {
 pub fn func_split_funds(ctx: &ScFuncContext, _f: &SplitFundsContext) {
     let mut iotas = ctx.allowance().iotas();
     let address = ctx.caller().address();
-    let transfer = wasmlib::ScTransfer::iotas(200);
-    while iotas >= 200 {
+    let transfer = wasmlib::ScTransfer::iotas(1000000);
```
lmoe, line 280: Should we add the `iscp.Mi` const to the wasmlib?

lmoe (approved):
Amazing work!

go.mod
```diff
@@ -1,82 +1,205 @@
...
-	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
-	google.golang.org/genproto v0.0.0-20220329172620-7be39ac1afc7 // indirect
+	golang.org/x/crypto v0.0.0-20220507011949-2cf3adece122
+	golang.org/x/xerrors v0.0.0-20220411194840-2f41105eb62f
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
-	lukechampine.com/blake3 v1.1.7 // indirect
 	nhooyr.io/websocket v1.8.7
 )
 
+require (
```
dessaya, line 43: Why two require blocks? :thinking:

contracts/wasm/testwasmlib/go/testwasmlib/testwasmlib.go
```diff
@@ -6,11 +6,17 @@ package testwasmlib
 import (
 	"bytes"
 
+	"github.com/iotaledger/iota.go/v3/tpkg"
 	"github.com/iotaledger/wasp/packages/wasmvm/wasmlib/go/wasmlib"
 	"github.com/iotaledger/wasp/packages/wasmvm/wasmlib/go/wasmlib/coreblocklog"
 	"github.com/iotaledger/wasp/packages/wasmvm/wasmlib/go/wasmlib/wasmtypes"
 )
 
+var (
```
BugFreeSoftware: This won't work in Wasm environment. We don't want Wasp/iotago dependencies. They can introduce horrible dependency chains, include stuff that isn't supported by TinyGo, include stuff that isn't deterministic, and we don't have the corresponding code for Rust and Typescript, so those versions of the SCs will now fail when I test them.
Please revert the lines in testwasmlib.go and testwasmlib_test.go that were changed to use TestNFTID and TestTokenID.

dessaya (approved):

BugFreeSoftware (changes requested):

contracts/wasm/testcore/src/testcore.rs
BugFreeSoftware: Will look into that after the merge. There is a design decision around dust in the Wasm code that I need to revisit after these changes to make them consistent again.

go.mod
jorgemmsilva: new go things :sparkles: 

https://stackoverflow.com/a/69433550/6749639

contracts/wasm/testwasmlib/go/testwasmlib/testwasmlib.go
jorgemmsilva: sure things, will remove the `iota.go` import. :+1: 
On another note, I'm thinking it would be nice if we could trim it down and import the `iscp` package in tinygo. It should be quite minimalistic and only export simple types. 

`iota.go` should probably work fine in tinyGo actually. Did you try it out? is there anything that is breaking there?

PR #1020, opened 2022-05-19
Removed unused networkprefix parameters and deprecated function

BugFreeSoftware:


packages/iscp/requestimpl.go
```diff
@@ -425,7 +425,8 @@ func (r *OnLedgerRequestData) SenderAccount() AgentID {
 	if r.SenderAddress() == nil || r.requestMetadata == nil {
 		return nil
 	}
-	return NewAgentIDFromAddressAndHname(r.SenderAddress(), r.requestMetadata.SenderContract)
+	chainID := ChainIDFromAddress(r.SenderAddress().(*iotago.AliasAddress))
```
dessaya: I don't think it is guaranteed that the SenderAddress is an AliasAddress here, or is it?

BugFreeSoftware: Yeah, had to change it

jorgemmsilva (approved):

PR #1036, opened 2022-05-24
Fix failing cluster test

fijter:
# Description of change

Made a mistake and the cluster tests stopped working, fixed it and manually ran the cluster tests to verify it works now


fijter:
Never mind, @jorgemmsilva is faster than light :joy:

PR #1042, opened 2022-05-30
Stardust vm julius

Juliusan:
A bugfix for random validator picking and some logging updates.


packages/nodeconn/nodeconn.go
```diff
@@ -225,7 +226,22 @@ func (nc *nodeConn) doPostTx(ctx context.Context, tx *iotago.Transaction) (*iota
...
+		} else {
+			inIDs[i] = fmt.Sprintf("%v=%T", input.Type(), input)
+		}
+	}
+	txID, err := tx.ID()
+	if err == nil {
+		nc.log.Debugf("Posted transaction id %v, inputs: %v output count: %v", iscp.TxID(txID), inIDs, len(tx.Essence.Outputs))
+	} else {
+		nc.log.Debugf("Posted transaction, inputs: %v output count: %v", inIDs, len(tx.Essence.Outputs))
+	}
```
jorgemmsilva: does it make sense to keep this log?

jorgemmsilva (approved):

packages/nodeconn/nodeconn.go
Juliusan: It is certainly no sense in logging the transaction itself. We need to log the transaction ID. I found it helpful to know the exact input id's of the transaction, maybe its worth keeping it? Or I can delete it and leave only transaction ID in logs.

jorgemmsilva: isn't what we need logged in `waitUntilConfirmed`?

Juliusan: `waitUntilConfirmed` doesn't log transaction ID neither it does log input IDs.

jorgemmsilva: it logs the msg( now named block) ID:
https://github.com/iotaledger/wasp/blob/stardust-vm/packages/nodeconn/nodeconn.go#L237-L240
if that's not enough, sure lets log the txID here

ivange94 (approved):

packages/nodeconn/nodeconn.go
Juliusan: Ok, I'll log the txID and remove everything else.

PR #1050, opened 2022-06-02
Misc improvements in EVM & `iscp/request`

dessaya:
Highlights:

* New `OffLedgerRequest` interface: now it is mandatory to call `Sign()` for the request to be valid (statically validated).
* EVM request is now an independent implementation of the `OffLedgerRequest` interface, with no way to specify allowance, gas budget, etc (i.e., no way to construct an inconsistent request).
* Gas for "bookkeeping" (e.g. creating the EVM "block") is no longer charged.

### DIFF
--- a/packages/iscp/request_evm.go
+++ b/packages/iscp/request_evm.go
@@ -1,113 +1,246 @@
 package iscp
 
 import (
-	"math"
+	"fmt"
 
 	"github.com/ethereum/go-ethereum"
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/iotaledger/hive.go/marshalutil"
-	"github.com/iotaledger/wasp/packages/cryptolib"
+	iotago "github.com/iotaledger/iota.go/v3"
 	"github.com/iotaledger/wasp/packages/evm/evmtypes"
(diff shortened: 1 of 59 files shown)


packages/iscp/request_evm.go
```diff
@@ -1,112 +1,253 @@
...
-	return &OffLedgerRequestData{
-		chainID:         chainID,
-		contract:        Hn(evmnames.Contract),
-		entryPoint:      Hn(evmnames.FuncSendTransaction),
-		params:          dict.Dict{evmnames.FieldTransaction: evmtypes.EncodeTransaction(tx)},
-		signatureScheme: signatureScheme,
-		nonce:           tx.Nonce(),
-		allowance:       NewEmptyAllowance(),
-		gasBudget:       evmtypes.EVMGasToISC(tx.Gas(), gasRatio) + EVMGasBookkeeping,
+	// TODO: verify tx.ChainId()?
```
jorgemmsilva, line 31: that could be done in the JSON rpc endpoint

jorgemmsilva (approved):

packages/iscp/request_evm.go
dessaya: That's right. I will do that in my next PR :+1:

PR #1057, opened 2022-06-07
BugFix of a BugFix: don't close files too early

Juliusan:
Due to mix up between `segmentFile` and regular file, files were closed too early. Simplified the code while fixing the bug.

### DIFF
--- a/packages/wal/wal.go
+++ b/packages/wal/wal.go
@@ -3,7 +3,6 @@ package wal
 import (
 	"bufio"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -36,15 +35,7 @@ func New(log *logger.Logger, dir string) *WAL {
 
 var _ chain.WAL = &chainWAL{}
 
-type segmentFile interface {
-	Stat() (os.FileInfo, error)
-	io.Writer
-	io.Closer
-	io.Reader
-}
-
 type segment struct {
-	segmentFile
 	index uint32
 	dir   string
 }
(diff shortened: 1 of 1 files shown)


jorgemmsilva (approved):

packages/wal/wal.go
```diff
@@ -82,30 +73,26 @@ func (w *chainWAL) Write(bytes []byte) error {
...
-	segName := segmentName(w.dir, i)
-	f, err := os.OpenFile(segName, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o666)
-	defer f.Close()
-	if err != nil {
-		return nil, fmt.Errorf("could not create segment: %w", err)
-	}
-	s := &segment{index: i, segmentFile: f, dir: w.dir}
-	w.segments[i] = s
-	return s, nil
+	return f.Close()
```
ivange94: If there's an error above, this will never be called hence leaving the file open. Am I wrong here?

Juliusan: Well spotted. This was how I found it originally, but yes, I'll change it to `defer`. Thanks

ivange94 (approved):

packages/wal/wal.go
```diff
@@ -82,30 +73,27 @@ func (w *chainWAL) Write(bytes []byte) error {
 	if err != nil {
 		return fmt.Errorf("Invalid block: %w", err)
 	}
-	segment, err := w.createSegment(block.BlockIndex())
+
+	index := block.BlockIndex()
+	segName := segmentName(w.dir, index)
+	f, err := os.OpenFile(segName, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o666)
+	defer f.Close()
```
dessaya, line 80: if `err != nil` then `f` is nil, I think? So `f.Close()` will panic.

PR #1061, opened 2022-06-07
Close only a file was opened

Juliusan:
Final BugFix of a BugFix, I hope, after Diego's comment.

### DIFF
--- a/packages/wal/wal.go
+++ b/packages/wal/wal.go
@@ -50,9 +50,10 @@ func (w *WAL) NewChainWAL(chainID *iscp.ChainID) (chain.WAL, error) {
 	}
 	// read all segments in log
 	f, err := os.Open(w.dir)
-	defer f.Close()
 	if err != nil {
 		return nil, fmt.Errorf("could not open wal: %w", err)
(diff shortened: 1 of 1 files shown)


packages/wal/wal.go
```diff
@@ -50,9 +50,10 @@ func (w *WAL) NewChainWAL(chainID *iscp.ChainID) (chain.WAL, error) {
 	}
 	// read all segments in log
 	f, err := os.Open(w.dir)
-	defer f.Close()
 	if err != nil {
 		return nil, fmt.Errorf("could not open wal: %w", err)
+	} else {
+		defer f.Close()
 	}
```
jorgemmsilva: we can ditch the "elses"
```suggestion
	} 
	defer f.Close()
```

jorgemmsilva (approved):

dessaya (approved):

packages/wal/wal.go
Juliusan: I hate that. But that's how Go does it. Will change :-)

PR #1066, opened 2022-06-08
Misc improvements

dessaya:
- dashboard: fix overflow & chainblock
- wasp-cli: it is now possible to deposit L1 funds into a L2 ethereum account with: `wasp-cli chain deposit 0x34Cc8... iota:1000000`
- Fix several wasp-cli commands that were not working
- Fix most of cluster wasp-cli tests

### DIFF
--- a/tools/cluster/tests/wasp-cli_test.go
+++ b/tools/cluster/tests/wasp-cli_test.go
@@ -87,6 +87,38 @@ func TestWaspCLI1Chain(t *testing.T) {
 	require.Len(t, out, 1)
 }
 
+func TestWaspCLIDeposit(t *testing.T) {
+	w := newWaspCLITest(t)
+
+	committee, quorum := w.CommitteeConfig()
+	w.Run("chain", "deploy", "--chain=chain1", committee, quorum)
+
+	checkBalance := func(out []string, expected int) {
+		amount := 0
+		for _, line := range out {
+			r := regexp.MustCompile(`(?m)iota\s+(\d+)`).FindStringSubmatch(line)
+			if r != nil {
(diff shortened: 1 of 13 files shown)


tools/cluster/tests/wasp-cli_test.go
```diff
@@ -87,14 +87,46 @@ func TestWaspCLI1Chain(t *testing.T) {
...
+
 func TestWaspCLIContract(t *testing.T) {
 	w := newWaspCLITest(t)
 
 	committee, quorum := w.CommitteeConfig()
 	w.Run("chain", "deploy", "--chain=chain1", committee, quorum)
 
 	// for running off-ledger requests
-	w.Run("chain", "deposit", "IOTA:10")
+	w.Run("chain", "deposit", "iota:10")
```
jorgemmsilva, line 129: I know this is old code, but depositing 10 iotas shouldn't be possible :thinking: 
This is probably adjusted to the minimum storage deposit inside

jorgemmsilva (approved):

tools/cluster/tests/wasp-cli_test.go
jorgemmsilva: on a second thought, transfers from the cli are probably broken rn, @lmoe said he was going to work on that.

dessaya: It is automatically adjusted, so in the end it deposits something like 100k.

jorgemmsilva: yeah, I figured 👍 . This way the tests are weird to read, tho 😢

PR #1069, opened 2022-06-08
Add TestCore solidity contract

ivange94:
# Description of change

Implementing TestCore contract in solidity and comparing gas usage with wasmvm.


packages/vm/core/evm/evmtest/testcore_test.go
```diff
@@ -0,0 +1,70 @@
...
+	evn := initEVM(t)
+	ethKey, _ := evn.soloChain.NewEthereumAccountWithL2Funds()
+	testCore := evn.deployTestCoreContract(ethKey)
+	require.EqualValues(t, 1, evn.getBlockNumber())
+
+	res, err := testCore.fibonacci(N)
+	require.NoError(t, err)
+	t.Log("evm gas used:", res.evmReceipt.GasUsed)
+	t.Log("iscp gas used:", res.iscpReceipt.GasBurned)
+	t.Log("iscp gas fee:", res.iscpReceipt.GasFeeCharged)
```
jorgemmsilva, line 28: I think the valuable think here would be to have the execution of equivalent code in WASM VM, so we can compare the gas

dessaya (approved):

packages/vm/core/evm/evmtest/testcore_test.go
ivange94: @jorgemmsilva I was hoping to add those in a separate pr but I just pushed a commit with that. The only files from `wasm` that need reviewing is `testcore.go/ts/rs` and `call_test.go`. The rest is auto generated.

ivange94:
Closing this as we decided not to use fibonacci at this stage for the tests. Will create another with the proper contracts.

PR #1075, opened 2022-06-10
Permutations

Juliusan:
# Description of change

Refactored permutation generation. Previously, seed was used in every call to `Shuffle` function. This caused a problem for the caller to generate a random seed for each call. In one occasion, time was used for that, which probably caused a problem for Windows users [1]. Moreover `Shuffle` function is already implemented in standard go libraries, so there is no need to reinvent it. The refactored solution may be used without passing any seed. In that case, cryptographically secure random seed is generated on permutation creation. However there is an option to create permutation with a fixed seed. This is useful in tests to get more predictable results and in consensus which needs a random delay for each node for transaction posted, which should be consistent between all the nodes. Moreover, each `Shuffle` call doesn't require a seed, as the source obtained from the seed, which was used on permutation creation is used.

[1] Eric's problem with `TestNextNoCycles` in Slack (https://iotafoundation.slack.com/archives/C02M65QLECW/p1654826351871749)

## Type of change

All the short tests passed as well as some of the cluster tests. I also tested `math/rand` package to make sure that it acts the same on different environments. The same code has been run on two Linux and one Windows machine running Go18.

## Change checklist

- [+] I have followed the [contribution guidelines](https://wiki.iota.org/smart-contracts/contribute) for this project
- [+] I have performed a self-review of my own code
- [+] I have selected the `develop` branch as the target branch
- [+] I have commented my code, particularly in hard-to-understand areas
- [+] I have made corresponding changes to the documentation
- [+] I have added tests that prove my fix is effective or that my feature works
- [+] I have checked that new and existing unit tests pass locally with my changes

### DIFF
--- a/packages/chain/committee/committee.go
+++ b/packages/chain/committee/committee.go
@@ -4,7 +4,6 @@
 package committee
 
 import (
-	"crypto/rand"
 	"time"
 
 	"github.com/iotaledger/hive.go/logger"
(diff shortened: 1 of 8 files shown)


BugFreeSoftware (approved):

dessaya (approved):

BugFreeSoftware:
I merged this one already in my wasmdust branch to fix the errors I had before. 
I also linted it properly to fix the github checks that failed here (had to install latest golangci-lint), so it is now part of PR 1077. Therefore I am closing this one.

PR #1083, opened 2022-06-21
feat: reject request outputs with SDRC

jorgemmsilva:


dessaya (approved):

ivange94 (approved):
LGTM

packages/nodeconn/nc_chain.go
```diff
@@ -209,7 +214,9 @@ func (ncc *ncChain) subscribeToChainOwnedUTXOs() {
 				}
 				outID := iotago.OutputIDFromTransactionIDAndIndex(*tid, outResponse.Metadata.OutputIndex)
 				ncc.log.Debugf("received UTXO, outputID: %s", outID.ToHex())
-				ncc.outputHandler(outID, out)
+				if shouldBeProcessed(out) {
```
ivange94, line 217: I feel readability will be better if this function is renamed to reflect what it actually does. Right now you'll have to scroll up to the definition of `shouldBeProcessed` to fully understand what happens here.

ivange94: Something like

```
if !outputHasDepositStorageReturnCondition(out) {
    ncc.outputHandler(outID, out)
}
```
or 
```
if outputDoesNotHaveDepositStorageReturnCondition(out) {
    ncc.outputHandler(outID, out)
}
```
Its much longer but self explanatory.

jorgemmsilva: Thanks for the comment @ivange94. I'm going to merge as-is, because this is mostly a matter of personal preference.

it could literally be written as:

```go
if !out.UnlockConditionSet().HasStorageDepositReturnCondition() {
    ncc.outputHandler(outID, out)
}
```

I opted to create another function to make it more readable (`if shouldBeProcessed(out)`) and easy to extend.
if we need more rules to dictate whether or not a request should be processed, this is very easy to extend:

```go
func shouldBeProcessed(out iotago.Output) bool {
	// only outputs without SDRC should be processed.
	return !out.UnlockConditionSet().HasStorageDepositReturnCondition() ||
		out.Something() ||
		out.SomethiingElse()
}
```

PR #1086, opened 2022-06-22
Rotation fixes

Juliusan:
# Description of change

Self governed chained rotation now works

## Type of change

Short and normal tests were run. All the short tests pass. Among the normal tests, `TestDepositWithdraw`, `TestWaspCLIMint`, `TestWaspCLIBalance` and `TestWaspCLIRejoinChain` cluster tests fail, but they were failing before these changes too. `TestBlobStoreManyBlobsNoEncoding`, `TestSuccessfulIncCounterIncreaseWithoutInstability` and `TestSuccessfulConsenseusWithReconnectingNodes` cluster tests experienced some random failures, but in general they pass. Passing all the tests using `make test` took  me more than 32 minutes, so I also increased the timeout.

## Change checklist


client/chainclient/chainclient.go
```diff
@@ -60,18 +60,59 @@ func (c *Client) Post1Request(
...
 ) (*iotago.Transaction, error) {
-	par := defaultParams(params...)
+	outputsSet, err := c.Layer1Client.OutputMap(c.KeyPair.Address())
+	if err != nil {
+		return nil, err
+	}
+	return c.post1RequestWithOutputs(contractHname, entryPoint, outputsSet, params...)
+}
+
+// Post1Request sends n consecutive on-ledger transactions with one request on each to the chain
```
jorgemmsilva:
```suggestion
// PostNRequests sends n consecutive on-ledger transactions with one request on each, to the chain
```

packages/chain/chainimpl/eventproc.go
```diff
@@ -115,93 +115,81 @@ func (c *chainObj) handleAliasOutput(msg *iscp.AliasOutputWithID) {
...
 				return
 			}
 		}
 		c.lastSeenOutputStateIndex = &msgStateIndex
+		if msgStateIndex != 0 && isRotation {
+			c.processChainTransition(&chain.ChainTransitionEventData{
+				IsGovernance:    true,
+				VirtualState:    c.lastSeenVirtualState,
+				ChainOutput:     msg,
+				OutputTimestamp: time.Now(),
```
jorgemmsilva, line 145: shouldn't `OutputTimestamp` be the timestamp from the latest milestone?

packages/iscp/rotate/rotate.go
```diff
@@ -28,28 +29,27 @@ func NewRotateRequestOffLedger(chainID *iscp.ChainID, newStateAddress iotago.Add
...
-	// }
-	// return essence, nil
+	output := chainInput.GetAliasOutput().Clone().(*iotago.AliasOutput)
+	for i := range output.Conditions {
+		if _, ok := output.Conditions[i].(*iotago.StateControllerAddressUnlockCondition); ok {
+			output.Conditions[i] = &iotago.StateControllerAddressUnlockCondition{Address: nextAddr}
+		}
+		// TODO: it is probably not the correct way to do the governance transition
+		if _, ok := output.Conditions[i].(*iotago.GovernorAddressUnlockCondition); ok {
+			output.Conditions[i] = &iotago.GovernorAddressUnlockCondition{Address: nextAddr}
```
jorgemmsilva, line 43: should we change the governor address? isn't it supposed to be the chain already? otherwise the transition would fail.

tools/cluster/cluster.go
```diff
@@ -226,6 +226,33 @@ func (clu *Cluster) addAllAccessNodes(chain *Chain, nodes []int) error {
...
+		}
+		accessNodePubKey, err := cryptolib.NewPublicKeyFromString(accessNodePeering.PubKey)
+		if err != nil {
+			return err
+		}
+		scArgs.Accept(accessNodePubKey)
+	}
+	scParams := chainclient.
+		NewPostRequestParams(scArgs.AsDict()).
+		WithIotas(1000)
```
jorgemmsilva:
```suggestion
		WithIotas(1*iscp.Mi)
```
1000 seems like not very much

packages/nodeconn/nodeconn.go
```diff
@@ -146,8 +146,20 @@ func (nc *nodeConn) UnregisterChain(chainID *iscp.ChainID) {
...
+	ncc, ok := nc.chains[chainID.Key()]
+	nc.chainsLock.RUnlock()
+	if !ok {
+		return xerrors.Errorf("Chain %v is not connected.", chainID.String())
+	}
+	return ncc.PublishTransaction(tx)
+}
+
+// PublishGovernanceTransaction implements chain.NodeConnection.
+// TODO: identical to PublishStateTransaction; needs to be reviewed
```
jorgemmsilva, line 161: do we need 2 different funcs?

jorgemmsilva (approved):

packages/nodeconn/nodeconn.go
kape1395: I would vote for 2 different functions, because they should behave differently in the future. The state transactions will have to support pipelining (and then the consecutive state numbering will be important, especially in the case of rejected transactions). That's not the case for the governance tx'es.

packages/iscp/rotate/rotate.go
Juliusan: Is there a way to do it? Governor address is first introduced in the origin state output. It cannot be chain alias ID, as it is made from origin output ID, which in turn is made from origin transaction ID and index 0. And transaction ID is calculated from marshalled transaction, which contains origin state output. Thus currently governor and state controller address are the same in the origin state output. Keeping this condition allowed tests to pass, although it is not optimal.

packages/chain/chainimpl/eventproc.go
Juliusan: In fact, it doesn't matter. It is used only to delay batch proposal for the next state. State manager sets it similarly for state transaction. And `chainimpl` doesn't keep reference to the latest milestone (only `consensus` does).

packages/iscp/rotate/rotate.go
jorgemmsilva: Maybe I'm wrong, but only the alias gov controller can change the governing address, right?
If the chain is not the gov controller already wouldn't this fail?

Juliusan: I am not sure, how it is supposed to be, but this way it works. E.g., `TestRotationMany` makes several consecutive rotations and if you keep state controller and governance controller addresses to be equal to the address of the current committee, then it passes.

jorgemmsilva: I just find it weird that the first rotation works, if the chain is not the governing identity already :thinking:

client/scclient/postrequest.go
```diff
@@ -10,6 +10,10 @@ func (c *SCClient) PostRequest(fname string, params ...chainclient.PostRequestPa
 	return c.ChainClient.Post1Request(c.ContractHname, iscp.Hn(fname), params...)
 }
 
+func (c *SCClient) PostRequests(fname string, n int, params ...chainclient.PostRequestParams) ([]*iotago.Transaction, error) {
```
dessaya: Nit: I would call this `PostNRequests`, for consistency.

(Although tbh I never liked the `Post1Request`, `PostNRequests` names... can't we just have `PostRequest` and if you need to call that N times just place it inside a loop?)

packages/chain/consensus/action.go
```diff
@@ -348,29 +348,29 @@ func (c *consensus) checkQuorum() { //nolint:funlen
...
+			block.SetApprovingOutputID(chainOutputID)
+			err = c.wal.Write(block.Bytes())
+			if err == nil {
+				c.log.Debugf("checkQuorum: block index %v written to wal", block.BlockIndex())
+			} else {
+				c.log.Warnf("checkQuorum: error writing block to wal: %v", err)
+			}
 		} else {
-			c.log.Warnf("checkQuorum: error writing block to wal: %v", err)
+			c.log.Warnf("checkQuorum: skipping writing block to was: error extracting block from state: %v", err)
```
dessaya: `wal` -> `was` -- is that change intentional?

tools/wasp-cli/metrics/nodeconn.go
```diff
@@ -43,17 +43,18 @@ var nodeconnMetricsCmd = &cobra.Command{
 func printMessagesMetrics(msgsMetrics *model.NodeConnectionMessagesMetrics, additionalRows [][]string) {
 	header := []string{"Message name", "", "Total", "Last time", "Last message"}
 	table := make([][]string, 9+len(additionalRows))
```
dessaya: shouldn't this be `10+len(additionalRows)`?

also, maybe this is better?

```go
table := [][]string{
    makeMessagesMetricsTableRow("Publish transaction", false, msgsMetrics.OutPublishTransaction),
    makeMessagesMetricsTableRow("Pull latest output", false, msgsMetrics.OutPullLatestOutput),
    ...
}
table = append(table, additionalRows...)
log.PrintTable(header, table)
```

dessaya (approved):

client/scclient/postrequest.go
Juliusan: Well... `Post1Request` obtains unused outputs from L1 and produces a transaction out of them. `PostNRequests` obtains outputs from L1, produces a transaction, and then uses outputs from this transaction to produce the second transaction. Not using L1 as a source for outputs (which are sometimes outdated) is the main reason I introduced `PostNRequests`. So it is not exactly the same as calling `Post1Request` N times.

PR #1094, opened 2022-06-27
Add stronghold & native keychain

lmoe:
# Description of change

This PR addresses the unsafe seed storage inside the wasp-cli config.

It introduces two new layers (Keychain and Stronghold). 

The user can decide which to choose by setting `wallet.scheme` to either `plain` or `stronghold`. For now `plain` is the default setting.

In `plain` mode, the seed will be saved inside the OS keychain. All major OSes are supported. This requires the entry of a password at the start - either by prompt or by setting the env variable `STORE_PASSWORD`. The session stays open for a while, so repeated wasp-cli commands do not require a reentry. 

If no keychain is available, a fallback [encrypted json file] store will be used. 

`stronghold` mode will initialize a complete Stronghold vault, the encryption key gets saved into the Keychain. 
Therefore, it's a layer on top of the keychain.

This scheme works, but should be seen as experimental for now.  

Furthermore, secrets are stored in a security enclave that hides the secrets that are stored in memory to prevent ReadProcessMemory attacks. This should be revisited at some time though. 

## Type of change

* make test


tools/wasp-cli/config/config.go
```diff
@@ -39,7 +42,15 @@ const (
...
 )
 
+const StorePasswordEnvKey string = "STORE_PASSWORD"
+
 func Init(rootCmd *cobra.Command) {
+	viper.AutomaticEnv()
+
+	var storePassword string
+	rootCmd.PersistentFlags().StringVarP(&storePassword, "store-password", "s", viper.GetString(StorePasswordEnvKey), "Password to open [file] store")
+	StorePassword = memguard.NewEnclave([]byte(storePassword))
```
dessaya: Is this correct? I think the flags may be actually parsed from the command line some time after this, so `storePassword` is always empty at this point?

tools/wasp-cli/config/config.go
```diff
@@ -49,7 +60,15 @@ func Init(rootCmd *cobra.Command) {
...
 func Read() {
 	viper.SetConfigFile(ConfigPath)
+	viper.SetDefault("wallet.scheme", WalletDefaultScheme)
+
 	_ = viper.ReadInConfig()
+
+	Store = NewSecureStore()
+	err := Store.Open()
+	if err != nil {
+		fmt.Println(err)
```
dessaya: Is this a fatal error? Maybe use `log.Check(err)` instead?

tools/wasp-cli/config/config.go
```diff
@@ -103,6 +122,33 @@ func WaspClient() *client.WaspClient {
...
+func WalletScheme() string {
+	scheme := viper.GetString("wallet.scheme")
+
+	switch scheme {
+	case WalletSchemePlain:
+		return WalletSchemePlain
+	case WalletSchemeStronghold:
+		return WalletSchemeStronghold
+	default:
+		return WalletDefaultScheme
```
dessaya: I think it's better to panic instead? `log.Fatalf("Unrecognized wallet scheme: %s", scheme)`

tools/wasp-cli/config/config.go
```diff
@@ -49,7 +60,15 @@ func Init(rootCmd *cobra.Command) {
 
 func Read() {
 	viper.SetConfigFile(ConfigPath)
+	viper.SetDefault("wallet.scheme", WalletDefaultScheme)
```
dessaya: Nit:

`const walletSchemeKey = "wallet.scheme"`

(because it is used twice)

tools/wasp-cli/store/info.go
```diff
@@ -0,0 +1,39 @@
...
+var resetCmd = &cobra.Command{
+	Use:   "reset",
+	Short: "Reset store",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		err := config.Store.Reset()
+
+		log.Check(err)
+
+		log.Printf("Store resetted!")
```
dessaya: I _think_ the past tense of "reset" is just "reset". So `Store has been reset` or similar.

tools/wasp-cli/wallet/wallet.go
```diff
@@ -7,53 +7,89 @@ import (
...
+		} else {
+			err = config.Store.InitializeNewStronghold()
+		}
 
-		log.Printf("Initialized wallet seed in %s\n", config.ConfigPath)
+		log.Check(err)
+
+		log.Printf("Initialized wallet seed, saved in key chain.\n")
 		log.Printf("\nIMPORTANT: wasp-cli is alpha phase. The seed is currently being stored " +
 			"in a plain text file which is NOT secure. Do not use this seed to store funds " +
```
dessaya: Is the `plain text` part still true?

dessaya (approved):

packages/cryptolib/stronghold_keypair.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (kp *StrongholdKeyPair) Address() *iotago.Ed25519Address {
+	address, _ := kp.stronghold.GetAddress(kp.addressIndex)
+	return (*iotago.Ed25519Address)(&address)
+}
+
+func (kp *StrongholdKeyPair) Sign(data []byte) []byte {
+	recordPath := fmt.Sprintf("seed.%d", kp.addressIndex)
+	signature, _ := kp.stronghold.Sign(recordPath, data)
```
jorgemmsilva: there are some `, _ := kp.stronghold....` I assume those are errors not checked? - should we check them?

lmoe: Good catch. I wasn't sure what to do with it and wanted to work around this, but that fell under the table it seems.

I'm unsure what the best way would be. The Keypair type was written without error reporting, because the used ed25519 lib doesn't provide any as well.

The Stronghold lib might return errors. I haven't seen any yet, but it might. If we change the Keypair and interface to return errors, this would result in hefty refactoring throughout the whole project. Not just in usage, but also in other functions that use the Keypair, and then return no errors. Panicing seems like the wrong approach, just logging as well..

I think the cleanest solution would be to add error returns in the interface and do the refactoring. But that will take some time. 

What do you think?

jorgemmsilva: >I think the cleanest solution would be to add error returns the the interface and do the refactoring. But that will take some time.

I tend to agree with this.

however, it could be worth checking what kind of errors `stronghold` can return - maybe panicking is correct, if the errors are caused by something that should never ever happen

lmoe: It varies from actual non handleable errors like missing the libstronghold binary (this would cause a panic in the CGO layer directly) to an uninitialized wallet error for example. 

If the Stronghold vault is not initialized correctly you are unable to sign data, as you are missing a master key and a derived key.  

Those things. I think very few would be actually "handable" and would leave the app in a weird state.

I've just spend a bit of time to check how much work this refactoring would be. The keyPair is often used directly in assignments. For example in `utxodb.go`:`genesisAddress` or in `accessnodes.go`:`NewNodeOwnershipCertificate` if you change one function to return the error, 5 new functions pop up you need to refactor. :shrug: 

If it would bring some value I would just dig through it. Otherwise, maybe panicing the errors in the Stronghold Keypair implementation might be enough for now?

jorgemmsilva: panic seems fine to me :+1:

jorgemmsilva (approved):

lmoe:
PR will be separated into two parts. First the secure storage will be opened up. I will close this PR.

PR #1096, opened 2022-06-29
Stardust vm julius

Juliusan:
# Description of change

An additional test for self governed chain rotation, which allowed fixing of several subtle bugs in state manager/consensus. As a side effect - random peer picking algorithm fixed for it not to return the same peer twice and seeding is now really taken into account in permutation.

## Type of change

- Documentation Fix

All the unit and short cluster tests pass. I'll run regular cluster tests and fix if some errors arise. There are some known problems, which occur randomly, which I'll detail in Slack.

## Change checklist

### DIFF
--- a/packages/chain/consensus/action.go
+++ b/packages/chain/consensus/action.go
@@ -718,17 +718,26 @@ func (c *consensus) setNewState(msg *messages.StateTransitionMsg) bool {
 		return false
 	}
 
-	c.stateOutput = msg.StateOutput
+	// If c.stateOutput.GetStateIndex() == msg.StateOutput.GetStateIndex() and the new state output is not a governance update, then either,
+	// a) c.stateOutput is the same as msg.StateOutput and there is no need to reasign c.stateOutput or b) msg.StateOutput is a regular state update
+	// output and c.stateOutput is a governance update output with the same index; in such case governance update is the last and should be taken
+	// into account. Regular state output is overwriden by governance uptade output and should be ignored.
+	// TODO: it is assumed, that at most one governance update may occur in between regural state updates. The situation of several consecutive
+	// governance updates must be yet to be discussed, designed and implemented. The main problem is that there is no way in knowing the exact order
+	// of governance updates, which have the same block index.
+	if (c.stateOutput == nil) || (c.stateOutput.GetStateIndex() < msg.StateOutput.GetStateIndex()) || msg.IsGovernance {
+		c.stateOutput = msg.StateOutput
+	} else {
+		c.log.Debugf("consensus::setNewState: ignoring the received state output %s in favor of the current one %s", iscp.OID(msg.StateOutput.ID()), iscp.OID(c.stateOutput.ID()))
+	}
 	c.stateTimestamp = msg.StateTimestamp
-	oid := msg.StateOutput.OutputID()
+	oid := c.stateOutput.OutputID()
+	c.acsSessionID = util.MustUint64From8Bytes(hashing.HashData(oid[:]).Bytes()[:8])
 	if msg.IsGovernance && !sameIndex {
 		c.currentState = nil
 		c.log.Debugf("SET NEW STATE #%d (rotate) and pausing consensus to wait for adequate state, output: %s",
 			c.stateOutput.GetStateIndex(), iscp.OID(c.stateOutput.ID()))
 	} else {
-		if !msg.IsGovernance {
-			c.acsSessionID = util.MustUint64From8Bytes(hashing.HashData(oid[:]).Bytes()[:8])
-		}
 		c.currentState = msg.State
 		r := ""
 		if msg.IsGovernance {
--- a/packages/chain/consensus/action.go
+++ b/packages/chain/consensus/action.go
@@ -699,7 +699,8 @@ func (c *consensus) finalizeTransaction(sigSharesToAggregate []*dss.PartialSig) 
 }
 
 func (c *consensus) setNewState(msg *messages.StateTransitionMsg) bool {
-	if msg.State.BlockIndex() != msg.StateOutput.GetStateIndex() {
+	sameIndex := msg.State.BlockIndex() == msg.StateOutput.GetStateIndex()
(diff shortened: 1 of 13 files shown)


dessaya (approved):

packages/chain/consensus/action.go
```diff
@@ -717,18 +718,34 @@ func (c *consensus) setNewState(msg *messages.StateTransitionMsg) bool {
 		return false
 	}
 
-	c.stateOutput = msg.StateOutput
-	c.currentState = msg.State
+	// If c.stateOutput.GetStateIndex() == msg.StateOutput.GetStateIndex() and the new state output is not a governance update, then either,
+	// a) c.stateOutput is the same as msg.StateOutput and there is no need to reasign c.stateOutput or b) msg.StateOutput is a regular state update
+	// output and c.stateOutput is a governance update output with the same index; in such case governance update is the last and should be taken
+	// into account. Regular state output is overwriden by governance uptade output and should be ignored.
+	// TODO: it is assumed, that at most one governance update may occur in between regural state updates. The situation of several consecutive
```
jorgemmsilva:
```suggestion
	// TODO: it is assumed, that at most one governance update may occur in between regular state updates. The situation of several consecutive
```

packages/chain/consensus/action.go
```diff
@@ -717,18 +718,34 @@ func (c *consensus) setNewState(msg *messages.StateTransitionMsg) bool {
 		return false
 	}
 
-	c.stateOutput = msg.StateOutput
-	c.currentState = msg.State
+	// If c.stateOutput.GetStateIndex() == msg.StateOutput.GetStateIndex() and the new state output is not a governance update, then either,
+	// a) c.stateOutput is the same as msg.StateOutput and there is no need to reasign c.stateOutput or b) msg.StateOutput is a regular state update
```
jorgemmsilva:
```suggestion
	// a) c.stateOutput is the same as msg.StateOutput and there is no need to reassign c.stateOutput or b) msg.StateOutput is a regular state update
```

packages/chain/consensus/action.go
```diff
@@ -717,18 +718,34 @@ func (c *consensus) setNewState(msg *messages.StateTransitionMsg) bool {
...
+	// output and c.stateOutput is a governance update output with the same index; in such case governance update is the last and should be taken
+	// into account. Regular state output is overwriden by governance uptade output and should be ignored.
```
jorgemmsilva:
```suggestion
	// into account. Regular state output is overwritten by governance update output and should be ignored.
```

packages/chain/consensus/action.go
```diff
@@ -717,18 +718,34 @@ func (c *consensus) setNewState(msg *messages.StateTransitionMsg) bool {
...
+	// TODO: it is assumed, that at most one governance update may occur in between regural state updates. The situation of several consecutive
+	// governance updates must be yet to be discussed, designed and implemented. The main problem is that there is no way in knowing the exact order
```
jorgemmsilva:
```suggestion
	// governance updates are yet to be discussed, designed and implemented. The main problem is that there is no way in knowing the exact order
```
what does this mean?

packages/peering/domain/domain.go
```diff
@@ -67,8 +67,17 @@ func (d *DomainImpl) GetRandomOtherPeers(upToNumPeers int) []*cryptolib.PublicKe
...
 		ret[i] = d.permPubKeys[d.permutation.NextNoCycles()]
+		distinct := true
+		for j := 0; j < i && distinct; j++ {
+			if ret[i].Equals(ret[j]) {
+				distinct = false
+			}
+		}
+		if distinct {
+			i++
+		}
```
jorgemmsilva, line 80: shouldn't the permutation already guarantee that there are not duplicates?

jorgemmsilva (approved):

packages/chain/consensus/action.go
Juliusan: This means that if two different governance rotation transactions are approved one after another without state transaction in between, there is no guarantee that the nodes are going to be synchronised correctly. Maybe the wording is difficult to understand. I'll try to improve it. And thanks for all your spelling corrections. I am terrible at that.

packages/peering/domain/domain.go
Juliusan: Well... the algorithm shuffles all the nodes at creation. Then on every call to `GetRandomOtherPeers(n)` returns `n` nodes. After exhausting the current permutation, nodes are reshuffled to keep the order of peers constantly changing. E.g., see `TestGetRandomOtherPeers`: we have 7 (other) nodes and constantly take 5 random ones. Let's say that on creation the following permutation was created: 4 2 1 5 3 0 6. Then first call would return "4 2 1 5 3". The second one should return "0 6" and the permutation gets reshuffled to get other 3 nodes. Let's say, that after reshuffling we get permutation 5 1 6 2 3 4 0. Then the full return from `GetRandomOtherPeers` would be "0 6 5 1 6" - and here you go: two duplicating nodes.
Another approach would be to shuffle nodes on each call to `GetRandomOtherPeers`. But this wouldn't ensure that random peers starts to repeat on different calls only when all the peers are exhausted. Currently this is used for requesting blocks from other nodes. It is beneficial to send requests to all the different nodes before rerequesting the node, which was already queried.

jorgemmsilva: I see, that is tricky. 
Something that came to mind: we could keep the initial permutation and just cycle through it, so for  `4 2 1 5 3 0 6`
 - first call returns `4 2 1 5 3`
 - second call returns `0 6 4 2 1`
- and so on

I guess the name `"GetRandomOtherPeers"` wouldn't make so much sense then, tho - but it could be a simpler solution.

just an idea, whatever works 💪

Juliusan: That was the previous solution, before I introduced shuffling after permutation exhaustion. This caused errors on some specific settings: 10 nodes total, picking 5 nodes - first call, third call, fifth call etc... always return the same result. This was causing some livelocks in state manager synchronization.

PR #1101, opened 2022-06-30
Stardust vm julius

Juliusan:
# Description of change

Changes in consensus for external governance rotation (so called maintenance mode).

## Type of change

All the short and regular tests pass.

## Change checklist

### DIFF
--- a/packages/chain/consensus/action.go
+++ b/packages/chain/consensus/action.go
@@ -160,13 +160,23 @@ func (c *consensus) runVMIfNeeded() {
 				c.log.Errorf("runVM result: VM task failed: %v", vmTask.VMError)
 				return
 			}
-			c.log.Debugf("runVM result: responding by state index: %d state commitment: %s",
-				vmTask.VirtualStateAccess.BlockIndex(), state.RootCommitment(vmTask.VirtualStateAccess.TrieNodeStore()))
-			c.EnqueueVMResultMsg(&messages.VMResultMsg{
-				Task: vmTask,
-			})
-			elapsed := time.Since(c.workflow.GetVMStartedTime())
-			c.consensusMetrics.RecordVMRunTime(elapsed)
+			finalRequestsCount := len(vmTask.Results)
+			if finalRequestsCount > 0 {
+				finalRequests := make([]iscp.Request, finalRequestsCount)
+				for i := range vmTask.Results {
+					finalRequests[i] = vmTask.Results[i].Request
+				}
+				c.log.Debugf("runVM result: responding by state index: %d, state commitment: %s, included %v requests: %v",
+					vmTask.VirtualStateAccess.BlockIndex(), state.RootCommitment(vmTask.VirtualStateAccess.TrieNodeStore()), finalRequestsCount, iscp.ShortRequestIDsFromRequests(finalRequests))
+				c.EnqueueVMResultMsg(&messages.VMResultMsg{
+					Task: vmTask,
+				})
+				elapsed := time.Since(c.workflow.GetVMStartedTime())
+				c.consensusMetrics.RecordVMRunTime(elapsed)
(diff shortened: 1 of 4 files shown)


packages/chain/consensus/action.go
```diff
@@ -160,13 +160,23 @@ func (c *consensus) runVMIfNeeded() {
...
+				}
+				c.log.Debugf("runVM result: responding by state index: %d, state commitment: %s, included %v requests: %v",
+					vmTask.VirtualStateAccess.BlockIndex(), state.RootCommitment(vmTask.VirtualStateAccess.TrieNodeStore()), finalRequestsCount, iscp.ShortRequestIDsFromRequests(finalRequests))
+				c.EnqueueVMResultMsg(&messages.VMResultMsg{
+					Task: vmTask,
+				})
+				elapsed := time.Since(c.workflow.GetVMStartedTime())
+				c.consensusMetrics.RecordVMRunTime(elapsed)
+			} else {
+				c.log.Debugf("runVM result: no requests included, ignoring the result and reseting the workflow")
```
jorgemmsilva:
```suggestion
				c.log.Debugf("runVM result: no requests included, ignoring the result and restarting the workflow")
```

packages/chain/consensus/action.go
```diff
@@ -160,13 +160,23 @@ func (c *consensus) runVMIfNeeded() {
...
 			}
-			c.log.Debugf("runVM result: responding by state index: %d state commitment: %s",
-				vmTask.VirtualStateAccess.BlockIndex(), state.RootCommitment(vmTask.VirtualStateAccess.TrieNodeStore()))
-			c.EnqueueVMResultMsg(&messages.VMResultMsg{
-				Task: vmTask,
-			})
-			elapsed := time.Since(c.workflow.GetVMStartedTime())
-			c.consensusMetrics.RecordVMRunTime(elapsed)
+			finalRequestsCount := len(vmTask.Results)
+			if finalRequestsCount > 0 {
```
jorgemmsilva: this is getting a bit too nested, could we invert the if and return early?

```
if finalRequestsCount == 0 {
	// log...
	c.resetWorkflow()
	return
}
... the rest
```

--- 

the same could be done in the `if` above actually
```
vmTask == nil {
	c.log.Errorf("runVM: error preparing VM task")
	return
}
... the rest
```

packages/chain/consensus/action.go
```diff
@@ -160,13 +160,23 @@ func (c *consensus) runVMIfNeeded() {
...
+				finalRequests := make([]iscp.Request, finalRequestsCount)
+				for i := range vmTask.Results {
+					finalRequests[i] = vmTask.Results[i].Request
+				}
```
jorgemmsilva: is this really worth executing this loop every block just for logging? 
the list of requests on each block is on blocklog anyway, I would remove it.

jorgemmsilva (approved):

dessaya (approved):

packages/chain/consensus/action.go
Juliusan: I hate that style of coding! :-D But I'll change it.

Juliusan: Obtained this, while changing:
```packages/chain/consensus/action.go:96: Function 'runVMIfNeeded' has too many statements (51 > 50) (funlen)```
Apparently, Go admits that such style increases function length :-D

packages/chain/consensus/action.go
Juliusan: It's not a very heavy loop. Its iteration is simply unpacking an object. Time should be negligible IMO and it uses some additional space, which is not that large too. I'll add a comment, but I'll leave it for now, because it is really convenient to know, what is the final result of the VM.

PR #1102, opened 2022-06-30
Add send-funds command

lmoe:
# Description of change

Adds the `send-funds` command. 
Introduces a new flag `--adjust-dust-deposit` which adjusts the iota transfer to the least min deposit amount if needed.
Adds the subSeed back into the keyPair to make use of the `--address-index` flag again.

## Type of change


tools/wasp-cli/wallet/send.go
```diff
@@ -1,38 +1,56 @@
...
+
+		txID, err := tx.ID()
+
+		log.Check(err)
+
+		err = client.PostTx(tx)
+
+		log.Check(err)
+
+		log.Printf("Transaction [%v] successfully sent", txID)
```
dessaya: This is just a matter of taste, but I find it more readable to remove the extra line between an operation and the error check, to emphasize that both lines form a single unit with a single purpose:

```suggestion
		tx, err := transaction.NewTransferTransaction(transaction.NewTransferTransactionParams{
			DisableAutoAdjustDustDeposit: !adjustDustDeposit,
			FungibleTokens:               tokens,
			SendOptions:                  iscp.SendOptions{},
			SenderAddress:                senderAddress,
			SenderKeyPair:                wallet.KeyPair,
			TargetAddress:                targetAddress,
			UnspentOutputs:               outputSet,
			UnspentOutputIDs:             iscp.OutputSetToOutputIDs(outputSet),
		})
		log.Check(err)

		txID, err := tx.ID()
		log.Check(err)

		err = client.PostTx(tx)
		log.Check(err)

		log.Printf("Transaction [%v] successfully sent", txID)
```

dessaya (approved):
Maybe add a test in `tools/cluster/tests/wasp-cli_test.go`? :)

PR #1104, opened 2022-06-30
Proper parameter names for governance contract

BugFreeSoftware:


dessaya (approved):

howjmay (approved):

packages/wasmvm/wasmlib/go/wasmlib/coregovernance/consts.go
```diff
@@ -17,28 +17,28 @@ const (
...
-	ResultAllowedStateControllerAddresses = "b"
+	ResultAllowedStateControllerAddresses = "a"
 	ResultChainID                         = "c"
-	ResultChainOwner                      = "c"
+	ResultChainOwner                      = "o"
 	ResultChainOwnerID                    = "o"
 	ResultDescription                     = "d"
-	ResultFeePolicyBytes                  = "d"
+	ResultFeePolicyBytes                  = "g"
 	ResultGasFeePolicyBytes               = "g"
```
jorgemmsilva, line 42: we have some params/results with the same value ("o","g"), isn't this potentially problematic?

BugFreeSoftware: No, they actually have to be. Schema tool separates them out in Params and Results, but for some reason the core contracts will reuse var and param definitions for results.

PR #1118, opened 2022-07-06
in Solo VM runner is configurable for each chain

lunfardo314:
Enhancement, non-breaking.
Now the VM runner is configurable for each chain in Solo. The default is StardustVM (the one in the `vm` package).
Preparation for validity proofs PoC and execution layer abstraction
All tests pass


lmoe (approved):

jorgemmsilva (approved):

packages/solo/solo.go
```diff
@@ -119,6 +122,16 @@ type InitOptions struct {
...
 }
 
+type InitChainOptions struct {
+	// optional parameters for init request call
+	InitRequestParameters dict.Dict
+	// optional VMRunner. Default is StardustVM
+	VMRunner vm.VMRunner
+	// flag forces bypassing any StardustVM ledger-dependent calls, such as init or blocklog
+	// To be used with provided non-standard VMRunner
+	BypassStardustVM bool
```
dessaya, line 132: Maybe the `VMRunner` interface can have `IsStardust() bool` instead?

packages/solo/solo.go
```diff
@@ -230,8 +256,12 @@ func (env *Solo) NewChainExt(chainOriginator *cryptolib.KeyPair, initIotas uint6
 	)
 	require.NoError(env.T, err)
 
+	anchor, _, err := transaction.GetAnchorFromTransaction(originTx)
+	require.NoError(env.T, err)
+
 	err = env.utxoDB.AddToLedger(originTx)
 	require.NoError(env.T, err)
+	env.AssertL1Iotas(originatorAddr, utxodb.FundsFromFaucetAmount-anchor.Deposit)
```
dessaya, line 264: Now the same `originatorAddr` is used for all chains by default. This assertion breaks if you deploy 2 chains (there is a test that fails with this).

PR #1122, opened 2022-07-06
refactor: deprecate redundant VMResult error

jorgemmsilva:
# Description of change

VMResult.Error didn't make sense anymore because the error was already being registered in the receipt 

## Type of change

- Refactor

all tests pass

### DIFF
--- a/packages/vm/core/blocklog/receipt.go
+++ b/packages/vm/core/blocklog/receipt.go
@@ -116,6 +116,19 @@ func (r *RequestReceipt) LookupKey() RequestLookupKey {
 	return NewRequestLookupKey(r.BlockIndex, r.RequestIndex)
 }
 
+func (r *RequestReceipt) ToTranslatedReceipt(translatedError *iscp.VMError) *iscp.Receipt {
+	return &iscp.Receipt{
+		Request:         r.Request.Bytes(),
+		Error:           r.Error,
+		GasBudget:       r.GasBudget,
+		GasBurned:       r.GasBurned,
+		GasFeeCharged:   r.GasFeeCharged,
+		BlockIndex:      r.BlockIndex,
+		RequestIndex:    r.RequestIndex,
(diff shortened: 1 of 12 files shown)


lmoe (approved):

packages/vm/vmtask.go
```diff
@@ -67,7 +67,6 @@ type RequestResult struct {
 	// Return is the return value of the call
 	Return dict.Dict
 	// Error is the error produced by the call, if any
-	Error error
```
dessaya, line 70: Forgot to remove the comment :stuck_out_tongue:

packages/vm/core/blocklog/receipt.go
```diff
@@ -116,6 +116,19 @@ func (r *RequestReceipt) LookupKey() RequestLookupKey {
 	return NewRequestLookupKey(r.BlockIndex, r.RequestIndex)
 }
 
+func (r *RequestReceipt) ToTranslatedReceipt(translatedError *iscp.VMError) *iscp.Receipt {
```
dessaya, line 119: The mix of "resolved" and "translated" might make things confusing.

Maybe call this `ToResolvedReceipt(resolvedError *iscp.VMError)`?

Or change all usages of "resolve" to "translate", but IMO "resolve" is better in this case: you have an `UnresolvedVMError` and you resolve it into a `VMError`. (Or you have a `VMError` and you resolve it into a `ResolvedVMError` :joy:  )

dessaya (approved):
Sorry, I wasn't able to review before it was merged.

packages/vm/core/blocklog/receipt.go
jorgemmsilva: thanks for pointing this out. this is confusing as it is, no need to add extra confusion with new terms :smile: .
we have other places where I've written "translate", so I'll produce a new PR to make everything consistent :+1:

PR #1127, opened 2022-07-07
Gascalibration

ivange94:
# Description of change

Gas calibration tests

Please write a summary of your changes and why you made them. 

## Change checklist


contracts/wasm/gascalibration/executiontime/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2020 IOTA Stiftung
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "executiontime"
+description = "executiontime description"
+license = "Apache-2.0"
+version = "0.1.0"
+authors = ["John Doe <john@doe.org>"]
```
dessaya: maybe put your name here? Or just IOTA Foundation? Same in other `Cargo.toml` files

dessaya (approved):

packages/vm/viewcontext/viewcontext.go
```diff
@@ -44,7 +44,7 @@ type ViewContext struct {
 	gasBudget      uint64
 	gasBurnEnabled bool
 	callStack      []*callContext
-	l1Params       *parameters.L1Params
+	l1Params       *parameters.L1Params // nolint
```
jorgemmsilva: question: why is nolint needed here?

btw, we should not ignore ALL the linting rules, only the rules that shouldn't apply to the code in question, example: `//nolint:deadcode`

jorgemmsilva (approved):
too many`nolint` state that don't explain why they are needed, and should disable only the specific rule that is needed.

the "execution time" benchmark doesn't make much sense, the wasm languages are basically flat :disappointed:

packages/vm/viewcontext/viewcontext.go
ivange94: I didn't just ignore all. I added nolint here and one other place because I wasn't sure how to handle them. The linter reported that these fields are not used. For one of them the field was actually being used. It's the `Chain.lastReceipt`. Not sure why the linter was saying its not used. For the one you commented on, I did a search in my editor and couldn't find a reference to the field but still wasn't sure if the right move was to delete it so I ignored.

jorgemmsilva: very rarely the liter is mistaken :stuck_out_tongue:  anyway, you should never "abolish all rules", in this case it would be something like `//nolint:unused`.

However, these variables should be *removed* if not in use.

jorgemmsilva (changes requested):
please remove the unused variables, instead of marking them as "lint ignore".

packages/vm/viewcontext/viewcontext.go
ivange94: In this particular case the linter isn't wrong. But it also complains about `Chain.lastReceipt` but that field is been used and deleting it causes compiler errors

ivange94: Strange. I couldn't find any reference to `Chain.lastReceipt` this time. I could swear the build was breaking before when I deleted it 😂

contracts/wasm/gascalibration/executiontime/test/executiontime_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+package test
+
+import (
+	"flag"
+	"testing"
+
+	. "github.com/iotaledger/wasp/contracts/wasm/gascalibration"
```
jorgemmsilva: is this needed?

jorgemmsilva (approved):

contracts/wasm/gascalibration/executiontime/test/executiontime_test.go
ivange94: Fixed

PR #1134, opened 2022-07-08
Make ETH addresses on chain clickable

lmoe:
# Description of change

ETH addresses were just shown as text inside the dashboard instead of clickable links. This adds this.

## Type of change


packages/dashboard/templates/base.tmpl
```diff
@@ -13,14 +13,22 @@
...
+	{{ if not $isValidAddr }}
+		<code>{{$agentidStr}}</code>
+	{{ else }}
+		{{ $addr := (addressFromAgentID $agentid) }}
+		{{ $isETHAddress := (isETHAddress $agentid) }}
+		{{ if $isETHAddress }}
+			{{ $agentidStr := (getETHAddress $agentid) }}
+		{{ end }}
+			<a href="{{ uri "chainAccount" $chainid $agentidStr }}"><code>{{ $agentidStr }}</code></a>
+		{{ if $addr }}
```
jorgemmsilva, line 28: Shouldn't all agent IDs link to on-chain accounts? (Maybe an extra icon with a link to the explorer for l1 addresses)

jorgemmsilva (approved):

packages/dashboard/templates/base.tmpl
lmoe: Yeah, the list has a fallback if the item is just text for what ever reason. That was there before, we can remove it if we are sure that every item is fine.

Besides this, it lists all accounts as links and if it's a parsable address, we show the l1 explorer next to the link.

dessaya (approved):

PR #1143, opened 2022-07-13
improvement: reduce off-ledger broadcast noise

jorgemmsilva:
off-ledger request gossip was creating a lot of noise (unnecessary messages).

This PR alleviates that issue with the following changes:
- remove ack messages - those are not needed because delivery is guaranteed by the transport layer
- don't broadcast to nodes we have received that same request from
- don't broadcast twice to the same node

### DIFF
--- a/packages/chain/chainimpl/eventproc.go
+++ b/packages/chain/chainimpl/eventproc.go
@@ -26,7 +26,6 @@ func (c *chainObj) recvLoop() {
 	dismissChainMsgChannel := c.dismissChainMsgPipe.Out()
 	aliasOutputChannel := c.aliasOutputPipe.Out()
 	offLedgerRequestMsgChannel := c.offLedgerRequestPeerMsgPipe.Out()
-	requestAckMsgChannel := c.requestAckPeerMsgPipe.Out()
 	missingRequestIDsMsgChannel := c.missingRequestIDsPeerMsgPipe.Out()
 	missingRequestMsgChannel := c.missingRequestPeerMsgPipe.Out()
 	timerTickMsgChannel := c.timerTickMsgPipe.Out()
@@ -50,12 +49,6 @@ func (c *chainObj) recvLoop() {
 			} else {
 				offLedgerRequestMsgChannel = nil
 			}
-		case msg, ok := <-requestAckMsgChannel:
-			if ok {
-				c.handleRequestAckPeerMsg(msg.(*messages.RequestAckMsgIn))
-			} else {
-				requestAckMsgChannel = nil
-			}
 		case msg, ok := <-missingRequestIDsMsgChannel:
 			if ok {
 				c.handleMissingRequestIDsMsg(msg.(*messages.MissingRequestIDsMsgIn))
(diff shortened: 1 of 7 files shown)


packages/chain/chainimpl/eventproc.go
```diff
@@ -291,22 +285,25 @@ func (c *chainObj) broadcastOffLedgerRequest(req iscp.OffLedgerRequest) {
...
 
-	sendMessage := func(ackPeers []*cryptolib.PublicKey) {
+	sendMessage := func(peersThatHaveTheRequest map[*cryptolib.PublicKey]bool) {
 		peerPubKeys := getPeerPubKeys(c.offledgerBroadcastUpToNPeers)
 		for _, peerPubKey := range peerPubKeys {
-			if shouldSendToPeer(peerPubKey, ackPeers) {
-				c.log.Debugf("sending offledger request ID: reqID: %s, peerPubKey: %s", req.ID(), peerPubKey.AsString())
-				c.chainPeers.SendMsgByPubKey(peerPubKey, peering.PeerMessageReceiverChain, chain.PeerMsgTypeOffLedgerRequest, msg.Bytes())
+			if peersThatHaveTheRequest[peerPubKey] {
+				// this peer already has the request
```
Juliusan, line 292: I would add a debug message here notifying that the peer was skipped.

packages/chain/chainimpl/eventproc.go
```diff
@@ -317,41 +314,18 @@ func (c *chainObj) broadcastOffLedgerRequest(req iscp.OffLedgerRequest) {
 			if !c.mempool.HasRequest(req.ID()) {
 				return
 			}
-			c.offLedgerReqsAcksMutex.RLock()
-			ackPeers := c.offLedgerReqsAcks[req.ID()]
-			c.offLedgerReqsAcksMutex.RUnlock()
-			if cmt != nil && len(ackPeers) >= int(cmt.Size())-1 {
+			c.offLedgerPeersHaveReqMutex.RLock()
+			peersThatHaveTheRequest := c.offLedgerPeersHaveReq[req.ID()]
+			c.offLedgerPeersHaveReqMutex.RUnlock()
```
Juliusan: Isn't it too early to release a lock? `peersThatHaveTheRequest` is a REFERENCE to a map. It's not a copy of underlying map. So occurring edits of `c.offLedgerPeersHaveReq[req.ID()]` might collide with usage of `peersThatHaveTheRequest`.

tools/cluster/templates/waspconfig.go
```diff
@@ -23,7 +23,7 @@ const WaspConfig = `
     "directory": "waspdb"
   },
   "logger": {
-    "level": "debug",
+    "level": "warn",
```
Juliusan, line 26: Someone (maybe me) might have committed "debug" as a default accidentally. But wasn't "info" the default level before?

Juliusan (approved):

tools/cluster/templates/waspconfig.go
jorgemmsilva: even "info" is super verbose, I feel warn is a good as things are right now :(

packages/chain/chainimpl/eventproc.go
jorgemmsilva: good point :+1:

dessaya (approved):

PR #1150, opened 2022-07-14
Update devnet to Hornet

lmoe:
# Description of change
Replaces the GoShimmer node with a Hornet setup. 
DockerImage was changed from `buster` to `bullseye` as librocksdb doesn't build on `buster`.

This includes the new hornet-nest image, keeping the folder clean. 

## Type of change


jorgemmsilva (commented):
not a big fan of copying these files from the Hornet repo. I would rather have a link to the correct files on the original source

lmoe:
Yeah me neither, but they compile Hornet by source. I want a fixed hornet docker version for now to keep things stable.

jorgemmsilva (approved):
I guess lets merge this to have the docs in a "usable state" for now, even if we should move to a "low maintenance", easier to setup version in the future.

howjmay:
I think rebase to the latest commit on develop branch can solve the required branch protection

jorgemmsilva:
@HowJMay yeah that is okay, just re-running the ci jobs should fix this.
We're waiting on another docker image from hornet that will make this setup much much simpler.

lmoe:
The bloated setup was now replaced with the new hornet-nest image. I've tested it with a locally built hornet-nest image and it works fine. Once the next Hornet release comes up, we can merge. @jorgemmsilva

PR #1157, opened 2022-07-18
Add allowance argument to post-request

lmoe:
# Description of change

Adds `Allowance` options to `post-request`

## Type of change

Added cluster test

### DIFF
--- a/tools/cluster/tests/wasp-cli_test.go
+++ b/tools/cluster/tests/wasp-cli_test.go
@@ -193,6 +193,10 @@ func TestWaspCLIContract(t *testing.T) {
 	// test off-ledger request
 	w.Run("chain", "post-request", name, "increment", "--off-ledger")
 	checkCounter(45)
+
+	// include an allowance transfer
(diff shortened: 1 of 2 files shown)


tools/cluster/tests/wasp-cli_test.go
```diff
@@ -193,6 +193,10 @@ func TestWaspCLIContract(t *testing.T) {
 	// test off-ledger request
 	w.Run("chain", "post-request", name, "increment", "--off-ledger")
 	checkCounter(45)
+
+	// include an allowance transfer
+	w.Run("chain", "post-request", name, "increment", "--transfer=iota:10", "--allowance=iota:10")
```
jorgemmsilva: it is not possible to send 10 iotas in 1 request (less than SD), I would suggest to increase this value to something more realistic.

jorgemmsilva (approved):

tools/cluster/tests/wasp-cli_test.go
lmoe: Ah, I've just copied the post-request test a few lines above. Thought the minDepositAdjust flag is enabled, as the tests have succeeded. 

I'll fix the tests then.

PR #1158, opened 2022-07-18
Set a constant read header timeout

lmoe:
# Description of change

Linter currently complains about a missing ReadHeaderTimeout in the metrics package.

I've set it to 3 seconds for now, as it's just about reading the header section, not the entire request. 

https://en.wikipedia.org/wiki/Slowloris_(computer_security)

### DIFF
--- a/packages/metrics/metrics.go
+++ b/packages/metrics/metrics.go
@@ -15,6 +15,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+const ReadHeaderTimeout = time.Second * 3
+
 type Metrics struct {
 	server                  *http.Server
 	log                     *logger.Logger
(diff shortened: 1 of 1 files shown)


jorgemmsilva (approved):
I'm all for this change, but isn't 3s excessive time to read the headers of a request? I'd say like 500ms

lmoe:
@jorgemmsilva I would say so, yes. But I don't know how the metrics service behaves with prometheus. 
That's why I've marked larry as well.

ivange94:
Yeah I did see that error and was going to fix in my pr but had to cook and
didn’t think this was a blocker. I think 500ms is fine. The metrics module
doesn’t do any heavy lifting before start up. So any delays will be based
on when the node gets to the plug-in as plug-ins are started sequentially.

lmoe:
thanks!

PR #1166, opened 2022-07-22
Artifacts

BugFreeSoftware:
Test removal of schema-tool-generated artifacts


dessaya:
:heart:

Makefile
```diff
@@ -33,8 +33,8 @@ test-full: install
 test: install
 	go test -tags $(BUILD_TAGS) $(TEST_PKG) --timeout 40m --count 1 -failfast $(TEST_ARG)
 
-test-short:
-	go test -tags $(BUILD_TAGS) --short --count 1 -failfast ./...
+test-short-no-wasm:
```
dessaya: The `test-short` rule is mentioned in the documentation. I suggest doing a global search & replace to update it. Or maybe better, don't change the name?

BugFreeSoftware: This is only a (failing) try-out for now. I agree that changing the name is not necessary when the original test-short is no longer there. I reported my findings to Yanghao so he can try to fix things.

jorgemmsilva (approved):
feels-good-man

PR #1176, opened 2022-07-27
docs: update core contracts section

dessaya:


documentation/docs/guide/core_concepts/core_contracts/accounts.md
```diff
@@ -9,122 +9,222 @@ keywords:
...
+
+Parameters:
+
+- `a` (`AgentID`): The account Agent ID
 
-### - `balance(a AgentID)`
+Returns:
 
-Returns the fungible tokens owned by any AgentID `a` on the chain.
+A map of Foundry ID ([`FoundryID`](#foundryid)) => `big.Int`. The L1 base tokens is represented by an empty Foundry ID (a key with length 0).
```
jorgemmsilva: I think we should use TokenID instead of FoundryID, its easier to understand we are talking about tokens.

documentation/docs/guide/core_concepts/core_contracts/accounts.md
```diff
@@ -9,122 +9,222 @@ keywords:
...
+```
+
+### `TokenScheme`
+
+A `TokenScheme` is encoded as the concatenation of:
+
+- Simple token scheme prefix: 0 (`byte`)
+- Amount of minted tokens (`iotago::uint256`)
+- Amount of melted tokens (`iotago::uint256`)
+- Maximum supply (`iotago::uint256`)
```
jorgemmsilva: why `iotago:uint256` instead of just `uint256` ?

jorgemmsilva (approved):
:raised_hands: great improvement! Thank you!

documentation/docs/guide/core_concepts/core_contracts/accounts.md
dessaya: Because iotago serializes big.Int in a specific way: https://github.com/iotaledger/hive.go/blob/master/serializer/serializer.go#L140

Frankly, our docs are still lacking a complete specification of the serialization schemes. For example, is `uint64` little or big endian?

Anyway, to "solve" this case I decided to just say `iotago::TokenScheme` and link to the TokenScheme source in the iota.go repository. Same for `iotago::Address`, `iotago::FoundryOutput`, etc. I don't know if there is a better place to link to.

jorgemmsilva: good point :ok_hand: . 
We need tools to handle the serialization for us :confused: nobody should need to care about low level details when interacting with SCs, The "high level" concepts are complex enough as it is

PR #1187, opened 2022-08-01
Develop tmp

Juliusan:
A temporary PR created in order to catch an error, reported in the following thread: https://iotafoundation.slack.com/archives/G01F91K2CRX/p1658921560434169. No approval needed. It is not going to be merged.


Juliusan:
No longer needed... probably.

PR #1191, opened 2022-08-03
docs: chain and nodes update

jorgemmsilva:
# Description of change

- Split and re-write some of the installation docs so its more user-friendly
- deprecate pure windows in favor of WSL
- update default config.json

### DIFF
--- a/documentation/docs/guide/chains_and_nodes/running-a-node.md
+++ b/documentation/docs/guide/chains_and_nodes/running-a-node.md
@@ -15,11 +15,9 @@ keywords:
 
 # Running a Node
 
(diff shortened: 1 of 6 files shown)


documentation/docs/guide/chains_and_nodes/running-a-node.md
```diff
@@ -29,100 +27,14 @@ If you just want to run a Wasp node, you can also use the [Wasp standalone Docke
 ### Software
 
 - [Go 1.18](https://golang.org/doc/install)
```
dessaya: Go is a prerequisite for installing, so I think it should be moved to the "installng wasp" section?

dessaya (approved):

documentation/docs/guide/chains_and_nodes/running-a-node.md
jorgemmsilva: Good point 👍

PR #1205, opened 2022-08-05
EVM fixes

dessaya:
* Increment nonce after failed tx
* Take assets from ISC caller account (= eth tx sender) when calling isc.send() & isc.call()


packages/vm/core/evm/evmimpl/iscmagic.go
```diff
@@ -61,6 +62,32 @@ func newMagicContract(ctx isc.Sandbox) vm.ISCContract {
...
+		req.FungibleTokens.BaseTokens = sd
+	}
+}
+
+// moveAssetsToCommonAccount moves the assets from the caller's L2 account to the common
+// account before sending to L1
+// TODO: should use allowance and c.ctx.TransferAllowedFunds() instead
+func moveAssetsToCommonAccount(ctx isc.Sandbox, caller vm.ContractRef, fungibleTokens *isc.FungibleTokens, nftIDs []iotago.NFTID) {
+	ctx.Privileged().MustMoveBetweenAccounts(
+		ctx.Caller(), // should be the eth caller?
```
jorgemmsilva, line 84: >// should be the eth caller?

I'd say so. is it not?

jorgemmsilva (approved):
noice🚀 
there are some linting issues.
any reason to not add the other test? 😂 I can push it later, I guess

jorgemmsilva:
this seems broken for some reason
>Test / Wasm contract tests (tswasm) (pull_request) Failing after 6m — Wasm contract tests (tswasm)

@HowJMay @BugFreeSoftware

dessaya:
> any reason to not add the other test?

Because the existing unit test tests the same thing with much less hassle I think.
But feel free to push it if you think it adds value.

packages/vm/core/evm/evmimpl/iscmagic.go
dessaya: The thing is, the "ethereum caller" at this point in the test that I made is the ISCTest contract, not the same as the "ethereum sender".

It's all very confusing, and I don't think this is the correct solution, but at least the tests now pass :stuck_out_tongue:

jorgemmsilva:
I get what you mean, but I wanted to test calling the contract directly which is what metamask is doing. I'll add it later 😄

packages/vm/core/evm/evmimpl/iscmagic.go
jorgemmsilva: That is correct imo

dessaya: If I put the ethereum caller here it fails because the caller is the ISCTest contract which has an empty balance.

```
tx sender --> calls --> ISCTest:sendBaseTokens --> calls --> iscmagic:send
```

The magic contract is adding an extra hop which further complicates things. Maybe we have to do a `delegatecall` instead of a regular call...

jorgemmsilva: >If I put the ethereum caller here it fails because the caller is the ISCTest contract which has an empty balance.

why is that wrong?

dessaya: Well, that my test doesn't pass in that case :P

But your test would still pass, because you are calling the magic contract directly.
So yes, I agree that taking funds from the eth caller would be more correct. But since there is no way for the sender to specify allowance, the more common use case (calling a Solidity contract that in turn calls `isc.send`) would not work.

jorgemmsilva: it needs to be the caller, otherwise there is no way to move funds owned by EVM some contract.

---
regarding how to implement the allowance in EVM, that's another rabbit hole :smile: 

My current thoughts:
The allowance can be a parameter of `iscmagic.Send()` (that can be used by contracts that call it).

Regarding the case of chain calls:
```
user => "contract X" => iscmagic.Send
 ```
 "contract X" method would need to have an "allowance" parameter, and then call `ctx.TransferAllowedFunds` via `iscmagic` first, so the funds are moved to their account. It is then able to spend those funds.
 
the implementation of `TransferAllowedFunds` by the iscmagic contract needs to have a special check to ensure the allowance is not tinkered by "contract X" (and check the original (signed) tx data).

This "claim mechanism" would only work for allowance provided by the sender, but maybe its good enough? :thinking:

PR #1206, opened 2022-08-08
Makefile

Juliusan:
# Description of change

Now make can be used to build contracts. Generator edited to not change the dates of files that are not changed and to change dates of files that are regenerated.

## Type of change


jorgemmsilva (approved):
It looks good to me, but @BugFreeSoftware and @howjmay should provide feedback as its most important for the WASM part.

howjmay:
There are some changes in schema tool, I don't understand their usages. Like return nil. Are they changes made by linter?

Juliusan:
> There are some changes in schema tool, I don't understand their usages. Like return nil. Are they changes made by linter?

The main change was in [tools/schema/generator/generator.go](https://github.com/iotaledger/wasp/pull/1206/files#diff-c66cbf45a4246201789bd126c014a5f76bf5d627466b387eb391e3067f14e1d6) in `generateFuncs` function. As I understood, `contractName.go` (or .rs, or .ts) file is edited to include some functionality and then it is used to generate other files. However, schema tool edits this file during go code generation to include more functions. Sometimes it results in no changes at all, but the file modification date is still changed as schema tool recreates the file. The algorithm before my change was: move the old file, create the new one and if everything ok remove the old one. I changed it to be: create new file and if it is the same as the current one, do not touch the original and simply delete the new one. Otherwise overwrite the original with the newly generated code, which is different. I believe, apart from leaving file modification date unchanged, this change doesn't alter the functionality.

`return nil` in various places was caused by change in `createFile` of [tools/schema/generator/generator.go](https://github.com/iotaledger/wasp/pull/1206/files#diff-c66cbf45a4246201789bd126c014a5f76bf5d627466b387eb391e3067f14e1d6). I felt, it is more natural to return error from generation function this way. Thus the interface of `createFile` function changed and this resulted in various `return nil` statements scattered through the code. This is not strictly necessary for this PR.

Another small change is in `Generate` function of [tools/schema/generator/generator_ts.go](https://github.com/iotaledger/wasp/pull/1206/files#diff-aea84cdcfea7f3f05119fb0dd01e649f664a1da9ffefa413b71c8e6d410ca538). This forces recreation of `tsconfig.json` file so that `make` would notice that it has done its job. Recreation itself is not necessary, just the update of file modification date.

BugFreeSoftware:
So the whole idea behind the schema tool is that we remove as many burdens from the programmer as possible. But now we add a burden of maintaining a makefile per contract, where the programmer is *not* the one who decides which files are being generated but the schema tool is. So the schema tool can decide to add files based on changes in the schema.yaml file without the programmer even realizing. So he does not know to update this makefile.
The schema tool uses a very simple logic. It determines if the schema tool itself or the schema.yaml are newer than consts.xx, which is a file that must always be generated. If not newer it will forego generating of everything as it is obviously up to date. Generating itself is lightning fast compared to the overhead of compilation and generating the Wasm file, so we do not care that all artifact files are regenerated. In the case of github these files anyway always need to be generated because the artifact files are not in the repo.
My goal for makefiles was to have an automatic rule that if it finds a schema.yaml in any folder it will run the schema tool in that folder. And that should happen before the respective compiler will be run, which could be done by a rule that makes based on consts.xx. Having to specify which folders contain the schema.yaml file and having to add a makefile and having to keep that one up to date goes against the spirit of the schema tool. So I don't know if we can do something like that.
It's been at least 2 decades since I last really used make, and the syntax has changed considerably, so I don't know all the tricks possible. But I do know that 30 years ago I already used to build a project based on scanning its subfolders and building according to rules for the file types found in these subfolders. So I would expect some makefile in contracts/wasm that can do that without needing all those makefiles in each folder.

BugFreeSoftware (changes requested):
So the whole idea behind the schema tool is that we remove as many burdens from the programmer as possible. But now we add a burden of maintaining a makefile per contract, where the programmer is not the one who decides which files are being generated but the schema tool is. So the schema tool can decide to add files based on changes in the schema.yaml file without the programmer even realizing. So he does not know to update this makefile.
The schema tool uses a very simple logic. It determines if the schema tool itself or the schema.yaml are newer than consts.xx, which is a file that must always be generated. If not newer it will forego generating of everything as it is obviously up to date. Generating itself is lightning fast compared to the overhead of compilation and generating the Wasm file, so we do not care that all artifact files are regenerated. In the case of github these files anyway always need to be generated because the artifact files are not in the repo.
My goal for makefiles was to have an automatic rule that if it finds a schema.yaml in any folder it will run the schema tool in that folder. And that should happen before the respective compiler will be run, which could be done by a rule that makes based on consts.xx. Having to specify which folders contain the schema.yaml file and having to add a makefile and having to keep that one up to date goes against the spirit of the schema tool. So I don't know if we can do something about that.
It's been at least 2 decades since I last really used make, and the syntax has changed considerably, so I don't know all the tricks possible. But I do know that 30 years ago I already used to build a project based on scanning its subfolders and building according to rules for the file types found in these subfolders. So I would expect some makefile in contracts/wasm that can do that without needing all those makefiles in each folder.

BugFreeSoftware:
Suggest we close this one as it has been superseded by other PRs

PR #1208, opened 2022-08-09
Update iotago, fix Pow

lmoe:
# Description of change

This PR updates the iota.go lib to v3.0.0-beta.4 and fixes the Pow failing when the MinPoWScore is set to 0.

## Type of change

`make test-short`


packages/nodeconn/pow.go
```diff
@@ -10,36 +10,49 @@ import (
...
 var ErrOperationAborted = errors.New("operation was aborted")
 
 // RefreshTipsFunc refreshes tips of the block if PoW takes longer than a configured duration.
 type RefreshTipsFunc = func() (tips iotago.BlockIDs, err error)
 
-// doPoW does the proof-of-work required to hit the given target score.
+// DoPoW does the proof-of-work required to hit the given target score.
 // The given iota.Block's nonce is automatically updated.
-func doPoW(ctx context.Context, block *iotago.Block, refreshTipsInterval time.Duration, refreshTipsFunc RefreshTipsFunc) error {
+func doPoW(ctx context.Context, block *iotago.Block, targetScore float64, parallelism int, refreshTipsInterval time.Duration, refreshTipsFunc RefreshTipsFunc) (blockSize int, err error) {
```
jorgemmsilva, line 28: the function signature doesn't need to change.

`blockSize` doesn't need to be returned

lmoe: It's a direct copy from https://github.com/iotaledger/inx-app/blob/ea47c776549669b81a5105a0ce78c7074694f44e/pow/pow.go

Alex suggested to use this one instead. I've figured instead of modifiying this file, I kept how it is to make it easier to update it should things change again.

Another way would be to import inx-app as a module, but I don't want to include it's dependencies it has.

jorgemmsilva: Sure, there seems to be a build error btw

lmoe: Fixed with https://github.com/iotaledger/wasp/pull/1209, just required a merge back from dev.
The linter issue is another topic though. :D

jorgemmsilva (approved):

PR #1210, opened 2022-08-09
Fix: produce EVM receipt when ISC panics

jorgemmsilva:
also adds a preemptive check to reject low gas tx's in JSON-RPC


packages/vm/core/evm/evmimpl/impl.go
```diff
@@ -101,19 +102,32 @@ func applyTransaction(ctx isc.Sandbox) dict.Dict {
...
+		gasErr = panicutil.CatchPanic(
+			func() {
+				ctx.Gas().Burn(gas.BurnCodeEVM1P, evmtypes.EVMGasToISC(result.UsedGas, &gasRatio))
+			},
+		)
 		ctx.Privileged().GasBurnEnable(false)
+		if gasErr != nil {
+			// out of gas when burning ISC gas, edit the EVM receipt so that it fails
+			receipt.Status = types.ReceiptStatusFailed
+			// TODO is this correct? - if this happens, will the TX be reverted?
```
jorgemmsilva: @dessaya do you think this makes sense?
I'm thinking about a test for this, but its not straightforward to implement

packages/evm/evmtest/ISCTest.sol
```diff
@@ -147,4 +149,31 @@ contract ISCTest {
...
+        // moves funds owned by the current contract to the targetAgentID
+        ISCDict memory params = ISCDict(new ISCDictItem[](1));
+        params.items[0] = ISCDictItem("a", targetAgentID.data);
+        isc.call(
+            isc.hn("accounts"),
+            isc.hn("transferAllowanceTo"),
+            params,
+            allowance
+        );
+    }
```
dessaya: IIUC the "current contract" here is the `evm` core contract. So this will take the funds from the `evm` core contract. Is that the intention here?

packages/vm/core/evm/evmimpl/iscmagic.go
```diff
@@ -88,8 +90,31 @@ func moveAssetsToCommonAccount(ctx isc.Sandbox, fungibleTokens *isc.FungibleToke
...
+		func() {
+			ret, remainingGas = run(evm, caller, input, gas, readOnly)
+		},
+		vmexceptions.AllProtocolLimits...,
+	)
+	if err != nil {
+		remainingGas = gas
+		err = vm.ErrExecutionReverted
+		// the ISC error is lost inside the EVM, a possible solution would be to wrap the ErrExecutionReverted error, but the ISC information still gets deleted at some point
+		// err = errors.Wrap(vm.ErrExecutionReverted, err.Error())
```
dessaya, line 113: At least log the error?

packages/vm/core/evm/evmtest/evm_test.go
```diff
@@ -612,3 +602,82 @@ func TestFibonacciContract(t *testing.T) {
...
+	env := initEVM(t)
+	ethKey, _ := env.soloChain.NewEthereumAccountWithL2Funds()
+	iscTest := env.deployISCTestContract(ethKey)
+
+	// credit base tokens to the ISC test contract
+	env.soloChain.GetL2FundsFromFaucet(isc.NewEthereumAddressAgentID(iscTest.address))
+
+	randAgentID := isc.NewAgentID(tpkg.RandEd25519Address())
+
+	t.Fail()
```
dessaya: ?

packages/vm/core/evm/iscmagic/types.go
```diff
@@ -173,8 +173,8 @@ func (n ISCNFT) MustUnwrap() *isc.NFT {
 // ISCAllowance matches the struct definition in ISC.sol
 type ISCAllowance struct {
 	BaseTokens uint64
-	Tokens     []NativeToken
-	NFTs       []NFTID
+	Assets     []NativeToken
+	Nfts       []NFTID
```
dessaya: Any reason for this change `NFTs` -> `Nfts`? `NFTs` is more idiomatic IIRC.

packages/vm/core/evm/evmimpl/impl.go
dessaya: It's probably correct, but not 100% sure. I think it would be important to have a test for this, if possible.

dessaya (approved):

packages/vm/core/evm/iscmagic/types.go
jorgemmsilva: Yeah, it would fail when serializing for EVM

packages/evm/evmtest/ISCTest.sol
jorgemmsilva: No, you're right - I'll add better assertions. Should be from isctest contract

packages/vm/core/evm/evmtest/evm_test.go
jorgemmsilva: Lol, well spotted. But the tests are passing, wtf 🤔 . I'll revisit this and improve the test so it checks for the correct accounts balances

packages/vm/core/evm/iscmagic/types.go
dessaya: I see. Then I propose keeping the name `Tokens`, and changing the definision in ISC.sol: `assets -> tokens`.

jorgemmsilva: Sounds good 👍

packages/vm/core/evm/evmimpl/impl.go
```diff
@@ -226,7 +239,6 @@ func callContract(ctx isc.SandboxView) dict.Dict {
 	}
 
 	ctx.RequireNoError(err)
-	ctx.Requiref(res.Err == nil, GetRevertErrorMessage(res, ctx.Contract()))
```
dessaya, line 229: Why is this removed? How does the caller tell if the call failed?

packages/vm/core/evm/evmimpl/iscmagic.go
```diff
@@ -93,13 +96,36 @@ func moveAssetsToCommonAccount(ctx isc.Sandbox, caller vm.ContractRef, fungibleT
...
 
+type RunFunc func(evm *vm.EVM, caller vm.ContractRef, input []byte, gas uint64, readOnly bool) (ret []byte, remainingGas uint64)
+
+// catchISCPanics executes a `Run` function (either from a call or view), and catches ISC exceptions, if any ISC exception happens, ErrExecutionReverted is issued
+func catchISCPanics(run RunFunc, evm *vm.EVM, caller vm.ContractRef, input []byte, gas uint64, readOnly bool, log isc.LogInterface) (ret []byte, remainingGas uint64, err error) {
+	err = panicutil.CatchAllExcept(
+		func() {
+			ret, remainingGas = run(evm, caller, input, gas, readOnly)
+		},
+		vmexceptions.AllProtocolLimits...,
```
dessaya, line 107: Should DBError be included in the `except` list?

dessaya (approved):

packages/vm/core/evm/evmimpl/impl.go
jorgemmsilva: there will be an error in the EVM receipt

packages/vm/core/evm/evmimpl/iscmagic.go
jorgemmsilva: good point, added inside the "catching func", as DBError is a struct and not a regular error for some reason

packages/vm/core/evm/evmimpl/impl.go
dessaya: There is no EVM receipt in a view call :thinking:

jorgemmsilva: That's right. `err` should contain the error.
In practice that line would never be reached, because its prefaced by `ctx.RequireNoError(err)` it would always panic before

PR #1211, opened 2022-08-09
evm: `allow` and `takeAllowedFunds` magic contract functions

dessaya:
* `isc.send` and friends now take the funds from the eth caller
* New `isc.allow` function: specifies an allowance between the caller
  and any Ethereum address. This allowance is stored in the magic
  contract's state.
* New `isc.takeAllowedFunds` function
* Removed magic functions to query the ISC allowance and caller, which
  do not make sense form EVM perspective.


jorgemmsilva (approved):
good stuff! :rocket: 
I like this "allow" mechanism - simple but effective. 
The only downside I can think of is that users would need 2 requests in some cases, but that is a small price to pay.

jorgemmsilva:
will merge this as I need to update my current work with the allowance stuff

PR #1214, opened 2022-08-10
wasp-cli: improve handling of `parameters.L1`

dessaya:
* `parameters.L1` is now a function, it will panic with an appropriate error message if not previously initialized.
* wasp-cli now lazily initializes `parameters.L1`, so it is no longer necessary to check `if L1 == nil` every time.
* wasp-cli now saves the L1 params in the config file, so the node is queried only once.

### DIFF
--- a/tools/wasp-cli/chain/accounts.go
+++ b/tools/wasp-cli/chain/accounts.go
@@ -8,9 +8,7 @@ import (
 	"github.com/iotaledger/wasp/packages/isc"
 	"github.com/iotaledger/wasp/packages/kv/codec"
 	"github.com/iotaledger/wasp/packages/kv/dict"
-	"github.com/iotaledger/wasp/packages/parameters"
 	"github.com/iotaledger/wasp/packages/vm/core/accounts"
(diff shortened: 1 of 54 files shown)


tools/wasp-cli/chain/accounts.go
```diff
@@ -86,8 +82,8 @@ var depositCmd = &cobra.Command{
 	Short: "Deposit L1 funds into the given (default: your) L2 account",
 	Args:  cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		if parameters.L1 == nil {
-			config.L1Client() // this will fill parameters.L1 with data from the L1 node
+		if parameters.L1() == nil {
+			config.L1Client() // this will fill parameters.L1() with data from the L1 node
```
jorgemmsilva: Think this `if` can be removed

jorgemmsilva (approved):

tools/wasp-cli/chain/accounts.go
dessaya: Nice catch!

PR #1222, opened 2022-08-12
Fix missing mutex unlock

muXxer:


jorgemmsilva (approved):

packages/chain/chainimpl/eventproc.go
```diff
@@ -316,15 +317,20 @@ func (c *chainObj) broadcastOffLedgerRequest(req isc.OffLedgerRequest) {
...
+
+			func() {
+				// deep copy the list of peers that have the request (otherwise we get a pointer to the map instead)
+				c.offLedgerPeersHaveReqMutex.Lock()
+				defer c.offLedgerPeersHaveReqMutex.Unlock()
+
+				peersThatHaveTheRequest := c.offLedgerPeersHaveReq[req.ID()]
+				if cmt != nil && len(peersThatHaveTheRequest) >= int(cmt.Size())-1 {
+					// this node is part of the committee and the message has already been received by every other committee node
+					return
```
jorgemmsilva, line 329: actually, now that I think about it, having this wrapped in a inline func means this `return` does not exit the outside function like it should. It won't break, but will run more times than needed :stuck_out_tongue: 
Will improve in a later PR

Issue #1194, opened 2022-08-03
`make` generates a binary call `cmd`


dessaya:
`make` currently generates the following binaries:

```
cmd*
dbinspector*
schema*
snapshot*
submsg*
wasp*
wasp-cli*
wasp-cluster*
```

Note that all are properly named except `cmd`, which corresponds to the `gascalibration`.

We should do one of the following:

* Fix the `gascalibration` tool to generate a binary called `gascalibration`
* Maybe we don't need to compile everything when running `make`? Just `wasp` and `wasp-cli`?

jorgemmsilva:
>Maybe we don't need to compile everything when running make? Just wasp and wasp-cli?

absolutely agree. I would add `wasp-cluster` as well, for convenience

PR #1227, opened 2022-08-13
Only build most used binaries with make build

ivange94:
# Description of change

Closes https://github.com/iotaledger/wasp/issues/1194

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ goshimmer.log
 
 *.trie
 *.snapshot
+snapshot
+gascalibration
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,11 @@ else
 	cd packages/evm/evmtest && go generate
 endif
 
-build: compile-solidity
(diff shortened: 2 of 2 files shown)


dessaya (approved):
Nice! Do the same with `make install`? :pray:

.gitignore
```diff
@@ -26,3 +26,5 @@ goshimmer.log
 
 *.trie
 *.snapshot
+snapshot
+gascalibration
```
lmoe, line 30: Do we still have this gascalibration binary? I've thought it's inside the wasp-cli now.

lmoe (approved):
Neat!

.gitignore
ivange94: I'm not aware of `gascalibration` being made part of `wasp-cli`. Do you have a link to the commit?

lmoe: Nope, seems like I was mistaken. :)

dessaya (approved):

PR #1231, opened 2022-08-16
evm: improvements in allowance mechanism

dessaya:
* `isc.takeAllowedFunds()` takes all allowed by default
* added `isc.getAllowanceFrom()` and `isc.getAllowanceTo()` views
* fixed indentation in ISC.sol
* renamed `SandboxView::State()` -> `StateR()` and moved to `SandboxBase`,
  to improve reusability


lmoe (approved):
Personally I would call it ReadOnlyState instead of StateR to make it a bit clearer if its for reading.

luke-thorne (approved):

dessaya:
> Personally I would call it ReadOnlyState instead of StateR to make it a bit clearer if its for reading.

Yeah, or maybe `StateReadOnly`?
Anyway, I'll leave it for a later PR, for now I think it is ok since we use the `WhateverR` convention in several other places.

PR #1250, opened 2022-08-25
Fix 2 race conditions

lunfardo314:
fixed two race conditions:
- in global index baseline. This one is very important
- in mempool counter. This one is less important

### DIFF
--- a/packages/chain/mempool/mempool.go
+++ b/packages/chain/mempool/mempool.go
@@ -97,13 +97,16 @@ func (m *mempool) addToInBuffer(req isc.Request) bool {
 	return true
 }
 
-func (m *mempool) removeFromInBuffer(req isc.Calldata) {
+func (m *mempool) removeFromInBuffer(req isc.Calldata) int {
 	m.inMutex.Lock()
 	defer m.inMutex.Unlock()
+
+	removed := 0
 	if _, ok := m.inBuffer[req.ID()]; ok {
 		delete(m.inBuffer, req.ID())
-		m.outBufCounter++
+		removed++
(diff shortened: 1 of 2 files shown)


dessaya (approved):

packages/isc/coreutil/globalindex.go
```diff
@@ -76,10 +79,16 @@ func newStateIndexBaseline(globalStateIndex *atomic.Uint32) *stateBaseline {
...
 
 func (g *stateBaseline) Set() {
+	g.mutex.Lock()
+	defer g.mutex.Unlock()
+
 	g.baseline = g.solidStateIndex.Load()
 }
 
 func (g *stateBaseline) IsValid() bool {
+	g.mutex.Lock()
```
muXxer, line 89: Isn't it enough if that is a RLock in case of a read operation?

packages/chain/mempool/mempool.go
```diff
@@ -464,7 +467,11 @@ func (m *mempool) moveToPoolLoop() {
 			}
 			for i, req := range buf {
 				if m.addToPool(req) {
-					m.removeFromInBuffer(req)
+					removed := m.removeFromInBuffer(req)
+
+					m.poolMutex.Lock()
+					m.outBufCounter += removed
+					m.poolMutex.Unlock()
```
muXxer, line 474: If `m.removeFromInBuffer(req)` would return a `bool` instead of a `count` (that only can be 0 or 1 anyway), you could prevent locking of the `poolMutex` in case it was `false`.

PR #1262, opened 2022-09-01
snap-cli: 'extractblocks' feature

lunfardo314:
added `-extractblocks` feature to `snap-cli`

### DIFF
--- a/tools/snap-cli/main.go
+++ b/tools/snap-cli/main.go
@@ -26,7 +26,7 @@ const usage = `USAGE: snap-cli [-create | -scanfile | -restoredb | -verify] <fil
 or
 USAGE: snap-cli -validate <chainID> <L1 API endpoint> <L2 API endpoint>
 or
-USAGE: snap-cli -extractblocks <chainID> <from> [<to>]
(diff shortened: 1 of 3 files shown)


tools/snap-cli/main.go
```diff
@@ -274,3 +276,49 @@ func createSnapshot(dbDir string) {
...
+		return true
+	})
+	sort.Slice(indices, func(i, j int) bool {
+		return indices[i] < indices[j]
+	})
+	var data []byte
+	for _, idx := range indices {
+		if idx < from {
+			continue
+		}
```
jorgemmsilva, line 306:
```go
	for i := from; i < len(indices); i++ {
		...
	}
```

jorgemmsilva (approved):

tools/snap-cli/main.go
lunfardo314: why? The `range` is supposed to be used like that

lunfardo314: indices may not make continuous sequence in the database, so we traverse the array and choose those which are bigger than `from`

PR #1264, opened 2022-09-02
Replace mqtt with inx

lmoe:
# Description of change

(Dont merge yet)!

This PR introduces INX and replaces Mqtt with it. 

Docker files were updated accordingly. 

## Type of change

`make test-short`


config.json
```diff
@@ -1,6 +1,6 @@
 {
   "l1": {
-    "apiAddress": "http://localhost:14265"
+    "inxAddress": "localhost:9011"
```
muXxer: our default INX port is `9029`.
private tangle points to `9011` because of the first node in the private setup.

packages/util/l1starter/l1starter.go
```diff
@@ -21,6 +21,7 @@ type L1Starter struct {
 func New(flags *flag.FlagSet) *L1Starter {
 	s := &L1Starter{}
 	flags.StringVar(&s.Config.APIAddress, "layer1-api", "", "layer1 API address")
```
muXxer, line 23: Is this still needed?

packages/testutil/privtangle/privtangle.go
```diff
@@ -29,11 +29,11 @@ import (
 )
 
 // requires hornet, and inx plugins binaries to be in PATH
-// https://github.com/gohornet/hornet (1653bf6)
-// https://github.com/gohornet/inx-mqtt (9447a3c)
-// https://github.com/gohornet/inx-indexer (2d612d1)
-// https://github.com/gohornet/inx-coordinator (4b5c458)
-// https://github.com/gohornet/inx-faucet (77dc47e) (requires `git submodule update --init --recursive` before building )
+// https://github.com/gohornet/hornet (v2.0.0-beta.6)
```
muXxer: Better use updated versions here because there was a breaking protocol change.

packages/parameters/parameters.go
```diff
@@ -37,6 +37,7 @@ const (
 	DashboardExploreAddressURL = "dashboard.exploreAddressUrl"
 	DashboardAuth              = "dashboard.auth"
 
+	L1INXAddress   = "l1.inxAddress"
 	L1APIAddress   = "l1.apiAddress"
```
muXxer: same here, can maybe be removed

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+// OutputMap implements L1Connection
+func (c *l1client) OutputMap(myAddress iotago.Address, timeout ...time.Duration) (iotago.OutputSet, error) {
+	ctxWithTimeout, cancelContext := newCtx(c.ctx, timeout...)
+	defer cancelContext()
+
+	bech32Addr := myAddress.Bech32(parameters.L1().Protocol.Bech32HRP)
+	queries := []nodeclient.IndexerQuery{
+		&nodeclient.BasicOutputsQuery{AddressBech32: bech32Addr},
```
muXxer, line 88: you handle that these outputs could have unlock conditions like timelock etc, right?

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+// PostTx implements L1Connection
+// sends any tx to the L1 node, then waits until the tx is confirmed.
+func (c *l1client) PostTx(tx *iotago.Transaction, timeout ...time.Duration) error {
+	ctxWithTimeout, cancelContext := newCtx(c.ctx, timeout...)
+	defer cancelContext()
+
+	// Build a Block and post it.
+	block, err := builder.NewBlockBuilder().
+		Payload(tx).
+		Tips(ctxWithTimeout, c.nodeAPIClient).
```
muXxer, line 124: why do you need to pass tips? let the node do this!
less back and forth. Remove the local PoW and let the node do it.

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+		Build()
+	if err != nil {
+		return fmt.Errorf("failed to build a tx: %w", err)
+	}
+	submitBlock := func(ctx context.Context, block *iotago.Block) error {
+		_, err := c.nodeAPIClient.SubmitBlock(ctx, block, parameters.L1().Protocol)
+		return err
+	}
+	err = nodeconn.DoBlockPow(ctxWithTimeout, block, c.config.UseRemotePoW, submitBlock, c.nodeAPIClient)
```
muXxer: Can be removed.

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+	c.log.Infof("Posted transaction id %v", isc.TxID(txID))
+
+	return c.waitUntilConfirmed(ctxWithTimeout, block)
+}
+
+const pollConfirmedTxInterval = 200 * time.Millisecond
+
+// waitUntilConfirmed waits until a given tx Block is confirmed, it takes care of promotions/re-attachments for that Block
+func (c *l1client) waitUntilConfirmed(ctx context.Context, block *iotago.Block) error {
```
muXxer, line 158: IMO this whole logic needs to be changed to work correctly with more than one consensus node.
But maybe create an issue and do it in another PR.

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+	initialAddrOutputs, err := c.OutputMap(addr)
+	if err != nil {
+		return err
+	}
+	ctxWithTimeout, cancelContext := newCtx(c.ctx, timeout...)
+	defer cancelContext()
+
+	faucetReq := fmt.Sprintf("{\"address\":%q}", addr.Bech32(parameters.L1().Protocol.Bech32HRP))
+	faucetURL := fmt.Sprintf("%s/api/enqueue", c.config.FaucetAddress)
+	httpReq, err := http.NewRequestWithContext(ctxWithTimeout, "POST", faucetURL, bytes.NewReader([]byte(faucetReq)))
```
muXxer: use `http.MethodPost` instead of "POST"

jorgemmsilva (approved):

packages/util/l1starter/l1starter.go
jorgemmsilva: yes, there are more components beside wasp will still need the api address (wasp-cli, apilib, cluster tool/tests, etc)

packages/testutil/privtangle/privtangle.go
jorgemmsilva: thanks for pointing that out. This is updated with the latest versions we tested with, at the time those were the latest. We will test/update this with the newest stuff

ivange94 (approved):

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to create request: %w", err)
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+	res, err := http.DefaultClient.Do(httpReq)
+	if err != nil {
+		return fmt.Errorf("unable to call faucet: %w", err)
+	}
+	if res.StatusCode != 202 {
```
muXxer: 202 is also not best practice, use http.Status something

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+		resBody, err := io.ReadAll(res.Body)
```
muXxer, line 266: Couldn't the body be nil in some cases?

packages/l1connection/l1connection.go
```diff
@@ -0,0 +1,378 @@
...
+	if len(timeout) > 0 {
+		t = timeout[0]
+	}
+	return context.WithTimeout(ctx, t)
+}
+
+func setL1ProtocolParams(info *nodeclient.InfoResponse) {
+	parameters.InitL1(&parameters.L1Params{
+		// There are no limits on how big from a size perspective an essence can be, so it is just derived from 32KB - Block fields without payload = max size of the payload
+		MaxTransactionSize: 32000, // TODO should this value come from the API in the future? or some const in iotago?
```
muXxer, line 373: 32000 ? how was this calculated?

packages/nodeconn/pow.go
```diff
@@ -9,9 +9,45 @@ import (
...
 	"github.com/iotaledger/iota.go/v3/pow"
+	"github.com/iotaledger/wasp/packages/parameters"
 )
 
+type submitBlockFn = func(ctx context.Context, block *iotago.Block) error
+
+func DoBlockPow(ctx context.Context, block *iotago.Block, useRemotePow bool, submitBlock submitBlockFn, nodeClient *nodeclient.Client) error {
+	if useRemotePow {
+		// remote PoW: Take the Block, clear parents, clear nonce, send to node
+		block.Parents = nil
```
muXxer, line 27: why would you clear the parents? what if you want to reference a special tip?
Like in the promoteTx ?

packages/nodeconn/pow.go
```diff
@@ -9,9 +9,45 @@ import (
...
+		block.Nonce = 0
+		err := submitBlock(ctx, block)
+		return err
+	}
+	// do the PoW
+	refreshTipsFn := func() (tips iotago.BlockIDs, err error) {
+		// refresh tips if PoW takes longer than a configured duration.
+		resp, err := nodeClient.Tips(ctx)
```
muXxer, line 35: Your refresh tips function should take care of reusing the given tips.
We have the correct logic in another inx project (I think the inx-spammer).
But to simplify it, I would let the node do the PoW.

packages/nodeconn/nc_chain.go
```diff
@@ -145,6 +92,11 @@ func (ncc *ncChain) PullStateOutputByID(id iotago.OutputID) {
 	ncc.stateOutputHandler(id, out)
 }
 
+func shouldBeProcessed(out iotago.Output) bool {
+	// only outputs without SDRC should be processed.
+	return !out.UnlockConditionSet().HasStorageDepositReturnCondition()
```
muXxer, line 119: This will be changed in the future, right?! Right?!?!??!?!??!

packages/nodeconn/nodeconn.go
```diff
@@ -235,67 +302,58 @@ func (nc *nodeConn) GetMetrics() nodeconnmetrics.NodeConnectionMetrics {
...
-	return block, nil
+
+	return &blockID, nil
 }
 
 const pollConfirmedTxInterval = 200 * time.Millisecond
 
 // waitUntilConfirmed waits until a given tx Block is confirmed, it takes care of promotions/re-attachments for that Block
-func (nc *nodeConn) waitUntilConfirmed(ctx context.Context, block *iotago.Block) error {
+func (nc *nodeConn) waitUntilConfirmed(ctx context.Context, blockID *iotago.BlockID) error {
```
muXxer: Same comments from my side like before in a similar function...

packages/l1connection/l1connection.go
jorgemmsilva: `l1connection` is used by tools that are to act as L1 clients, not connected via INX

packages/l1connection/l1connection.go
jorgemmsilva: same as https://github.com/iotaledger/wasp/pull/1264#discussion_r962626050

packages/l1connection/l1connection.go
jorgemmsilva: https://github.com/iotaledger/wasp/pull/1264#discussion_r962626050

jorgemmsilva: we can indeed improve the logic, but its not here

packages/l1connection/l1connection.go
jorgemmsilva: as the comment above says:

> // There are no limits on how big from a size perspective an essence can be, so it is just derived from 32KB - Block fields without payload = max size of the payload

this was taken from a slack conversation, afaik there currently is no constant provided by iota.go

packages/nodeconn/pow.go
jorgemmsilva: this is only used in "remotePow". I the parents were removed as per the code provided by Alex.

packages/nodeconn/pow.go
jorgemmsilva: again, not every component of wasp can rely on "INX PoW". This code was taken directly from inx-spammer iirc.

packages/nodeconn/nc_chain.go
jorgemmsilva: >?!?!??!?!??!

less drama please.
Yes, that's the plan, but right now the focus is on making the node withstand many days without crashing.

packages/l1connection/l1connection.go
jorgemmsilva: those are handled inside each chain. We could potentially add another check for that here, but its not strictly needed

packages/nodeconn/nodeconn.go
jorgemmsilva: https://github.com/iotaledger/wasp/issues/1267

packages/l1connection/l1connection.go
muXxer: what about something like this? (payload type is not included, because its part of the payload)

```go
	payloadSizeMax := iotago.BlockBinSerializedMaxSize - // BlockSizeMax
		serializer.OneByte - // ProtocolVersion
		serializer.OneByte - // ParentCount
		(iotago.BlockMaxParents * iotago.BlockIDLength) - // Parents
		serializer.UInt32ByteSize - // PayloadLenght
		serializer.UInt64ByteSize // Nonce
```

packages/nodeconn/nc_chain.go
muXxer: Sorry, I was a bit shocked tbh, because all this storage deposit return stuff was added to the protocol in the first place to be able to send on chain requests without fees. Would be a shame if that is not used at the end.

packages/nodeconn/pow.go
muXxer: But there are parts in your code where you rely on the given parents.
We need to talk about that and not just ignore the problem.

packages/l1connection/l1connection.go
muXxer: Still, this is not needed.
The node will add the tips, so less overhead.

packages/nodeconn/nc_chain.go
jorgemmsilva: I don't think that is the case. IIRC it was to keep the "protocol is feeless" and you can "send stuff for free" narratives, nothing to do with smart contracts.

jorgemmsilva: anyway, it should be supported - but its not a priority IMO

packages/l1connection/l1connection.go
jorgemmsilva: thanks, will update in another PR

packages/nodeconn/pow.go
jorgemmsilva: where do you see that?

packages/l1connection/l1connection.go
muXxer: `http.StatusAccepted`

packages/l1connection/l1connection.go
jorgemmsilva: >	// For server requests, the Request Body is always non-nil

https://pkg.go.dev/net/http#Response

muXxer:
meeeeerge it :rocket:

muXxer (approved):

PR #1269, opened 2022-09-05
Implement event driven confirmation checks

muXxer:
This PR changes the logic for confirmation checks from active polling to an event driven approach.
Confirmation is checked with every applied ledger update (new confirmed milestones).

It is based on the transaction ID instead of the ID of the attached block.
This way it also safe to post the transaction in a leaderless fashion by every validator.

PR is work in progress, needs some small cleanups....
And I didn't test the code at all :sweat_smile:

### DIFF
--- a/packages/nodeconn/nodeconn.go
+++ b/packages/nodeconn/nodeconn.go
@@ -32,9 +32,10 @@ import (
 )
 
 const (
-	inxTimeoutBlockMetadata     = 500 * time.Millisecond
-	inxTimeoutSubmitBlock       = 60 * time.Second
-	reattachWorkerPoolQueueSize = 100
+	inxTimeoutBlockMetadata      = 500 * time.Millisecond
+	inxTimeoutSubmitBlock        = 60 * time.Second
+	inxTimeoutPublishTransaction = 120 * time.Second
+	reattachWorkerPoolQueueSize  = 100
 )
 
 type ChainL1Config struct {
@@ -79,7 +80,7 @@ func setL1ProtocolParams(info *nodeclient.InfoResponse) {
 
 const defaultTimeout = 1 * time.Minute
 
-func newCtx(ctx context.Context, timeout ...time.Duration) (context.Context, context.CancelFunc) {
+func newCtxWithTimeout(ctx context.Context, timeout ...time.Duration) (context.Context, context.CancelFunc) {
 	t := defaultTimeout
 	if len(timeout) > 0 {
 		t = timeout[0]
@@ -90,7 +91,7 @@ func newCtx(ctx context.Context, timeout ...time.Duration) (context.Context, con
 func New(config ChainL1Config, log *logger.Logger, timeout ...time.Duration) chain.NodeConnection {
 	ctx, ctxCancel := context.WithCancel(context.Background())
 
-	ctxWithTimeout, cancelContext := newCtx(ctx, timeout...)
+	ctxWithTimeout, cancelContext := newCtxWithTimeout(ctx, timeout...)
 	defer cancelContext()
 
 	nb, err := nodebridge.NewNodeBridge(ctxWithTimeout, config.INXAddress, log.Named("NodeBridge"))
@@ -307,7 +308,7 @@ func (nc *nodeConn) PublishStateTransaction(chainID *isc.ChainID, stateIndex uin
 		return err
 	}
 
-	return ncc.PublishTransaction(tx)
+	return ncc.PublishTransaction(tx, inxTimeoutPublishTransaction)
 }
 
 // PublishGovernanceTransaction implements chain.NodeConnection.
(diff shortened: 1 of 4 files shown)


packages/nodeconn/pending_tx.go
```diff
@@ -0,0 +1,125 @@
...
+	if tx.conflicting.Load() {
+		return xerrors.Errorf("transaction was conflicting: %s", tx.transactionID.ToHex())
+	}
+
+	if !tx.confirmed.Load() {
+		return xerrors.Errorf("context was canceled but transaction was not confirmed: %s, error: %s", tx.transactionID.ToHex(), tx.ctx.Err())
+	}
+
+	// transaction was confirmed
+	return nil
```
jorgemmsilva, line 124: `return nil` will never be reached, `tx.confirmed.Load()` is either true, or false

jorgemmsilva: nevemind, I read this wrong.

packages/nodeconn/pending_tx.go
```diff
@@ -0,0 +1,125 @@
...
+func (tx *PendingTransaction) ConflictReason() error {
+	return tx.conflictReason
+}
+
+// WaitUntilConfirmed waits until a given tx Block is confirmed, it takes care of promotions/re-attachments for that Block
+func (tx *PendingTransaction) WaitUntilConfirmed() error {
+	// wait until the context is done
+	<-tx.ctx.Done()
+
+	if tx.conflicting.Load() {
```
jorgemmsilva:
```suggestion
	if tx.Conflicting() {
```

packages/nodeconn/pending_tx.go
```diff
@@ -0,0 +1,125 @@
...
+		return xerrors.Errorf("transaction was conflicting: %s", tx.transactionID.ToHex())
+	}
+
+	if !tx.confirmed.Load() {
```
jorgemmsilva:
```suggestion
	if !tx.Confirmed() {
```

packages/nodeconn/pending_tx.go
```diff
@@ -0,0 +1,125 @@
...
+	return tx.conflictReason
+}
+
+// WaitUntilConfirmed waits until a given tx Block is confirmed, it takes care of promotions/re-attachments for that Block
+func (tx *PendingTransaction) WaitUntilConfirmed() error {
+	// wait until the context is done
+	<-tx.ctx.Done()
+
+	if tx.conflicting.Load() {
+		return xerrors.Errorf("transaction was conflicting: %s", tx.transactionID.ToHex())
```
jorgemmsilva: probably worth it to print `tx.ConflictReason()`

packages/nodeconn/nodeconn.go
```diff
@@ -301,89 +385,123 @@ func (nc *nodeConn) GetMetrics() nodeconnmetrics.NodeConnectionMetrics {
...
-	for {
-		metadataResp, err := nc.nodeBridge.BlockMetadata(ctx, *currentBlockID)
-		if err != nil {
-			return xerrors.Errorf("failed to get msg metadata: %w", err)
-		}
+// clearPendingTransactionWithoutLocking removes tracking of a pending transaction.
+// write lock must be acquired outside.
+func (nc *nodeConn) clearPendingTransactionWithoutLocking(transactionID iotago.TransactionID) {
+	delete(nc.pendingTransactionsMap, transactionID)
+}
```
jorgemmsilva, line 422: don't see the point of this function that only does 1 thing and its only called once, why not just do it inline? :joy:

packages/nodeconn/pending_tx.go
```diff
@@ -0,0 +1,125 @@
...
+	"go.uber.org/atomic"
+	"golang.org/x/xerrors"
+
+	iotago "github.com/iotaledger/iota.go/v3"
+)
+
+// PendingTransaction holds info about a sent transaction that is pending.
+type PendingTransaction struct {
+	ctx            context.Context
+	CtxCancel      context.CancelFunc
```
jorgemmsilva:
```suggestion
	ctxCancel      context.CancelFunc
```

can be private, as it is not used from outside

packages/nodeconn/nodeconn.go
```diff
@@ -130,9 +148,65 @@ func (nc *nodeConn) subscribeToLedgerUpdates() {
...
+	// outputs that are created and consumed in the same milestone exist in both maps.
+	newSpentsMap := make(map[iotago.OutputID]struct{})
+	for _, spent := range update.Consumed {
+		newSpentsMap[spent.GetOutput().GetOutputId().Unwrap()] = struct{}{}
+	}
+
+	newOutputsMap := make(map[iotago.OutputID]struct{})
+	for _, output := range update.Created {
+		newOutputsMap[output.GetOutputId().Unwrap()] = struct{}{}
+	}
```
jorgemmsilva, line 162: small optimization: we are already iterating through `update.Created` on line 210. We could move that code up and re-use that loop to construct `newOutputsMap`

jorgemmsilva (approved):
Thank you for this PR!
It looks quite good, just some minor comments

packages/nodeconn/nodeconn.go
muXxer: you are absolutely right.
I would leave it for now for better readability. But feel free to change it :)

packages/nodeconn/nodeconn.go
muXxer: It's just to make it more clear what this does.
The compiler will inline it anyway. :)
But you can also change that. :thumbsup:

jorgemmsilva (approved):

PR #1273, opened 2022-09-06
Use hive.go app pkg and reflection based configs

muXxer:
This PR updates to the latest hive.go version and replaces the node package with the app package.
The app package introduces reflection based config loading and dependency injection.
The PR only uses dependency injection where it is absolutely needed. 
Another PR should be done to utilize the full possibilities of the dep injection to untangle the code.

It also updates to the latest version of all other libs (except go-ethereum and wasm related stuff).
Since go-ethereum is a fork, but has a nasty "go module import bug" in a dependency (chainhash), we should update our fork.

The libp2p dependency switched finally to a monorepo, which simplifies updating it without incompatibilities between submodules.
The old libp2p repos are deprecated anyway, so I updated it as well.

The PR also adds the gendoc tool, which automatically updates the documentation of all config parameters, and create a config_defaults.json as well.

Known issues of the PR:
- User struct fields in config start with capital letter

PS:
I just assumed what would be core components and what would be actual plugins.
This might be wrong.
I also didn't check if the node crashes if the plugins are deactivated, because some code might rely on the existence of the plugins.

Hint:
wasp-cli should be moved to an own go submodule. That is rather easy and reduces the dependency hell of wasp.


core/logger/params.go
```diff
@@ -0,0 +1,25 @@
...
+
+var ParamsLogger = &ParametersLogger{}
+
+var params = &app.ComponentParams{
+	Params: map[string]any{
+		"logger": ParamsLogger,
+	},
+	Masked: nil,
+}
+*/
```
jorgemmsilva: why is this commented?

core/app/app.go
```diff
@@ -0,0 +1,84 @@
...
+			"app.checkForUpdates",
+			"app.profile",
+			"config",
+			"help",
+			"peering",
+			"version",
+		},
+		AdditionalConfigs: []*app.ConfigurationSet{
+			app.NewConfigurationSet("peering", "peering", "peeringConfigFilePath", "peeringConfig", false, true, false, "peering.json", "n"),
+		},
```
jorgemmsilva: this seems quite cryptic:
```
"peering", "peering", "peeringConfigFilePath", "peeringConfig", false, true, false, "peering.json", "n"
```
guess we need to look into that function signature to understand what those arguments are

core/logger/component.go
```diff
@@ -0,0 +1,25 @@
...
+		},
+	}
+}
+
+var (
+	CoreComponent *app.CoreComponent
+)
+
+func configure() error {
+	initGoEthLogger(CoreComponent.Logger().Named("go-ethereum"))
```
jorgemmsilva: guessing this doesn't need a plugin just for this line, we should refactor this, probably in a later PR

jorgemmsilva (commented):
Thanks again for this PR!
looks good, left some minor comments.

Linting and docs CI checks are not passing, tho.

lets merge this after we iron out the current issues :+1:

core/logger/params.go
muXxer: As a reminder that I wanted to cleanly handle these things in hive.go.
I did that now, so I could remove that whole file :+1:

core/app/app.go
muXxer: I removed it for now.

jorgemmsilva (approved):

PR #1298, opened 2022-09-13
Use inx CoreComponent to handle inx connection problems

muXxer:
This utilizes the INX CoreComponent that is used in all other INX plugins.
It automatically creates an INX connection to the node, and calls the shutdown handler of the app if the connection via INX encounters an error.

It also adds a wait time until the indexer plugin might be available on the node.

@kape1395 fyi


jorgemmsilva:
we should deprecate the "freeze" tests that rely on the "reconnection logic" we had before

packages/nodeconn/nc_chain.go
```diff
@@ -97,21 +99,40 @@ func (ncc *ncChain) PublishTransaction(tx *iotago.Transaction, timeout ...time.D
...
+		out, err := resp.GetOutput().UnwrapOutput(serializer.DeSeriModePerformValidation, ncc.nc.nodeBridge.ProtocolParameters())
+		if err != nil {
+			ncc.log.Errorf("PullOutputByID: error getting output from response - chainID %s OutputID %s:  %s", ncc.chainID, id, err)
+			return
+		}
+		output = out
+
+	//nolint:nosnakecase // grpc uses underscores
+	case *inx.OutputResponse_Spent:
+		// MH: With this you would also apply spent outputs to the current state, is that intended?
```
jorgemmsilva, line 123: yeah, I think so. I believe this is for when the node was offline for a while and is trying to sync the state.

but this pulling logic in general should be removed IMO

CC @Juliusan

jorgemmsilva (approved):

lunfardo314:
I didn't look into the code but behavior wrt INX connection should be the following:
* If INX connection is failing, wasp should be retrying to reconnect. Meanwhile Wasp won't be functional neither as committee, not as an access node because it won't have synced state. It could respond top requests for older blocks from other nodes
* probably after some timeout of failed reconnections the Wasp node should shut down. This may be  configurable
* after reconnect, the Wasp node should do all necessary subscriptions to INX and similar to recover the UTXO stream. It may need some polling, but probably not (if re-subscription does this function)

Dr-Electron:
Would be nice if we can merge this today as currently the wiki breaks and this PR fixes it 😅

PR #1329, opened 2022-09-19
Implement a text registry

ivange94:
# Description of change

Build a text registry thats compatible with the text key/value store


jorgemmsilva (approved):
looks good, but I would like to see "readable keys", since we're re-implementing a new registry.

jorgemmsilva (approved):
could you add an example of what the new text registry looks like?

PR #1346, opened 2022-09-26
Fix docker build

muXxer:


lmoe (approved):

.github/workflows/docker-push.yml
```diff
@@ -13,17 +13,15 @@ jobs:
       image_name: iotaledger/wasp
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+
+      #- name: Set up QEMU
+      #  uses: docker/setup-qemu-action@v2
```
jorgemmsilva, line 20: qemu?

jorgemmsilva (approved):

.github/workflows/docker-push.yml
alexsporn: that is used when you want to also build arm64 versions of the containers

.github/workflows/docker-push.yml
```diff
@@ -35,15 +33,22 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
+            type=match,pattern=v(\d+.\d+),suffix=-alpha,group=1,enable=${{ contains(github.ref, '-alpha') }}
+            type=match,pattern=v(\d+.\d+),suffix=-beta,group=1,enable=${{ contains(github.ref, '-beta') }}
+            type=match,pattern=v(\d+.\d+),suffix=-rc,group=1,enable=${{ contains(github.ref, '-rc') }}
```
alexsporn, line 38: I would remove these, because Wasp is using the 0.x.y versioning right now, and not the 1.0.0-beta.X pattern

muXxer: maybe they should switch to 1.0.0-beta.1 :D

PR #1353, opened 2022-09-28
fix: local setup

jorgemmsilva:
fix local docker setup with nest


tools/local-setup/docker-compose.yml
```diff
@@ -18,16 +18,14 @@ services:
...
+      - "--database.directory=/app/waspdb"
+      - "--logger.level=debug"
     container_name: wasp
     depends_on:
       - hornet-nest
     restart: on-failure:10
     networks:
       - wasp-net
-    volumes:
-      - wasp-db:/waspdb
```
lmoe, line 29: Why have you removed the waspdb volume? We should keep it, otherwise the waspdb will be saved inside the container only.

tools/local-setup/README.md
```diff
@@ -0,0 +1,64 @@
+# Local Setup
```
lmoe, line 1: Maybe we should call it something like "local-dev-setup" and clarify that this setup contains the faucet/coordinator and is only for local dev use - maybe even linking to the prod setup to show an alternative.

lmoe (approved):

tools/local-setup/docker-compose.yml
jorgemmsilva: IMO that's not a problem, this setup is meant to be ephemeral. Also saving the DB is not worth anything if we aren't able to restore the L1 state

tools/local-setup/README.md
jorgemmsilva: I mean, this seems pretty clear :joy: 

>This Directory contains a Docker-based setup to run your own Wasp development setup

PR #1359, opened 2022-10-04
Fix: evm decimals

jorgemmsilva:
enables EVM value transfer with decimals that are more user friendly.

because Ethereum has 18 decimal cases and SMR/IOTA only 6, any thing past the 6th decimal case will be ignored.

TODO:
 - make this work for non-base tokens (custom L1 native tokens)
 - understand why in some cases the gas estimated is different from the real operation (already identified the cause for tx's that send the entire user balance, but other cases remaining)


packages/vm/core/evm/emulator/emulator.go
```diff
@@ -204,7 +206,18 @@ func (e *EVMEmulator) applyMessage(msg core.Message, statedb vm.StateDB, header
 		gasBurnEnable(true)
 		defer gasBurnEnable(false)
 	}
-	return core.ApplyMessage(vmEnv, msg, &gasPool)
+
+	catchedErr := panicutil.CatchAllExcept(
```
dessaya: caughtErr :)

packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -209,6 +212,7 @@ func paramBlockNumber(ctx isc.SandboxView, emu *emulator.EVMEmulator, allowPrevi
 	return current
 }
 
+// TODO dropping "customtokens gas fee" mgith be the way to go
```
dessaya:
```suggestion
// TODO dropping "customtokens gas fee" might be the way to go
```

packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -241,7 +245,10 @@ func getBalanceFunc(ctx isc.SandboxBase) emulator.GetBalanceFunc {
 			accounts.ViewBalanceBaseToken.Hname(),
 			dict.Dict{accounts.ParamAgentID: isc.NewEthereumAddressAgentID(addr).Bytes()},
 		)
-		return new(big.Int).SetUint64(codec.MustDecodeUint64(res.MustGet(accounts.ParamBalance), 0))
+		// TODO this won't work for custom native tokens for now
```
dessaya: Maybe add an assertion?

```
ctx.Require(token is native)
```

packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -258,14 +265,28 @@ func fungibleTokensForFee(ctx isc.SandboxBase, amount *big.Int) *isc.FungibleTok
 
 func getSubBalanceFunc(ctx isc.Sandbox) emulator.SubBalanceFunc {
 	return func(addr common.Address, amount *big.Int) {
-		tokens := fungibleTokensForFee(ctx, amount)
+		// TODO this will only work for L1 base tokens, we don't have a way to know about decimals for custom tokens
```
dessaya: Same here, add an assert?

packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -258,14 +265,28 @@ func fungibleTokensForFee(ctx isc.SandboxBase, amount *big.Int) *isc.FungibleTok
...
+		) {
+			panic(vm.ErrNotEnoughTokensLeftForGas)
+		}
 	}
 }
 
 func getAddBalanceFunc(ctx isc.Sandbox) emulator.AddBalanceFunc {
 	return func(addr common.Address, amount *big.Int) {
-		tokens := fungibleTokensForFee(ctx, amount)
+		// TODO this will only work for L1 base tokens, we don't have a way to know about decimals for custom tokens
```
dessaya: And here :sweat_smile:

dessaya (approved):
Sorry for the delay! LGTM

PR #1361, opened 2022-10-05
Consensus redesign

kape1395:


packages/chain/aaa2/node/node.go
```diff
@@ -0,0 +1,376 @@
...
+	added := false
+	for i := 0; i < consensusInstsInAdvance; i++ {
+		if _, ok := cni.consensusInsts[committeeID][addLogIndex]; !ok {
+			consGrCtx, consGrCancel := context.WithCancel(ctx)
+			logIndexCopy := addLogIndex
+			cgr := consGR.New(
+				consGrCtx, cni.chainID, dkShare, &logIndexCopy, cni.nodeIdentity,
+				cni.procCache, cni.mempool, cni.stateMgr, cni.net,
+				recoveryTimeout, redeliveryPeriod, printStatusPeriod, cni.log,
+			)
```
jorgemmsilva: this is a bit awkward to read, probably better to pass a custom struct, instead of 13 parameters

packages/gpa/ack_handler.go
```diff
@@ -79,7 +79,7 @@ func (a *ackHandler) Output() Output {
 }
 
 func (a *ackHandler) StatusString() string {
-	return fmt.Sprintf("{AckHandler, nested=%s}", a.nested.StatusString())
```
jorgemmsilva, line 82: I see this "ack handler" already existed before this PR, but shouldn't we rely on the transport layer?
(TCP or QUIC)?

these protocols already have ACKs implemented (https://datatracker.ietf.org/doc/html/rfc9002)

packages/chain/aaa2/node/node.go
kape1395: Yes, that's a bit hard to read. But using a structure for the parameters is very easy to forget to pass a parameter, when new parameters are introduced. In this way a compiler shows you all the places.

packages/gpa/ack_handler.go
kape1395: Yes, they have. But they only do this for a single connection. This AckHandler will handle reconnects, and nodes that are offline for some time (e.g. restarts).

packages/chain/aaa2/node/node.go
muXxer: What if we use generics and the "Options" package?
Looks cleaner and we could panic if a certain option is not passed maybe?

muXxer:
Closed because this PR was rebased and merged to develop in #1484

PR #1368, opened 2022-10-11
jsonify cli output

ivange94:
# Description of change
JSONify cli out.


packages/log/log.go
```diff
@@ -73,9 +78,55 @@ func Fatalf(format string, args ...interface{}) {
 	os.Exit(1)
 }
 
+func DefaultJSONFormatter(i interface{}) ([]byte, error) {
+	return json.MarshalIndent(i, "", " ")
```
muXxer: I would separate by two white spaces for better readability.

```suggestion
	return json.MarshalIndent(i, "", "  ")
```

muXxer (approved):

PR #1370, opened 2022-10-13
fix: evm show revert reason

jorgemmsilva:


dessaya:
This seems like redoing the work reverted in 760df830f014c6c9ef74c659355502a638672ae3 ?

(which was already tested here https://github.com/iotaledger/wasp/blob/develop/packages/vm/core/evm/evmtest/evm_test.go#L411 )

What is the difference? Why was the previous implementation removed?

jorgemmsilva:
@dessaya thanks for pointing that out!

https://github.com/iotaledger/wasp/commit/760df830f014c6c9ef74c659355502a638672ae3 was part of https://github.com/iotaledger/wasp/pull/1210 , which solved the problem where metamask transactions would stay as "pending", even tho they have failed (because no receipts would be produced).

However, due my lack of understanding at the time of how the revert reason was specified, I also added a regression that made so that the revert reason was never shown :trollface: .

This brings back the logic to get the revert reason, it's bit simpler (just 12 lines instead of 100), but now that you mentioned it, I noticed I did left out the VMError from the magic contract - It should be removed IMO, I don't see a good reason for it to exist :confused:  .

Pushing an update that handles this even better and show the ISC panic errors as the revert reason (which I wanted to do initially but didn't know how)

dessaya (approved):
OK, simple is better than complex :+1:

PR #1412, opened 2022-11-08
feat: new mempool implementation

jorgemmsilva:
cleanup of https://github.com/iotaledger/wasp/pull/1379

### DIFF
--- a/packages/chain/aaa2/chainMgr/chainMgr.go
+++ b/packages/chain/aaa2/chainMgr/chainMgr.go
@@ -197,7 +197,7 @@ func (cmi *chainMgrImpl) Input(input gpa.Input) gpa.OutMessages {
 	case *inputConsensusTimeout:
 		return cmi.handleInputConsensusTimeout(input)
 	}
-	panic(xerrors.Errorf("unexpected input %T: %+v", input, input))
+	panic(fmt.Errorf("unexpected input %T: %+v", input, input))
 }
 
 // Implements the gpa.GPA interface.
--- a/packages/chain/aaa2/mempool/interface.go
+++ b/packages/chain/aaa2/mempool/interface.go
@@ -10,7 +10,6 @@ import (
 type Mempool interface {
 	consGR.Mempool // TODO should this be unified with the Mempool interface
 	ReceiveRequests(reqs ...isc.Request) []bool
-	RemoveRequests(reqs ...isc.RequestID)
 	HasRequest(id isc.RequestID) bool
 	GetRequest(id isc.RequestID) isc.Request
 	Info(currentTime time.Time) MempoolInfo
--- a/packages/chain/aaa2/mempool/interface.go
+++ b/packages/chain/aaa2/mempool/interface.go
@@ -1,0 +1,30 @@
+package mempool
+
+import (
+	"time"
+
+	consGR "github.com/iotaledger/wasp/packages/chain/aaa2/cons/gr"
+	"github.com/iotaledger/wasp/packages/isc"
+)
+
+type Mempool interface {
+	consGR.Mempool // TODO should this be unified with the Mempool interface
+	ReceiveRequests(reqs ...isc.Request) []bool
+	RemoveRequests(reqs ...isc.RequestID)
+	HasRequest(id isc.RequestID) bool
+	GetRequest(id isc.RequestID) isc.Request
+	Info(currentTime time.Time) MempoolInfo
+	Close()
+}
+
+// for testing (only for use in solo)
+type SoloMempool interface {
+	Mempool
+	WaitPoolEmpty(timeout ...time.Duration) bool
+}
+
+type MempoolInfo struct {
+	TotalPool      int
+	InPoolCounter  int
+	OutPoolCounter int
+}
--- a/packages/chain/aaa2/mempool/interface.go
+++ b/packages/chain/aaa2/mempool/interface.go
@@ -14,7 +14,6 @@ type Mempool interface {
 	HasRequest(id isc.RequestID) bool
 	GetRequest(id isc.RequestID) isc.Request
 	Info(currentTime time.Time) MempoolInfo
-	Close()
 }
 
 // for testing (only for use in solo)
--- a/packages/chain/aaa2/mempool/interface.go
+++ b/packages/chain/aaa2/mempool/interface.go
@@ -1,8 +1,6 @@
 package mempool
 
 import (
-	"time"
-
 	consGR "github.com/iotaledger/wasp/packages/chain/aaa2/cons/gr"
 	"github.com/iotaledger/wasp/packages/isc"
 )
@@ -10,15 +8,10 @@ import (
 type Mempool interface {
 	consGR.Mempool // TODO should this be unified with the Mempool interface
 	ReceiveRequests(reqs ...isc.Request) []bool
+	RemoveRequests(reqs ...isc.RequestID)
 	HasRequest(id isc.RequestID) bool
 	GetRequest(id isc.RequestID) isc.Request
-	Info(currentTime time.Time) MempoolInfo
-}
-
-// for testing (only for use in solo)
-type SoloMempool interface {
-	Mempool
-	WaitPoolEmpty(timeout ...time.Duration) bool
+	Info() MempoolInfo
 }
 
 type MempoolInfo struct {
--- a/packages/chain/aaa2/mempool/mempool.go
+++ b/packages/chain/aaa2/mempool/mempool.go
@@ -2,6 +2,7 @@ package mempool
 
 import (
 	"context"
+	"fmt"
 	"math"
 	"sync"
 	"time"
@@ -134,6 +135,7 @@ func (m *mempool) attachToIncomingRequests(handler func(isc.Request)) *events.Cl
 }
 
 func (m *mempool) hasBeenProcessed(reqID isc.RequestID) (hasBeenProcessed bool) {
+	// this will need to be refectored for the new immutable state implementation
 	var err error
 	err = optimism.RetryOnStateInvalidated(
 		func() error {
@@ -360,14 +362,56 @@ func (m *mempool) Empty() bool {
 	return len(m.pool) == 0
 }
 
+func (m *mempool) getProcessedReqsFromTo(from, to *isc.AliasOutputWithID) []isc.RequestID {
+	// this will need to be refectored for the new immutable state implementation
+	ret := []isc.RequestID{}
+
+	fromIdx := uint32(0)
+	if from != nil {
+		fromIdx = from.GetStateIndex()
+	}
+	toIdx := to.GetStateIndex()
+	// from+1 ~ to
+	blockIndexes := make([]uint32, toIdx-fromIdx)
+	for i := range blockIndexes {
+		blockIndexes[i] = fromIdx + uint32(i+1)
+	}
+
+	for _, idx := range blockIndexes {
+		stateReadErr := optimism.RetryOnStateInvalidated(
+			func() error {
+				reqIds, err := blocklog.GetRequestIDsForBlock(m.stateReader.KVStoreReader(), idx)
+				ret = append(ret, reqIds...)
+				return err
+			},
+		)
+		if stateReadErr != nil {
+			m.log.Warnf("error while checking GetRequestIDsForBlock,  err:%s", stateReadErr.Error())
+		}
+	}
+	return ret
+}
+
 const checkForRequestsInPoolInterval = 200 * time.Millisecond
 
 // ConsensusProposalsAsync returns a list of requests to be sent as a batch proposal
 func (m *mempool) ConsensusProposalsAsync(ctx context.Context, aliasOutput *isc.AliasOutputWithID) <-chan []*isc.RequestRef {
 	// TODO handle reorgs (if possible, TBD)
 
-	// !!!!
-	// TODO use aliasOutput to clean processed requests
+	if aliasOutput.GetStateIndex() == 0 {
+		m.lastSeenChainOutput = aliasOutput
+	} else {
+		// clean mempool from requests processed since lastSeenChainOutput until aliasOutput
+		lastSeenStateIndex := uint32(0)
+		if m.lastSeenChainOutput != nil {
+			lastSeenStateIndex = m.lastSeenChainOutput.GetStateIndex()
+		}
+		if aliasOutput.GetStateIndex() < lastSeenStateIndex {
+			panic(fmt.Sprintf("reorg happened, last seen: %s, received: %s", m.lastSeenChainOutput.String(), aliasOutput.String()))
+		}
+		processedReqs := m.getProcessedReqsFromTo(m.lastSeenChainOutput, aliasOutput)
+		m.RemoveRequests(processedReqs...)
+	}
 
 	retChan := make(chan []*isc.RequestRef, 1)
 	go func() {
--- a/packages/chain/aaa2/mempool/mempool.go
+++ b/packages/chain/aaa2/mempool/mempool.go
@@ -218,11 +218,11 @@ func (m *mempool) addTimelockedRequestsToMempool() {
 			// try add To pool
 			func() {
 				m.poolMutex.Lock()
-				defer m.poolMutex.Unlock()
 				for id, req := range nextUnlockReqs {
 					m.addToPoolNoLock(req)
 					delete(timelockedRequests, id)
 				}
+				m.poolMutex.Unlock()
 			}()
 			// find the next set of requests to be unlockable
 			nextUnlock = time.Time{}
--- a/packages/chain/aaa2/mempool/mempool.go
+++ b/packages/chain/aaa2/mempool/mempool.go
@@ -16,11 +16,8 @@ import (
 	"github.com/iotaledger/wasp/packages/gpa"
 	"github.com/iotaledger/wasp/packages/hashing"
 	"github.com/iotaledger/wasp/packages/isc"
-	"github.com/iotaledger/wasp/packages/kv/optimism"
 	metrics_pkg "github.com/iotaledger/wasp/packages/metrics"
 	"github.com/iotaledger/wasp/packages/peering"
-	"github.com/iotaledger/wasp/packages/state"
-	"github.com/iotaledger/wasp/packages/vm/core/blocklog"
 )
 
 const (
@@ -29,7 +26,6 @@ const (
 
 type mempool struct {
 	chainAddress           iotago.Address
-	stateReader            state.OptimisticStateReader
 	lastSeenChainOutput    *isc.AliasOutputWithID
 	poolMutex              sync.RWMutex
 	timelockedRequestsChan chan isc.OnLedgerRequest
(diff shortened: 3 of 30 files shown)


packages/chain/aaa2/chainMgr/chainMgr.go
```diff
@@ -197,7 +197,7 @@ func (cmi *chainMgrImpl) Input(input gpa.Input) gpa.OutMessages {
 	case *inputConsensusTimeout:
 		return cmi.handleInputConsensusTimeout(input)
 	}
-	panic(xerrors.Errorf("unexpected input %T: %+v", input, input))
+	panic(fmt.Errorf("unexpected input %T: %+v", input, input))
```
kape1395, line 200: Why have you changed errors -> fmt?

packages/vm/core/blocklog/external.go
```diff
@@ -80,7 +80,7 @@ func GetRequestRecordDataByRequestID(stateReader kv.KVStoreReader, reqID isc.Req
 		if err != nil {
 			return nil, err
 		}
-		if rec.Request.ID().Equals(reqID) {
+		if rec.Request.ID() == reqID {
```
kape1395: Julius is changing those to Equals, you are changing that back.
I think the Equals variant survives refactorings better.

packages/vm/core/blocklog/internal.go
```diff
@@ -160,7 +160,7 @@ func getCorrectRecordFromLookupKeyList(partition kv.KVStoreReader, keyList Reque
 		if err != nil {
 			return nil, err
 		}
-		if rec.Request.ID().Equals(*reqID) {
+		if rec.Request.ID() == *reqID {
```
kape1395: Julius is changing those to Equals, you are changing that back.
I think the Equals variant survives refactorings better.

packages/chain/aaa2/mempool/mempoolgpa/messages.go
```diff
@@ -0,0 +1,108 @@
...
+		OutMessages: gpa.NoMessages().Add(msg),
+		Staggered:   false,
+	}
+}
+
+// ----------------------------------------------------------------
+
+const (
+	msgTypeShareRequest   byte = iota
+	msgTypeMissingRequest byte = iota
```
kape1395: That's OK to mark both constants with iota?

packages/chain/aaa2/mempool/mempoolgpa/mempool_gpa.go
```diff
@@ -0,0 +1,152 @@
...
+	askMissingReqInterval = 300 * time.Millisecond
+)
+
+type Impl struct {
+	receiveRequests         func(reqs ...isc.Request) []bool
+	getRequest              func(id isc.RequestID) isc.Request
+	hasRequestBeenProcessed func(id isc.RequestID) bool
+	committeeNodes          []gpa.NodeID
+	accessNodes             []gpa.NodeID
+	peersMutex              sync.RWMutex
```
kape1395: I was trying to keep the algorithmic part free of threading, locking, etc.

packages/chain/aaa2/mempool/mempoolgpa/mempool_gpa_test.go
```diff
@@ -0,0 +1,147 @@
+package mempoolgpa
```
kape1395, line 1: gpa.TestContext would emulate a cluster and would share all the messages between the nodes.
That would test the "distributed part" of the algorithm.

packages/chain/aaa2/mempool/mempoolgpa/messages.go
```diff
@@ -0,0 +1,108 @@
+package mempoolgpa
+
+import (
+	"time"
+
+	"github.com/iotaledger/wasp/packages/gpa"
+	"github.com/iotaledger/wasp/packages/isc"
+)
+
+type MempoolMessages struct {
```
kape1395: My impression was that the selection and periodic delivery should be done by the gpa part. Violating that probably forced to have the MempoolMessages structure. Not sure, if that makes things simpler.

packages/chain/aaa2/mempool/mempoolgpa/mempool_gpa.go
```diff
@@ -0,0 +1,152 @@
...
+		func() bool {
+			reqHasLeftTheMempool := m.getRequest(req.ID()) == nil
+			return reqHasLeftTheMempool
+		})
+}
+
+func (m *Impl) Input(input gpa.Input) *MempoolMessages {
+	switch inp := input.(type) {
+	case isc.Request:
+		return m.NewShareRequestMessages(inp, gpa.NodeID(""))
```
kape1395: Sounds suspicious to have empty node ids.

packages/chain/aaa2/mempool/mempoolgpa/mempool_gpa.go
```diff
@@ -0,0 +1,152 @@
...
+	shareReqNNodes   = 1
+	shareReqInterval = 1 * time.Second
+)
+
+const (
+	askMissingReqNNodes   = 2
+	askMissingReqInterval = 300 * time.Millisecond
+)
+
+type Impl struct {
```
kape1395, line 27: It don't actually implement the GPA interface.

packages/chain/aaa2/mempool/interface.go
```diff
@@ -0,0 +1,30 @@
...
+	ReceiveRequests(reqs ...isc.Request) []bool
+	RemoveRequests(reqs ...isc.RequestID)
+	HasRequest(id isc.RequestID) bool
+	GetRequest(id isc.RequestID) isc.Request
+	Info(currentTime time.Time) MempoolInfo
+	Close()
+}
+
+// for testing (only for use in solo)
+type SoloMempool interface {
```
kape1395: Maybe that "wrapper" for the solo tests should be moved to the solo package?

packages/chain/aaa2/mempool/interface.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+type Mempool interface {
+	consGR.Mempool // TODO should this be unified with the Mempool interface
+	ReceiveRequests(reqs ...isc.Request) []bool
+	RemoveRequests(reqs ...isc.RequestID)
+	HasRequest(id isc.RequestID) bool
+	GetRequest(id isc.RequestID) isc.Request
+	Info(currentTime time.Time) MempoolInfo
+	Close()
```
kape1395: Maybe we can use contexts for closing components. Then the Close() function would be not needed.

packages/chain/aaa2/mempool/mempool_test.go
```diff
@@ -0,0 +1,447 @@
+package mempool
```
kape1395, line 1: These tests don't check the distributed aspects of the algorithm either.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+			return
+		}
+		// process the incoming message and send whatever is needed to other peers
+		pool.sendNetworkMessages(pool.gpa.Message(msg))
+	})
+	pool.netDisconnect = func() {
+		net.Detach(attachID)
+	}
+
+	go pool.addTimelockedRequestsToMempool()
```
kape1395, line 103: Maybe the time has to come from the consensus with the request for proposals?
Or that would complicate things here?

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+
+	return pool
```
kape1395, line 107: Was expecting to have a run-thread here. That thread would handle all the communication in the component, making the main algoritghm a single threaded.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+// TODO this must be called from chainMGR
+func (m *mempool) SetPeers(committee []gpa.NodeID, accessNodes []gpa.NodeID) {
+	m.gpa.SetPeers(committee, accessNodes)
```
kape1395, line 112: I would send to the main mempool thread via pipe or a channel to keep the main alg in a single thread only,

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func (m *mempool) sendNetworkMessages(outMsgs *mempoolgpa.MempoolMessages) {
```
kape1395: This also probably runs in the peering's thread.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+		peerMsg := &peering.PeerMessageData{
+			PeeringID:   m.netPeeringID,
+			MsgReceiver: peering.PeerMessageReceiverChainCons,
+			MsgType:     msgTypeMempool,
+			MsgData:     msgData,
+		}
+		m.net.SendMsgByPubKey(m.peers[msg.Recipient()], peerMsg)
+	}
+	if !outMsgs.Staggered {
```
kape1395: Still unable to bet used to the word "Staggered" :)

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+		outMsgs.MustIterate(sendMsg)
+		return
+	}
+	sent := 0
+	outMsgs.MustIterate(func(msg gpa.Message) {
```
kape1395: I would leave this logic in the main alg.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+func (m *mempool) isRequestReady(req isc.Request) bool {
+	if onLedgerReq, ok := req.(isc.OnLedgerRequest); ok {
+		return isc.RequestIsUnlockable(onLedgerReq, m.chainAddress, time.Now())
+	}
+	return true
+}
+
+func (m *mempool) Close() {
+	m.netDisconnect()
+	m.ctx.Done()
```
kape1395: This only returns a channel notifying, when the context is closed by its issuer.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+			if timelock == nil {
+				panic("request without timelock shouldn't have been added here")
+			}
+			unlockTime := time.Unix(int64(timelock.UnixTime), 0)
+			if nextUnlock.IsZero() || unlockTime.Before(nextUnlock) {
+				nextUnlock = unlockTime
+				nextUnlockReqs = make(map[isc.RequestID]isc.Request)
+				nextUnlockReqs[req.ID()] = req
+			}
+			if unlockTime.Equal(nextUnlock) {
```
kape1395, line 227: Is it OK to use Equal here? Not the After and similar?
Looks like there is low chance to match the timestamps if they are in nanosecond precision.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+				nextUnlockReqs[req.ID()] = req
+			}
+
+		case <-time.After(timeUntilNextUnlock):
+			// try add To pool
+			func() {
+				m.poolMutex.Lock()
+				defer m.poolMutex.Unlock()
```
kape1395: Maybe just unlock after the loop?

packages/chain/aaa2/mempool/mempool.go
kape1395: OK, that probably reasonable considering how the nextUnlock is set.
I wold do that in some time intervals, not for each message separately. Not sure, which approach is better.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+		if _, ok := m.pool[rid]; !ok {
+			continue
+		}
+		m.removeFromPoolNoLock(rid)
+	}
+}
+
+// ConsensusProposalsAsync returns a list of requests to be sent as a batch proposal
+func (m *mempool) ConsensusProposalsAsync(ctx context.Context, aliasOutput *isc.AliasOutputWithID) <-chan []*isc.RequestRef {
+	// TODO ctx not used here, nor is aliasOutput
```
kape1395: That means rollbacks/rejections are not supported right now...

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+	retChan := make(chan []*isc.RequestRef, 1)
+
+	go func() {
```
kape1395, line 385: I would try to keep a single goroutine per mempool. Seen to much of deadlocks in our codebase.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+			}
+			if m.isRequestReady(req) {
+				ret = append(ret, &isc.RequestRef{
+					ID:   req.ID(),
+					Hash: isc.RequestHash(req),
+				})
+			}
+		}
+		retChan <- ret
+		go m.RemoveRequests(toRemove...)
```
kape1395: Another thread is started from a thread?
Hard to imagine, what race conditions are caused by that.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+	}()
+
+	return retChan
```
kape1395, line 416: The mempool has to wait in this function, if no (or to few) requests exists in the pool.
The wait should be with some upper bound ofc.
Otherwise consensus will run in a loop without delays.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+		requests, missingReqs := m.getRequestsFromRefs(requestRefs)
+		if len(missingReqs) == 0 {
+			// we have all the requests
+			retChan <- requests
+			return
+		}
+		var missingRequestsChan chan isc.Request
+		for _, idx := range missingReqs {
+			missingRef := requestRefs[idx]
+			go m.sendNetworkMessages(
```
kape1395: Would like to avoid starting those goroutines everywhere...
And we have to message redelivery in this scenario.

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,474 @@
...
+				m.gpa.Input(missingRef),
+			)
+		}
+		closure := m.attachToIncomingRequests(func(req isc.Request) {
```
kape1395: what if the responses to the previously sent `m.gpa.Input(missingRef)` are received before this call?

kape1395 (commented):
Most comments are for discussion probably.

Maybe a critical part is to add a delay before responding to consensus proposal request if mempool is low.

For the remaining parts, maybe I can try to use as a blackbox for now.

packages/chain/aaa2/chainMgr/chainMgr.go
jorgemmsilva: afaik `fmt` is the "more standard" way to do it nowadays

packages/vm/core/blocklog/external.go
jorgemmsilva: sure, I'll keep the `Equals` func. I changed the RequestID to be a fixed size array so it can be used as keys in maps

packages/chain/aaa2/mempool/mempoolgpa/messages.go
jorgemmsilva: oh snap, nice find!

packages/chain/aaa2/mempool/mempoolgpa/mempool_gpa.go
jorgemmsilva: its just to pass to `NewShareRequestMessages` as the `receivedFrom` node, in this case there is no such node, as the request was received from the webapi

packages/chain/aaa2/mempool/mempoolgpa/mempool_gpa.go
jorgemmsilva: yeah, I ended up removing the GPA interface, because the Ouput part would not be used, and it would mean I couldn't return a custom message type (the "staggered messages")

packages/chain/aaa2/mempool/interface.go
jorgemmsilva: sure, that can be done for sure :+1: , good point

packages/chain/aaa2/mempool/mempool.go
jorgemmsilva: I don't see the value of doing it that way, but happy to discuss in a call

packages/chain/aaa2/mempool/mempool.go
jorgemmsilva: what would be the advantage of having the mempool being strictly single thread?

packages/chain/aaa2/mempool/mempool.go
jorgemmsilva: would something like "throttle messages" be more clear?

packages/chain/aaa2/mempool/mempool.go
jorgemmsilva: sure, we can add some waiting interval, and create a definition for "minimum N requests"

packages/chain/aaa2/mempool/mempool.go
jorgemmsilva: god point, I can do the "subscription" before sending the messages :+1:

packages/chain/aaa2/mempool/mempool.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+var _ consGR.Mempool = &mempool{}
+
+func New(
+	ctx context.Context,
+	chainID *isc.ChainID,
+	myNodeID gpa.NodeID,
+	net peering.NetworkProvider,
+	hasBeenProcessed HasBeenProcessedFunc,
```
kape1395, line 55: Maybe I can pass the state manager here instead of this?
```
	hasBeenProcessed HasBeenProcessedFunc,
	getProcessedRequests GetProcessedRequestsFunc,
```
In the current implementation ChainNode should handle that. Would like to keep that in the mempool.

jorgemmsilva: This was to make mocking those functions in the tests easier, but we could make it an interface with those 2 methods for example.

kape1395: For me OK to have callbacks. OK to have an inter face as well.
The problem is that those functions have to be implemented.
I think the current variant is hardly to implementable at all.

```
HasBeenProcessedFunc     func(reqID isc.RequestID) bool
```
You have to know the state to check that. And maybe it is better to remember that state in the mempool.

```
GetProcessedRequestsFunc func(from, to *isc.AliasOutputWithID) []isc.RequestID
```
Should it return all the history of processed requests till that alias output? Would be a lot.

kape1395: StateMgr has its interface. You could mock it as well.

jorgemmsilva: Imo that logic should be implemented in state mgr

PR #1413, opened 2022-11-08
evm!: add ERC20 proxy for native tokens

dessaya:
- The ERC20 contract for a foundry must be registered by the foundry owner, by calling the new `registerERC20NativeToken` endpoint of the `evm` core contract.

- BREAKING CHANGE: Reorganized ISC magic contract definitions in Solidity:

  - ISC magic interface methods are now categorized into sub-interfaces `ISCSandbox`, `ISCAccounts`, `ISCUtil`, etc.

  - Instead of accessing magic methods with `isc.*`, now it's `ISC.sandbox.*`, `ISC.accounts.*`, etc. E.g.: `ISC.sandbox.getChainID()`.

Internals:

- Refactored handlers in EVM magic contract: replaced the large `switch` with reflection + automatic type conversions. Now there is a distinct Go method for each Solidity method.

- Got tired of implementing a map function for the umpteenth time, so I added dependency to `samber/lo` which "is a Lodash-style Go library", providing generic functions like `Map`, `Filter`, `Reduce`, etc.


packages/vm/core/evm/evmimpl/iscmagic_handler.go
```diff
@@ -0,0 +1,87 @@
...
+		caller: caller,
+	}, method, args)
+}
+
+func titleCase(s string) string {
+	return strings.ToUpper(s[:1]) + s[1:]
+}
+
+func reflectCall(handler any, method *abi.Method, args []any) []byte {
+	handlerMethod := reflect.ValueOf(handler).MethodByName(titleCase(method.Name))
```
jorgemmsilva, line 47: could this be exploited somehow to crash the node/ cause unpredictable behavior?

packages/vm/core/evm/iscmagic/embed.go
```diff
@@ -45,13 +62,32 @@ var (
...
 	copy(ret[0:2], AddressPrefix)
 	ret[2] = kind
-	copy(ret[3:], data)
+	copy(ret[3:], payload)
 	return ret
 }
+
+func ERC20NativeTokensFoundrySN(addr common.Address) uint32 {
+	payload := addr[3:]
+	return codec.MustDecodeUint32(payload[:4])
```
jorgemmsilva:
```suggestion
	return codec.MustDecodeUint32(addr[3:7])
```

jorgemmsilva (approved):

packages/vm/core/evm/evmimpl/iscmagic_handler.go
dessaya: Uh... I don't imagine how, but then I'm not a security expert :stuck_out_tongue: 

At least I don't think it is possible to crash the node here, since all panics are recovered en `catchISCPanics`.

There is also the possibility of attempting to call a "privileged" method, but that is handled in `parsedCall`.

PR #1429, opened 2022-11-14
fix: (wasp-cli) deposit to another AgentID

jorgemmsilva:


client/client.go
```diff
@@ -62,6 +63,11 @@ func processResponse(res *http.Response, decodeTo interface{}) error {
 }
 
 func (c *WaspClient) do(method, route string, reqObj, resObj interface{}) error {
+	if c.baseURL == "" {
+		log.Printf("wasp webapi not defined, request to %s not done", route)
+		return nil
```
dessaya: Why not return an error here?

Actually shoudn't the `WaspClient` constructor panic instead if `baseURL` is empty?

tools/wasp-cli/config/config.go
```diff
@@ -160,21 +160,21 @@ func SetToken(token string) {
 
 func WaspClient(i ...int) *client.WaspClient {
 	// TODO: add authentication for /adm
-	log.Verbosef("using Wasp host %s\n", WaspAPI())
 	L1Client() // this will fill parameters.L1() with data from the L1 node
-	return client.NewWaspClient(WaspAPI(i...)).WithToken(GetToken())
+	apiAddress := WaspAPI(i...)
+	if apiAddress == "" {
+		return &client.WaspClient{}
```
dessaya: I'm confused... How does this work? I mean a `WaspClient` with no `baseURL`? :thinking:

jorgemmsilva: I wanted to return nil, instead of using a made up url as it was before. But that would cause a bunch of nil pointer exceptions on the cli code.

client/client.go
jorgemmsilva: I didn't wanna error because all over wasp cli this gets called, just wanted to print that no calls are being made because the user didn't specify a webapi url

dessaya: But this doesn't belong exclusively to wasp-cli, it is the generic wasp client object. It could be used anywhere, so printing to the console is not correct :-\

tools/wasp-cli/config/config.go
dessaya: But why not panic? Isn't this an error condition?

client/client.go
jorgemmsilva: Fair enough, I'll just return nil on the constructor for empty url and refactor usages 👍

dessaya: Sorry for insisting, but why not panic? Maybe I'm missing something but calling a constructor that returns nil sounds like a smell :grimacing: .

tools/wasp-cli/config/config.go
jorgemmsilva: No, this can be used to send funds to any chain where you don't have direct access to the API or didn't fill in the config.json for whatever reason

client/client.go
jorgemmsilva: >sounds like a smell

:smile: 

cli code calls: `client := config.WaspClient()` when trying to query an api. What should that return when there is no such api defined in `wasp-cli.json`? 

I guess the design flaw here is that this shouldn't receive an optional index and try to find the config after the fact, but rather just receive it as a parameter in the constructor:
https://github.com/iotaledger/wasp/blob/develop/tools/wasp-cli/config/config.go#L161

dessaya: Yes, I see you already changed that :+1: 

So now this condition should never happen?

```
if c.baseURL == ""
```

It's safe to remove this check or panic?

dessaya (approved):

PR #1431, opened 2022-11-15
Refactor registries (json format)

### LINKED ISSUE #1411: Clean up ChainRegistry to only use json format

muXxer:
This PR changes the way informations are stored in wasp.

The possibility to store configuration data either in a database or a text file was removed.
Chain registry, Trusted Peers, and DKShares are now stored as a JSON file. 
The TextKVStore was removed, because we use direct JSON marshalling now.
This improves the administration of the wasp node, because the data is now human readable.

The PR also fixes some bugs in the `Clone` methods.


.golangci.yml
```diff
@@ -50,7 +50,7 @@ issues:
 
 linters-settings:
     funlen:
-        lines: 100
+        lines: 150
```
jorgemmsilva: why, tho? :joy: 
100 lines is plenty

core/database/component.go
```diff
@@ -34,13 +40,15 @@ type dependencies struct {
 	dig.In
 
 	DatabaseManager *dbmanager.DBManager
+	// TODO: remove this in the database refactor PR
```
jorgemmsilva, line 43: ?

packages/cryptolib/public_key.go
```diff
@@ -52,6 +52,30 @@ func NewPublicKeyFromBytes(publicKeyBytes []byte) (*PublicKey, error) {
 	return &PublicKey{publicKeyBytes}, nil
 }
 
+func PublicKeyToHex(pubKey *PublicKey) string {
+	return iotago.EncodeHex(pubKey.AsBytes())
```
jorgemmsilva, line 55: we use `hexutil.Encode` everywhere, please use that as well (I know this is just a wrapper, but for consistency sake)

packages/database/dbkeys/dbkeys.go
```diff
@@ -5,18 +5,18 @@ import (
...
-	ObjectTypeNodeIdentity
-	ObjectTypeBlobCache
-	ObjectTypeBlobCacheTTL
-	ObjectTypeTrustedPeer
-	ObjectTypeConsensusJournal
+	ObjectTypeNodeIdentity     // deprecated
+	ObjectTypeBlobCache        // unused
+	ObjectTypeBlobCacheTTL     // unused
+	ObjectTypeTrustedPeer      // deprecated
+	ObjectTypeConsensusJournal // deprecated
```
jorgemmsilva, line 19: just remove them?

jorgemmsilva (approved):
minor comments, please run some of the cluster tests before merging, just to make sure everything is working as expected :pray:

core/database/component.go
muXxer: This is my upcoming PR, I split it up into two different PRs, because it would have been even bigger than it is now.
I only added this stuff here so the wasp node still works in the meantime.

packages/database/dbkeys/dbkeys.go
muXxer: I wanted to leave them here in case we want to convert current databases to the new format (e.g. for already running chains).

packages/cryptolib/public_key.go
muXxer: But using hexutil directly results in `0x` if you call it for an empty slice.
Thats why this check in iota.go was added:

```go

// EncodeHex encodes the bytes string to a hex string. It always adds the 0x prefix if bytes are not empty.
func EncodeHex(b []byte) string {
	if len(b) == 0 {
		return ""
	}
	return hexutil.Encode(b)
}
```
Maybe we should replace usage of `hexutil.Encode` with `iotago.EncodeHex` everywhere instead.

muXxer: I changed it to iotago.EncodeHex/DecodeHex everywhere now.

PR #1433, opened 2022-11-16
use IterateKeys from trie.go

dessaya:


packages/state/kv.go
```diff
@@ -31,8 +31,7 @@ func (t *trieKVAdapter) Iterate(prefix kv.Key, f func(kv.Key, []byte) bool) erro
 }
 
 func (t *trieKVAdapter) IterateKeys(prefix kv.Key, f func(kv.Key) bool) error {
-	// TODO: use IterateKeys
-	t.TrieReader.Iterator([]byte(prefix)).Iterate(func(k []byte, v []byte) bool {
+	t.TrieReader.Iterator([]byte(prefix)).IterateKeys(func(k []byte) bool {
```
muXxer, line 34: Sorry I'm only on mobile currently. But doesn't the `IterateKeys` return an error which we can forward instead of returning `nil`. Same problem in the function `Iterate` above.

dessaya: No, the `trie.go` package returns void, making the assumption that an I/O error is unrecoverable and panics instead. These functions return `error` just to conform to Wasp's `kv` interfaces. We should probably at some point refactor our `kv` also removing the `error` return values.

muXxer: :+1:

muXxer (approved):

PR #1435, opened 2022-11-18
feat: add access nodes (gpa)

jorgemmsilva:


packages/registry/providers.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/iotaledger/wasp/packages/tcrypto"
 )
 
+// TODO why does this exist?
```
muXxer: Yeah was wondering the same :)
We should get rid of these provider functions in the whole codebase imo.

packages/webapi/admapi/accessnodes.go
```diff
@@ -0,0 +1,108 @@
...
+
+	peers, err := a.networkMgr.TrustedPeers()
+	if err != nil {
+		return httperrors.ServerError("error getting trusted peers")
+	}
+	_, ok := lo.Find(peers, func(p *peering.TrustedPeer) bool {
+		return p.PubKey.Equals(pubKey)
+	})
+	if !ok {
+		return httperrors.NotFound(fmt.Sprintf("couldn't find peer with public key %s", pubKey))
```
muXxer, line 67: Sorry, am on mobile that's why it's hard to check, but does the public key has a stringer interface? If not, we should convert to hex here :)
If it does, ignore this comment 😂

packages/webapi/admapi/accessnodes.go
```diff
@@ -0,0 +1,108 @@
...
+	adm echoswagger.ApiGroup,
+	registryProvider registry.Provider,
+	tnm peering.TrustedNetworkManager,
+) {
+	a := &accessNodesService{
+		registry:   registryProvider,
+		networkMgr: tnm,
+	}
+	adm.POST(routes.AdmAddAccessNode(":chainID"), a.handleAddAccessNode).
+		AddParamPath("", "chainID", "ChainID (string)").
```
muXxer: Should we write bech32 as a hint instead of string here?

packages/webapi/admapi/accessnodes.go
```diff
@@ -0,0 +1,108 @@
...
+		AddParamPath("", pubKeyParam, "PublicKey (hex string)").
+		SetSummary("Add an access node to a chain")
+
+	adm.POST(routes.AdmRemoveAccessNode(":chainID"), a.handleRemoveAccessNode).
+		AddParamPath("", "chainID", "ChainID (string)").
```
muXxer: Should we write bech32 as a hint instead of string here?

packages/webapi/admapi/accessnodes.go
muXxer:
```go

func (pkT *PublicKey) String() string {
	return hexutil.Encode(pkT.key)
}
```
seems like it :runner: 
so ignore this one :trollface:

packages/webapi/admapi/accessnodes.go
jorgemmsilva: it should be a bech32 string, but good point - should be more explicit

muXxer: yeah it is, I checked it. but the hint would be nice, so we don't have to look at the code ourselves all the time :trollface:

PR #1439, opened 2022-11-20
Refactor local PoW and submitting of blocks

### LINKED ISSUE #1440: Refactor local PoW and submitting of blocks

muXxer:
This PR uses the PoW code from inx-app, there is no need to copy that code to wasp itself.

New functions are added to the `l1connection.Client` interface:
```diff
+// sends a block (including tipselection and local PoW if necessary)
+PostBlock(block *iotago.Block, timeout ...time.Duration) (*iotago.Block, error)

-// sends a tx, waits for confirmation
+// sends a tx (including tipselection and local PoW if necessary)
PostTx(tx *iotago.Transaction, timeout ...time.Duration) (*iotago.Block, error)

+// sends a tx (including tipselection and local PoW if necessary) and waits for confirmation
+PostTxAndWaitUntilConfirmation(tx *iotago.Transaction, timeout ...time.Duration) (*iotago.Block, error)
```

The function that waits for block confirmation was fixed, because it could potentially "deadlock" if a block needs a reattachment.
We didn't update the tracked `blockID` after reattachments of the block with identical payload were sent.

This PR also introduces a cooldown time for block promotions. Otherwise blocks that need to be promoted, would be promoted in every cycle of the loop (every 200ms).

Another problem was that we cleared potential necessary parents in the block if we do remote PoW.
In case of remote PoW we also submitted the same block twice (inside the `doBlockPow` and again with `SubmitBlock`).


packages/l1connection/l1connection.go
```diff
@@ -30,8 +35,12 @@ type Config struct {
...
 	// requests funds from faucet, waits for confirmation
 	RequestFunds(addr iotago.Address, timeout ...time.Duration) error
-	// sends a tx, waits for confirmation
-	PostTx(tx *iotago.Transaction, timeout ...time.Duration) error
+	// sends a block (including tipselection and local PoW if necessary)
+	PostBlock(block *iotago.Block, timeout ...time.Duration) (*iotago.Block, error)
+	// sends a tx (including tipselection and local PoW if necessary)
+	PostTx(tx *iotago.Transaction, timeout ...time.Duration) (*iotago.Block, error)
+	// sends a tx (including tipselection and local PoW if necessary) and waits for confirmation
+	PostTxAndWaitUntilConfirmation(tx *iotago.Transaction, timeout ...time.Duration) (*iotago.Block, error)
```
jorgemmsilva, line 43:
```suggestion
	PostTxAndWaitUntilConfirmation(tx *iotago.Transaction, timeout ...time.Duration) (*iotago.Block, error)
```
seems only `PostTxAndWaitUntilConfirmation` is used, no need to expose the other funcions

jorgemmsilva (approved):

packages/l1connection/l1connection.go
muXxer: Oh sorry, automerge was enabled :sweat_smile: 
Should I remove these in the next PR?

jorgemmsilva: yeah, just noticed after approving. sure :+1:

PR #1443, opened 2022-11-21
Fix cluster tests

muXxer:
This PR fixes some problems in the cluster tests on current develop branch.


tools/wasp-cli/log/log.go
```diff
@@ -39,6 +39,7 @@ func HiveLogger() *logger.Logger {
 			DisableEvents:     true,
 			DisableCaller:     true,
 			DisableStacktrace: true,
+			StacktraceLevel:   "panic",
```
jorgemmsilva, line 42: why?

jorgemmsilva (approved):

tools/wasp-cli/log/log.go
muXxer: We introduced that at some point in the hive logger, and it was not set here.
So it printed stracktraces even at "info" level in some cases. (l1connection tests)

PR #1461, opened 2022-11-25
trie: remove VCommitment, TCommitment interfaces

dessaya:
`trie.VCommitment` is now called `trie.Hash` and is just a `[20]byte` array. `Equals()` is no longer needed (just use `==` instead).


packages/trie/nodedata.go
```diff
@@ -235,13 +230,42 @@ func (n *nodeData) Read(r io.Reader) error {
...
+func (n *nodeData) update(childUpdates map[byte]*Hash, newTerminalUpdate *tcommitment, pathExtension []byte) {
+	for i, upd := range childUpdates {
+		n.children[i] = upd
+	}
+	n.terminal = newTerminalUpdate // for hash commitment just replace
+	n.pathExtension = pathExtension
+	if n.ChildrenCount() == 0 && n.terminal == nil {
+		return
+	}
+	n.commitment = n.makeHashVector().Hash()
```
muXxer, line 251: I don't know how this commitment stuff works, but shouldn't it be set to nil or zero value in case someone updates the children count to 0 and the terminal to nil?

packages/trie/terminal.go
```diff
@@ -0,0 +1,147 @@
...
+		// empty slice -> no data (deleted)
+		return nil
+	}
+	var commitmentBytes []byte
+	var isValue bool
+
+	if len(data) > tcommitmentDataSizeMax {
+		isValue = false
+		// taking the hash as commitment data for long values
+		hash := blake2b160(data)
```
muXxer, line 45: Do we know what was the reasoning we use blake2b160 in trie.go?
There are already discussions to increase the hash size in the ETH community (also blake2b160 afaik).

Thats why we also increased the hash size for native token IDs at the end, because luca decided we will use the same hash size blake2b256 everywhere.

Is there a special reason to use 160 here?

packages/trie/nodedata.go
dessaya: Nice catch! This was actually a bug. I added a test and updated this function.

packages/trie/terminal.go
dessaya: It was chosen as a result of some benchmarks, some of this still available in the trie.go repository: https://github.com/iotaledger/trie.go/blob/master/README.md

muXxer (approved):

PR #1465, opened 2022-11-28
refactor: move EVMGasRatio to gov contract

jorgemmsilva:
moves EVM gas ratio so it belongs to the feepolicy object. This way everything about gas and fees is on a single place, also makes it easier to retrieve all needed information to craft requests.


dessaya (approved):
Shouldn't the documentation also be updated?
`documentation/docs/guide/core_concepts/core_contracts/evm.md`

jorgemmsilva:
Good point, thanks @dessaya 👍

PR #1467, opened 2022-11-28
Refactor DatabaseManager

### LINKED ISSUE #1392: Implement database health tracker
- Reuse the hornet database health manager

### LINKED ISSUE #1396: Implement database engine check
We need a way to check if the given database engine in the config even matches the database folder.

muXxer:
This PR adds the database health trackers from hornet to the wasp codebase.

Databases are checked for corruption at startup. If the databases are verified as clean, they are temporary marked as corrupted.
During a clean shutdown of the wasp node, the corrupted flag is removed from the database and the databases are flushed and closed. If the wasp node didn't shut down properly or paniced during execution, the corrupted flag still remains in the database.
This is used to detect a clean shutdown of the wasp node. The database health trackers also stored the version of the used database layout. If the codebase implements a newer version of the database layout that is used in the database itself, a function is executed to migrate to the new database layout version.

The PR also adds the `db.engine` flag. This is used to specify the used database engine. Wasp now supports `mapDB` for in memory databases and `rocksdb`. For file based databases, a database info file is stored in the database folder at creating. This way the correct engine for the created database files can be check during startup.

`RocksDB` is now opened with custom parameters for better compaction and concurrent access.
```go
rocksdb.IncreaseParallelism(runtime.NumCPU() - 1),
rocksdb.Custom([]string{
	"periodic_compaction_seconds=43200",
	"level_compaction_dynamic_level_bytes=true",
	"keep_log_file_num=2",
	"max_log_file_size=50000000", // 50MB per log file
}),
```

Two little helper scripts were added to run `go mod tidy` in the codebase and to run the `go test` including cluster tests and dump the results into a file `tests_outputs.log`


core/database/params.go
```diff
@@ -4,16 +4,30 @@ import (
...
+		Path string `default:"waspdb/chains/consensus" usage:"the path to the consensus journal database folder"`
+	}
+
+	ChainState struct {
+		// Path defines the path to the chain state databases folder.
+		Path string `default:"waspdb/chains/data" usage:"the path to the chain state databases folder"`
+	}
+
+	// Debug defines whether to ignore the check for corrupted databases (should only be used for debug reasons).
+	Debug bool `default:"false" usage:"ignore the check for corrupted databases (should only be used for debug reasons)"`
```
jorgemmsilva: `Debug` is not very explicit. I propose something like: `DebugIgnoreDBChecks`

core/peering/params.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 type ParametersPeering struct {
-	NetID string `default:"0.0.0.0:4000" usage:"node host address as it is recognized by other peers"`
```
jorgemmsilva: this was changed to 0's because the localhost IP would cause problems in docker IIRC

scripts/run_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+CURRENT_DIR="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+PARENT_DIR="$( builtin cd ${CURRENT_DIR}/.. >/dev/null 2>&1 ; pwd -P )"
+cd ${PARENT_DIR}
+
+go test -v -timeout=5h ./... 2>&1 | tee tests_output.log
```
jorgemmsilva:
```suggestion
go test -v -failfast -timeout=5h ./... 2>&1 | tee tests_output.log
```

packages/database/database.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, fmt.Errorf("unknown database engine: %s, supported engines: rocksdb/mapdb", dbEngine)
+	}
+}
+
+type databaseWithHealthTracker struct {
+	database           *Database
+	storeHealthTracker *kvstore.StoreHealthTracker
+}
+
+func newDatabaseWithHealthTracker(path string, dbEngine hivedb.Engine, autoFlush bool, storeVersion byte, storeVersionUpdateFunc kvstore.StoreVersionUpdateFunc) (*databaseWithHealthTracker, error) {
```
jorgemmsilva: this stuff with `kvstore.StoreVersionUpdateFunc` feels incomplete, do you have any example of how this would work? maybe on hornet?

jorgemmsilva (approved):

core/database/params.go
muXxer: maybe we should avoid the abbreviation "DB" because the JSON would look like this: `db.debugIgnoreDbChecks`

what about:
`db.debugIgnoreDatabaseChecks`
`db.debugSkipDatabaseChecks`
`db.debugSkipConsistencyCheck`
`db.debugSkipHealthCheck`

Please choose :trollface:

core/peering/params.go
muXxer: Ok will update it :+1:

scripts/run_tests.sh
muXxer: Do we want to fail fast? What if we want to ignore a check and see what else fails (e.g. the IncCounter check that fails now?)

packages/database/database.go
muXxer: good catch, there was a small problem with the kvstore that is created in `newDatabaseWithHealthTracker`, but I pushed a fix.

here a small example how I would imagine such an update:
```go

// load consensus journal database
databaseConsensusJournal, err := newDatabaseWithHealthTracker(m.databasePathConsensusJournal, m.engine, true, StoreVersionConsensusJournal, func(store kvstore.KVStore, oldVersion, newVersion byte) error {
if newVersion != 2 {
	return fmt.Errorf("unknown new store version: %d", newVersion)
}

if oldVersion != 1 {
	return fmt.Errorf("unknown old store version: %d", oldVersion)
}

batched, err := store.Batched()
if err != nil {
	return fmt.Errorf("database layout update failed: %w", err)
}

var innerErr error
if err := store.Iterate([]byte{common.ObjectTypeBlobCache}, func(key, value kvstore.Value) bool {
	// delete old key
	if err := batched.Delete(key); err != nil {
		innerErr = err
		return false
	}

	// move value to new key
	if err := batched.Set([]byte{common.ObjectTypeBlobCacheTTL}, value); err != nil {
		innerErr = err
		return false
	}

	return true
}); err != nil {
	return fmt.Errorf("database layout update failed: %w", err)
}

if innerErr != nil {
	return fmt.Errorf("database layout update failed: %w", innerErr)
}

return batched.Commit()
})
```

core/database/params.go
jorgemmsilva: `db.debugSkipHealthCheck` sounds good :ok_hand:

scripts/run_tests.sh
jorgemmsilva: that's how I usually operate the cluster tests take so long and are so verbose that I just stop at the first failure, then re-run. most tests results are cached anyway.

Feel free to ignore

packages/database/database.go
jorgemmsilva: okay, thanks :+1:  - this feels okay, but a bit ad-hoc, it will get hairy fast if someone needs to update say from version 1 to 3 (more than 1 schema migration).
Maybe we should create some sort of migration system (like Ruby on Rails) where we can run any number of migrations in a row.

muXxer: yeah I mean we are free to do whatever we want in this function. we can also update from 1 to 3 by updating from 1 to 2 first, and then from 2 to 3, without even leaving that function.

It's up to the developer to decide how the migration should be done. Maybe in some cases going from 1 to 3 directly also makes sense.

PR #1472, opened 2022-11-30
[gpa] use value receivers for Hash

dessaya:


packages/state/l1commitment.go
```diff
@@ -132,7 +132,7 @@ func (s *L1Commitment) Read(r io.Reader) error {
 }
 
 func (s *L1Commitment) String() string {
-	return fmt.Sprintf("trie root: %s, block hash: %s", s.trieRoot.String(), s.GetBlockHash())
+	return fmt.Sprintf("trie root: %s, block hash: %s", s.GetTrieRoot(), s.GetBlockHash())
```
kape1395, line 135: Is this change intended?

dessaya: Yes :) Go now automatically calls String()

kape1395: It was not calling it, probably because of the pointer/value receiver mismatch.

PR #1476, opened 2022-12-01
evm: disallow non-view calls with staticcall (fixes #1475)

### LINKED ISSUE #1475: [audit] evm: disallow non-view calls with staticcall
"Ethereum implements a concept of 'static call' which is a security mechanism meant to ensure that everything called via this method, including any sub-calls are read-only and do not modify the state. The ISC magic contract which exposes various ISC functions to EVM does not respect the static call flag and allows for state changes when called from a static call.

For example to trigger a state change directly from a static call:
    function doStaticCall() public {   ISC_MAGIC_ADDRESS.staticcall(abi.encodeWithSignature(""triggerEvent(string)"", ""from_static""));
    }

"

dessaya:


packages/vm/core/evm/evmtest/evm_test.go
```diff
@@ -1350,14 +1350,28 @@ func TestSandboxStackOverflow(t *testing.T) {
...
 }
+
+func TestStaticCall(t *testing.T) {
+	env := initEVM(t)
+	ethKey, _ := env.soloChain.NewEthereumAccountWithL2Funds()
+	iscTest := env.deployISCTestContract(ethKey)
+
+	res, err := iscTest.callFn([]ethCallOptions{{
+		sender:   ethKey,
+		gasLimit: 100_000, // skip estimate gas (which will fail)
```
jorgemmsilva: why does the gas estimation fail?

jorgemmsilva (approved):

packages/vm/core/evm/evmtest/evm_test.go
dessaya: Ah, that's not necessary here. Good catch :+1:

PR #1480, opened 2022-12-01
evm: prevent DoS attack in DecodeLogs (fixes #1479)

### LINKED ISSUE #1479: [audit] evm: prevent DoS attack in DecodeLogs
"There is a potential for denial of service in the DecodeLogs function. It reads a uint32 and then allocates memory based on this value. If an attacker can specify a large (up to 4GB) value then it will allocate a large amount of memory n*sizeof(types.Log). It seems like this is not exploitable right now since I don’t see anywhere where an attacker can specify this encoded Logs value, but it would be good to check that this n is at least smaller than the remainder of bytes in the buffer.
https://github.com/iotaledger/wasp/blob/c0c4e8a56787e2eb13b19783c5514da100fed125/packages/evm/evmtypes/logs.go#L99-L103"

dessaya:


muXxer (approved):

jorgemmsilva (approved):

packages/evm/evmtypes/logs.go
```diff
@@ -100,6 +101,10 @@ func DecodeLogs(b []byte) ([]*types.Log, error) {
 	if err != nil {
 		return nil, err
 	}
+	if int(n) > len(b) {
+		// using len(b) as an upper bound to prevent DoS attack allocating the array
+		return nil, fmt.Errorf("DecodeLogs: invalid amount")
```
jorgemmsilva:
```suggestion
		return nil, fmt.Errorf("DecodeLogs: invalid length")
```

PR #1483, opened 2022-12-01
Refactor node metrics

muXxer:
The metrics package was now renamed to prometheus.

There is now a prometheus registry, that gets passed to all functions where prometheus collectors need to be registed, instead of using the global prometheus way. This is much cleaner and allows us to add other metrics as well.

Additional metrics that got added:
- node metrics (wasp version)
- node connection metrics
- go metrics
- process metrics
- promhttp metrics
- webAPI metrics
- blockWAL metrics


config.json
```diff
@@ -75,9 +75,16 @@
...
     "enabled": true,
-    "bindAddress": "0.0.0.0:2112"
+    "bindAddress": "0.0.0.0:2112",
+    "nodeMetrics": true,
+    "nodeConnMetrics": true,
+    "blockWALMetrics": true,
+    "restAPIMetrics": true,
+    "goMetrics": true,
+    "processMetrics": true,
+    "promhttpMetrics": true
```
ivange94, line 87: This part is a little confusing and hard to remember without referencing the code/documentation. Are all these prometheus metrics? And why can we not just have all of them turned on when the prometheus plugin is turned on?

muXxer: This is because maybe you don't want to scrape all the metrics all the time, for the reason that it might need more CPU/memory/bandwidth to do that, and the responses might be slower.
Its part of the documentation. All the parameters are described there.
This is done in the same way as in hornet and all other INX.

ivange94: So to my understanding, the metrics code only records an event that happens whether or not the prometheus plugin is enabled. For example, the metrics counting blocks created only records the blocks created. Whether the plugin is on or off, these blocks will be created. Not sure registering the metrics, performs any heavy computation to warrant a complex configuration file like this. 

Another concern, with the original implementation, when `metrics.Enabled` is false, a metrics object with empty implementation for all those interfaces will be provided. With this change, how does that work? I mean when `prometheus` is enabled but some metrics say `processMetrics` is turned off, will you need to configure individual default objects for these metrics?

muXxer: > Not sure registering the metrics, performs any heavy computation to warrant a complex configuration file like this.
For `restAPIMetrics`, `promhttpMetrics` `goMetrics`, `processMetrics` we hook into that stuff, so that will create some performance hits.

I don't think this config is so complex, no one ever complained about these configs in hornet, and there are way way more settings there. Most of the times people leave it like it is, or enable more metrics if they need to. We can disable some of them by default later, but I wanted to make them enabled by default for now to see if everything works.

> when prometheus is enabled but some metrics say processMetrics is turned off, will you need to configure individual default objects for these metrics?

no, in the new implementation then these collections are simply not added to the prometheus registry, and then they are not even scraped.
Only for our nodeconn metrics, we provide the non-empty implemention in normal execution, because this is also provided via webapi as far as I have seen. In some test cases we provide the empty implementation, so it should be fine I guess.

muXxer: We can try this behaviour as soon as the gpa branch works correctly to check.

ivange94: > no, in the new implementation then these collections are simply not added to the prometheus registry, and then they are not even scraped.

Oh ok. Fine then.

ivange94 (approved):

PR #1488, opened 2022-12-04
chore: Add new ignore objects and correct EventHandlers

howjmay:
`lib.go` and `lib.ts` are not ignored in gitignore, and some usage of `EventHandlers` are not appropriate naming there


BugFreeSoftware (changes requested):
The IEventHandlers-based structs contain *all* event handlers for a SC. Plural. So no, don't rename it to singular. 
Also the gitignore changes do not take my latest changes to the schema tool into account.
So I would drop this entire PR.

PR #1493, opened 2022-12-05
Fix: webapi

jorgemmsilva:
reqstatus was broken due to gpa changes, fixed now.

also adds an improvement so that request contexts are derived from the webapi plugin context


kape1395 (approved):

packages/webapi/reqstatus/reqstatus.go
```diff
@@ -70,53 +89,8 @@ func (r *reqstatusWebAPI) handleWaitRequestProcessed(c echo.Context) error {
...
-		case <-time.After(req.Timeout):
-			// check again, in case event was triggered just before we subscribed
-			found, ret := tryGetReceipt()
-			if found {
-				return ret
-			}
-			return httperrors.Timeout("Timeout while waiting for request to be processed")
-		}
-	*/
+	rec := <-ch.AwaitRequestProcessed(c.Request().Context(), reqID)
```
muXxer, line 92: Should we maybe add a maximum timeout here?
Otherwise someone could DOS that by create more and more waiting requests in the pipe.

packages/webapi/reqstatus/reqstatus.go
```diff
@@ -70,53 +89,8 @@ func (r *reqstatusWebAPI) handleWaitRequestProcessed(c echo.Context) error {
...
+	return resolveReceipt(c, ch, rec)
```
muXxer, line 93: What exactly happens if the context is canceled?
Will the `rec` be `nil`?
Is that case handled correctly?

packages/webapi/reqstatus/reqstatus.go
jorgemmsilva: good point, I'm thinking we should actually define a max timeout for any request (using the base context) and also set these values in the http server: `ReadTimeout` `ReadHeaderTimeout` `WriteTimeout`

jorgemmsilva: DOS in particular (rate limiting, load balancing, etc) should be handled by the infrastructure, tho

packages/webapi/reqstatus/reqstatus.go
jorgemmsilva: the user should receive a HTTP 408

PR #1502, opened 2022-12-06
fix: reading anchor tx ID from blockinfo

jorgemmsilva:


tools/cluster/tests/rotation_test.go
```diff
@@ -216,7 +216,7 @@ func TestRotationFromSingle(t *testing.T) {
 	case incCounterResult := <-incCounterResultChan:
 		require.NoError(t, incCounterResult)
 	case <-time.After(10 * time.Second):
-		t.FailNow()
+		t.Fatal()
```
dessaya, line 219: Just curious, what's the difference?

dessaya (approved):

muXxer (approved):

muXxer:
Sorry that was my fault :sweat_smile:

tools/cluster/tests/rotation_test.go
jorgemmsilva: I was trying to find where it was failing, from my experience fatal prints the call stack

jorgemmsilva:
the only people that don't produce bugs are the ones that don't write code :smile:

PR #1508, opened 2022-12-06
Adapt nodeConn to gpa consensus changes

### LINKED ISSUE #1390: Update nodeconn to work with the new consensus changes
- Talk to karolis what needs to be changed

muXxer:
This PR adapts the nodeconn to the new gpa consensus changes.

AliasOutputs in the ledger update are now sorted before they are applied to the mempool.
There might still be race conditions in case of L1 milestone changes, but this will be handled in another PR.


packages/chain/node.go
```diff
@@ -437,14 +437,18 @@ func (cni *chainNodeImpl) handleChainMgrOutput(ctx context.Context, outputUntype
...
+				// TODO: why is *iotago.Transaction unused?
 				cni.recvTxPublishedPipe.In() <- &txPublished{
 					committeeAddr:   txToPost.CommitteeAddr,
 					txID:            txToPost.TxID,
 					nextAliasOutput: txToPost.NextAliasOutput,
 					confirmed:       confirmed,
 				}
-			})
+			}); err != nil {
+				// TODO: what should happen if a chain doesn't exist in nodeConn?
```
jorgemmsilva: panic if a chain doesn't exist seems correct, chains must be "registered" in the nodeconn unpon startup.

but is it guaranteed that his error is caused because a chain is missing?

packages/chains/chains.go
```diff
@@ -98,13 +99,20 @@ func New(
...
+
+		if c.ctx != nil {
+			return errors.New("chains already running")
+		}
+		c.ctx = ctx
+
 		return nil
+	}(); err != nil {
+		return err
 	}
```
jorgemmsilva: Thinking it would be more readable to just extract this to a function, rather than in-lining it

packages/nodeconn/nc_chain.go
```diff
@@ -6,85 +6,116 @@ package nodeconn
...
+	"github.com/iotaledger/wasp/packages/chain"
 	"github.com/iotaledger/wasp/packages/isc"
 	"github.com/iotaledger/wasp/packages/parameters"
 )
 
-// nodeconn_chain is responsible for maintaining the information related to a single chain.
+func shouldBeProcessed(out iotago.Output) bool {
+	// only outputs without SDRC should be processed.
+	// TODO: that should be changed in the future, because the SDRC was a
+	// requirement by the wasp team and was added to the protocol because of that.
```
jorgemmsilva: this is fake news. as I recall SDRUC was added because "feeless txs must be possible"

packages/nodeconn/nc_chain.go
```diff
@@ -6,85 +6,116 @@ package nodeconn
...
+	return chain, nil
 }
 
-func (ncc *ncChain) Close() {
-	ncc.nc.DetachMilestones(ncc.milestoneClosure)
+func (ncc *ncChain) queryInititalState() error {
+	ncc.LogInfo("Querying initial state and owned outputs...")
+
+	// TODO: there is a potential race condition if the milestone index
+	// on L1 changes during querying of the initial chain state
```
jorgemmsilva, line 67: we could keep track of the latest AO sent to the callback, and only send the AO if its more recent. would that fix this edge case?

old (already processed) UTXOS should not be problematic to the mempool

packages/nodeconn/nc_chain.go
```diff
@@ -95,125 +126,76 @@ func (ncc *ncChain) PublishTransaction(tx *iotago.Transaction, timeout ...time.D
...
+	var innerErr error
 	for _, query := range queries {
-		if ctxWithTimeout != nil && ctxWithTimeout.Err() == nil {
-			// cancel the ctx of the last query
-			cancelContext()
-		}
-		// TODO what should be an adequate timeout for each of these queries?
-		ctxWithTimeout, cancelContext = newCtxWithTimeout(ncc.nc.ctx)
+		// inline function used to cancel the ctx with defer
+		if err := func() error {
```
jorgemmsilva: again here, wouldn't this be more readable if we just extract it to its own function?

packages/nodeconn/nodeconn.go
```diff
@@ -97,222 +104,241 @@ func New(ctx context.Context, log *logger.Logger, nodeBridge *nodebridge.NodeBri
...
-		nodeBridge:              nodeBridge,
-		nodeClient:              inxNodeClient,
-		pendingTransactionsMap:  make(map[iotago.TransactionID]*PendingTransaction),
+	nc := &nodeConnection{
+		WrappedLogger: logger.NewWrappedLogger(log),
+		ctx:           nil,
+		chainsMap: shrinkingmap.New[iotago.AliasID, *ncChain](
+			shrinkingmap.WithShrinkingThresholdRatio(chainsCleanupThresholdRatio),
+			shrinkingmap.WithShrinkingThresholdCount(chainsCleanupThresholdCount),
+		),
```
jorgemmsilva, line 116: this is over-engineering IMO, a node shouldn't have more than 1 or 2 chains, even if its 100 not garbage collecting this map wouldn't be a problem.

packages/nodeconn/util.go
```diff
@@ -0,0 +1,290 @@
...
+	case iotago.OutputNFT:
+		addressToCheck = output.(*iotago.NFTOutput).Ident()
+
+	default:
+		panic(fmt.Errorf("%w: type %d", iotago.ErrUnknownOutputType, output.Type()))
+	}
+
+	if addressToCheck.Type() != iotago.AddressAlias {
+		// output is not owned by an alias address => ignore it
+		// TODO: what if we have nested ownerships? do we need to take care of that?
```
jorgemmsilva: like chain owns an NFT that owns more assets?
we don't . that would be a never ending rabbit hole 😂

packages/util/generics.go
```diff
@@ -25,3 +25,15 @@ func CloneMap[K comparable, T Cloneable[T]](base map[K]T) map[K]T {
...
+func KeyOnlyBy[K comparable, V any](collection []V, iteratee func(V) K) map[K]struct{} {
+	result := make(map[K]struct{}, len(collection))
+
+	for _, v := range collection {
+		k := iteratee(v)
+		result[k] = struct{}{}
+	}
+
+	return result
+}
```
jorgemmsilva: is this used anywhere?

jorgemmsilva (approved):

packages/nodeconn/nc_chain.go
kape1395: What is SDRUC?

jorgemmsilva: Storage deposit return unlock condition

packages/chains/chains.go
muXxer: Mhh but this function is only used here. Would you mind if I leave it as it is?
Or I remove that weird error message, and just return because the chain is already running (don't see where this should be a problem).

packages/chain/node.go
muXxer: Thats the only reason for an error so far, yes. Should I remove the TODO comment?

packages/nodeconn/nc_chain.go
muXxer: Yeah if the fees that have to be paid to the chain are lower than the storage deposit needed to create the output for the on-ledger-request, what will we do?
Is this planned to be added in the future?
Maybe [@lzpap](https://github.com/lzpap) can explain what was planned here?

packages/nodeconn/nc_chain.go
muXxer: Can we have a small huddle call on this after merging this PR? Who should be involved? You and @kape1395 ?

packages/nodeconn/nc_chain.go
jorgemmsilva: This is the same issue as depositing native tokens with minimum SD, what do we do if the SD is not enough for the fee? Right now the deposit request will fail, but the user gets credited the native token anyway. :/ 
Maybe we need to create the concept of "negative balance" where users must top up their account before being able to use it

packages/chains/chains.go
jorgemmsilva: Feel free to ignore my comments, those are just suggestions and much of programming style is taste-related. We don't have any style guide in IF, so just do as you feel is best 😄

packages/chain/node.go
jorgemmsilva: I'd say so

packages/nodeconn/nc_chain.go
muXxer: Can this own function be inline as well? It won't be used anywhere else IMO.

Example:
```go

func (ncc *ncChain) queryChainUTXOs() error {
	bech32Addr := ncc.chainID.AsAddress().Bech32(parameters.L1().Protocol.Bech32HRP)

	falseCondition := false
	queries := []nodeclient.IndexerQuery{
		&nodeclient.BasicOutputsQuery{AddressBech32: bech32Addr, IndexerStorageDepositParas: nodeclient.IndexerStorageDepositParas{
			HasStorageDepositReturn: &falseCondition,
		}},
		&nodeclient.FoundriesQuery{AliasAddressBech32: bech32Addr},
		&nodeclient.NFTsQuery{AddressBech32: bech32Addr, IndexerStorageDepositParas: nodeclient.IndexerStorageDepositParas{
			HasStorageDepositReturn: &falseCondition,
		}},
		// &nodeclient.AliasesQuery{GovernorBech32: bech32Addr}, // TODO chains can't own alias outputs for now
	}

	processChainUTXOQuery := func(query nodeclient.IndexerQuery) error {
		ctx, cancel := newCtxWithTimeout(ncc.nodeConn.ctx, inxTimeoutIndexerOutputs)
		defer cancel()

		res, err := ncc.nodeConn.indexerClient.Outputs(ctx, query)
		if err != nil {
			return fmt.Errorf("failed to query address outputs: %w", err)
		}

		for res.Next() {
			if res.Error != nil {
				return fmt.Errorf("error iterating indexer results: %w", err)
			}

			outputs, err := res.Outputs()
			if err != nil {
				return fmt.Errorf("failed to fetch address outputs: %w", err)
			}

			outputIDs, err := res.Response.Items.OutputIDs()
			if err != nil {
				return fmt.Errorf("failed to get outputIDs from response items: %w", err)
			}

			for i := range outputs {
				outputID := outputIDs[i]
				ncc.LogDebugf("received UTXO, outputID: %s", outputID.ToHex())
				ncc.requestOutputHandler(isc.NewOutputInfo(outputID, outputs[i], iotago.TransactionID{}))
			}
		}

		return nil
	}

	for _, query := range queries {
		if err := processChainUTXOQuery(query); err != nil {
			return err
		}
	}

	return nil
}

```

packages/nodeconn/nodeconn.go
muXxer: But should we remove this then? In the end its a long living map and a potential memory leak.
By adding these few lines, at least we can exclude it, if we ever hunt for memory leaks?

packages/nodeconn/util.go
muXxer: Should I change the comment to: "Nested ownership is ignored" ?

packages/util/generics.go
muXxer: Ah, I implemented that in an intermediate solution I had first.
Should I remove it? I think it could potentially be useful somewhere else, but I could also add it to hive.go instead.

packages/nodeconn/util.go
jorgemmsilva: Sure 👌

packages/nodeconn/nc_chain.go
jorgemmsilva: I think that helps. Don't see any issue as being a function that is only used in one place, but this is an improvement IMO

packages/nodeconn/nodeconn.go
jorgemmsilva: I feel like the standard stuff should be used, unless it's clear that it has some downside, but feel free to ignore this comment 😛

packages/nodeconn/nc_chain.go
muXxer: Lets create an issue in the project board then.

packages/nodeconn/nodeconn.go
dessaya: I agree with Jorge here: let's fix stuff when it is clear that it is a problem, not before.

packages/util/generics.go
dessaya: It looks like useful only in very specific cases, so I don't think it is valuable to keep if it's unused.

packages/nodeconn/nc_chain.go
jorgemmsilva: https://github.com/iotaledger/wasp/issues/1514

PR #1522, opened 2022-12-07
Remove evm artifacts from the repository

ivange94:
# Description of change

Remove evm artifacts from the repository.


.github/workflows/build-test.yml
```diff
@@ -18,6 +18,15 @@ jobs:
...
         uses: actions/checkout@v2
 
+      - name: install Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
+
+      - name: Install solc
+        run: npm install -g solc
```
jorgemmsilva: maybe just this will suffice and we don't need npm
```
sudo apt-get install solc
```

jorgemmsilva (approved):

dessaya (approved):

muXxer (changes requested):
`//go:embed` fails if we do this.
so the linter can't run properly. can we somehow tell the linter to compile the files before running?

Maybe we need to add the compile step to the lint action in `build-test.yml` as well?

muXxer (approved):

PR #1528, opened 2022-12-09
Chain redesign: Integration tests.

kape1395:
`./packages/chain/...` tests are passing now.


packages/chain/node.go
```diff
@@ -14,6 +14,8 @@
 // This object interacts with:
 //   - NodeConn.
 //   - Administrative functions.
+//
+// TODO: Ensure access nodes contain the committee nodes.
```
muXxer, line 18: you did that already in this PR, no?

muXxer (approved):

packages/chain/node.go
kape1395: Not yet. Will resolve along with handling the per-node access nodes.

PR #1535, opened 2022-12-12
Make Publisher work again.

kape1395:
The publisher component was introduced. It acts as a listener for a chain.
That's an initial implementation. WS instead of nano should be used, but the main construction works.


packages/publisher/publisher.go
```diff
@@ -11,6 +23,101 @@ var Event = events.NewEvent(func(handler interface{}, params ...interface{}) {
...
+		running:          &atomic.Bool{},
+		log:              log,
+	}
+	return p
+}
+
+// Implements the chain.ChainListener interface.
+// NOTE: Do not block the caller!
+func (p *Publisher) BlockApplied(chainID isc.ChainID, block state.Block) {
+	p.blockAppliedPipe.In() <- &publisherBlockApplied{chainID: chainID, block: block}
```
jorgemmsilva, line 56: why use pipe instead of chan?

packages/publisher/publisher.go
```diff
@@ -11,6 +23,101 @@ var Event = events.NewEvent(func(handler interface{}, params ...interface{}) {
...
+	ctxDone := ctx.Done()
+	for {
+		select {
+		case blockAppliedUntyped := <-blockAppliedPipeOutCh:
+			p.handleBlockApplied(blockAppliedUntyped.(*publisherBlockApplied))
+		case <-ctxDone:
+			return
+		}
+	}
+}
```
jorgemmsilva: minor suggestion (to remove `ctxDone := ctx.Done()`)

```suggestion
func (p *Publisher) Run(ctx context.Context) {
	blockAppliedPipeOutCh := p.blockAppliedPipe.Out()
	for {
		select {
		case blockAppliedUntyped := <-blockAppliedPipeOutCh:
			p.handleBlockApplied(blockAppliedUntyped.(*publisherBlockApplied))
		case <-ctx.Done():
			return
		}
	}
}
```

tools/cluster/tests/deploy_test.go
```diff
@@ -19,8 +19,8 @@ func TestDeployChain(t *testing.T) {
 
 	counter1, err := e.Clu.StartMessageCounter(map[string]int{
 		"dismissed_chain": 0,
-		"state":           2,
-		"request_out":     1,
+		"state":           0, // TODO: Was 2, is it meaningful to check this?
```
jorgemmsilva, line 22: >is it meaningful to check this?
I don't think so.

IMO these `MessageCounter` tests are a bit odd and should be removed

jorgemmsilva (approved):

packages/publisher/publisher.go
kape1395: To avoid locking. The pipe never blocks. Channel will block if its capacity is exhausted.
Here we are interacting with outside world, some bad nano/ws client could read from the socket slowly making the chain to block.

packages/publisher/publisher.go
kape1395: Isn't then the function Done() is executed on each loop iteration?

packages/publisher/publisher.go
```diff
@@ -11,6 +23,101 @@ var Event = events.NewEvent(func(handler interface{}, params ...interface{}) {
...
+
+// This is called by the component to run this.
+func (p *Publisher) Run(ctx context.Context) {
+	blockAppliedPipeOutCh := p.blockAppliedPipe.Out()
+	ctxDone := ctx.Done()
+	for {
+		select {
+		case blockAppliedUntyped := <-blockAppliedPipeOutCh:
+			p.handleBlockApplied(blockAppliedUntyped.(*publisherBlockApplied))
```
muXxer, line 69: wouldn't that panic if the channel gets closed? or does this never happen?

packages/publisher/publisher.go
jorgemmsilva: guess so, but that's probably negligible.

this is how the go docs do it:
https://pkg.go.dev/context#Context

packages/publisher/publisher.go
```diff
@@ -11,6 +23,101 @@ var Event = events.NewEvent(func(handler interface{}, params ...interface{}) {
...
+	chainID isc.ChainID
+	block   state.Block
+}
+
+func NewPublisher(log *logger.Logger) *Publisher {
+	p := &Publisher{
+		blockAppliedPipe: pipe.NewDefaultInfinitePipe(),
+		mutex:            &sync.RWMutex{},
+		running:          &atomic.Bool{},
+		log:              log,
```
muXxer, line 48: If you use the `WrappedLogger` like we do in some places in the code, the code wouldn't panic if `log == nil`.

packages/publisher/publisher.go
jorgemmsilva: we should read from the channel always, then send to any subscribed client, if they don't read it - that should never block the publisher (if it blocks, that's poor implementation of WS/NanoMsg/whatever lib we use).

I always prefer standard chan to pipe, unless there is a good reason to use them

packages/publisher/publisher.go
kape1395: Ok, ok.

packages/publisher/publisher.go
kape1395: I added a  check.

packages/publisher/publisher.go
kape1395: Should we expect the nil logger? We expect it to be non-nil in the codebase.

muXxer: You can leave it as it is, just wanted to mention that we also use the other way in some place. 
Maybe I can clean that up some day, after the node runs.

tools/cluster/tests/deploy_test.go
kape1395: Let's solve that in other PR.

packages/publisher/publisher.go
kape1395: I think that's exactly the reason to use a pipe instead of a plain channel. I prefer to use channels as well. But they are for synchronization and not for queueing.

PR #1538, opened 2022-12-12
Access nodes to be used

kape1395:


packages/chain/node.go
```diff
@@ -222,11 +236,14 @@ func New(
...
 		publishingTXes:         map[iotago.TransactionID]context.CancelFunc{},
 		procCache:              processors.MustNew(processorConfig),
+		configUpdatedCh:        make(chan *configUpdate, 1),
 		listener:               listener,
 		accessLock:             &sync.RWMutex{},
 		activeCommitteeDKShare: nil,
 		activeCommitteeNodes:   []*cryptolib.PublicKey{},
-		activeAccessNodes:      []*cryptolib.PublicKey{},
+		activeAccessNodes:      nil,                // updated bellow.
+		accessNodesFromNode:    accessNodesForNode, // TODO: Update it on event.
```
jorgemmsilva:
```suggestion
		accessNodesFromNode:    accessNodesForNode,
```
we can probably remove the TODO

jorgemmsilva (approved):

packages/chains/chains.go
```diff
@@ -110,6 +111,14 @@ func (c *Chains) Run(ctx context.Context) error {
 	}
 	c.ctx = ctx
 
+	c.chainRecordRegistryProvider.Events().ChainRecordModified.Attach(event.NewClosure(func(event *registry.ChainRecordModifiedEvent) {
+		c.mutex.Lock()
```
muXxer: Wouldn't a RLock be sufficient?
And you can move the defer directly after the RLock().

muXxer (changes requested):

packages/chain/node.go
kape1395: Thanks!

packages/chains/chains.go
kape1395: Good point, thanks.

muXxer (approved):

PR #1539, opened 2022-12-12
Fix linter errors

muXxer:


packages/testutil/testchain/mock_nodeconn.go
```diff
@@ -12,17 +12,19 @@ import (
 	"github.com/iotaledger/wasp/packages/metrics/nodeconnmetrics"
 )
 
+// TODO: fix MockedNodeConn (do we need the commented code?)
```
jorgemmsilva: Probably not, let's just remove it

jorgemmsilva (approved):
👍

packages/testutil/testchain/mock_nodeconn.go
muXxer: Removed! :+1:

dessaya (approved):

PR #1541, opened 2022-12-13
Fix TestNodeConn cluster test

muXxer:
The callbacks for AttachChain were in the wrong order, so the test was blocking in the channels.

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -17,6 +17,9 @@ fi
 FILES=$(go list ./... | grep -v github.com/iotaledger/wasp/contracts/wasm)
 ${GO_EXECUTABLE} clean -testcache
 
+make wasm
+make install
+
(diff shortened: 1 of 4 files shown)


scripts/run_tests.sh
```diff
@@ -17,6 +17,10 @@ fi
 FILES=$(go list ./... | grep -v github.com/iotaledger/wasp/contracts/wasm)
 ${GO_EXECUTABLE} clean -testcache
 
+make compile-solidity
+make wasm
+make install
```
jorgemmsilva:
```suggestion
make wasm
make install
```

I think make install already does the solidity bit

jorgemmsilva (approved):

scripts/run_tests.sh
muXxer: Will fix :+1:

kape1395 (approved):

PR #1542, opened 2022-12-13
Statemanager

Juliusan:
Test enhancement and some other minor fixes for state manager


packages/chain/statemanager/state_manager_test.go
```diff
@@ -19,18 +19,26 @@ import (
 	"github.com/iotaledger/wasp/packages/testutil/testpeers"
 )
 
-func TestCruelWorld(t *testing.T) {
+func TestCruelWorld(t *testing.T) { //nolint: gocyclo
```
jorgemmsilva: could we break this up into smaller pieces, instead of ignoring the linter altogether?

packages/chain/statemanager/state_manager_test.go
```diff
@@ -150,3 +222,32 @@ func requireTrueForSomeTime(t *testing.T, ch <-chan bool, timeout time.Duration)
...
+
+func waitForTrue(predicate func() bool, iterationWait time.Duration, maxIterations int) bool {
+	for i := 0; i < maxIterations; i++ {
+		if predicate() {
+			return true
+		}
+		time.Sleep(iterationWait)
+	}
+	return false
+}
```
jorgemmsilva: seems very similar to https://github.com/iotaledger/wasp/blob/develop/tools/cluster/tests/util.go#L211-L221
maybe both should be unified and re-used?

jorgemmsilva (approved):

packages/chain/statemanager/state_manager_test.go
Juliusan: I can move all the function declarations out of the test, but I am no sure if that is going to make the test easier to understand.

packages/chain/statemanager/state_manager_test.go
Juliusan: Yes, they are similar. Apparently `tools/cluster/tests/` is not a right place for this function.

jorgemmsilva: yes, should be on a more generic place, maybe `testutils`

packages/chain/statemanager/state_manager_test.go
BugFreeSoftware: You may want to rebase on develop to pass the tests. Diego had the same problem and that fixed it.

Juliusan: Removed this nolint and handled the warning.

packages/chain/statemanager/state_manager_test.go
Juliusan: The function is not needed any more and is gone from the code.

kape1395 (approved):

PR #1550, opened 2022-12-14
tests: fix cluster tests with permissioned access nodes

jorgemmsilva:
- fixes putRecord webapi not using accessnodes
- removes "pubkey to hex" "from hex", just use "string/fromString" 
- fixes cluster tests with permissioned access nodes (there is a small snippet of code that can be removed if we automate adding the correct peers to chain records, marked with a "TODO remove")
- fixes log levels being forcefully changed in some places
- updates cluster tests default ports so that they don't conflicts with commonly used ports (specially when running tests with 20 or 30 nodes)

### DIFF
--- a/packages/registry/chain_registry.go
+++ b/packages/registry/chain_registry.go
@@ -281,8 +281,6 @@ func (p *ChainRecordRegistryImpl) UpdateChainRecord(chainID isc.ChainID, callbac
 	}
 
 	if modified {
-		s, _ := chainRecord.MarshalJSON()
-		println("AAA ", string(s))
 		p.events.ChainRecordModified.Trigger(&ChainRecordModifiedEvent{
 			ChainRecord: chainRecord,
 		})
--- a/packages/registry/chain_registry.go
+++ b/packages/registry/chain_registry.go
@@ -27,11 +27,11 @@ type ChainRecord struct {
 	AccessNodes []*cryptolib.PublicKey
 }
 
-func NewChainRecord(chainID isc.ChainID, active bool) *ChainRecord {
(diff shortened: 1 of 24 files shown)


packages/registry/chain_registry.go
```diff
@@ -281,6 +281,8 @@ func (p *ChainRecordRegistryImpl) UpdateChainRecord(chainID isc.ChainID, callbac
 	}
 
 	if modified {
+		s, _ := chainRecord.MarshalJSON()
```
muXxer: was this debug stuff? :D

packages/webapi/model/chain_record.go
```diff
@@ -3,21 +3,40 @@
...
 }
 
-func (bd *ChainRecord) Record() *registry.ChainRecord {
-	return registry.NewChainRecord(bd.ChainID.ChainID(), bd.Active)
+func (bd *ChainRecord) Record() (*registry.ChainRecord, error) {
+	accessNodes := make([]*cryptolib.PublicKey, len(bd.AccessNodes))
+
+	for i, pubKeyStr := range bd.AccessNodes {
+		pubKey, err := cryptolib.NewPublicKeyFromString(pubKeyStr)
+		accessNodes[i] = pubKey
```
muXxer: maybe check the error before setting the pubKey in the slice

dessaya (approved):

packages/registry/chain_registry.go
jorgemmsilva: well spotted! thanks

muXxer (approved):

PR #1583, opened 2022-12-19
refactor: publisher get events

jorgemmsilva:
refactors the way the publisher package gets events, in preparation for the implementation of "subscription filters".

fixes the cluster publisher test.

There are some TODOs in that test that point to a race condition in the "wait for request" endpoint. To be looked into.

### DIFF
--- a/packages/publisher/events.go
+++ b/packages/publisher/events.go
@@ -1,0 +1,100 @@
+package publisher
+
+import (
+	"fmt"
+
+	"github.com/iotaledger/wasp/packages/isc"
+	"github.com/iotaledger/wasp/packages/kv"
+	"github.com/iotaledger/wasp/packages/kv/subrealm"
+	"github.com/iotaledger/wasp/packages/state"
+	"github.com/iotaledger/wasp/packages/vm/core/blocklog"
+)
+
+const (
+	ISCEventKindNewBlock      = "new_block"
+	ISCEventKindReceipt       = "receipt" // issuer will be the request sender
+	ISCEventKindSmartContract = "contract"
+)
+
+const ISCEventIssuerVM = "vm"
+
+type ISCEvent struct {
+	Kind      string
+	Issuer    isc.AgentID // nil means issued by the VM
+	RequestID isc.RequestID
+	Content   string
+}
+
+// kind is not printed right now, because it is added when calling p.publish
+func (e *ISCEvent) String() string {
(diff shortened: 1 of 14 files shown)


packages/publisher/events.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return fmt.Sprintf("%s - %s", issuerStr, e.Content)
+}
+
+// EventsFromBlock extracts the events from a block, its returns a chan of ISCEvents so they can be filtered
+func EventsFromBlock(block state.Block) (chan ISCEvent, chan error) {
+	resultCh := make(chan ISCEvent)
+	errCh := make(chan error)
+
+	go func() {
```
kape1395: Why can't we provide a predicate for filtering instead of spawning routines here for filtering?

dessaya: +1, why does this function return channels instead of slices?

packages/webapi/reqstatus/reqstatus.go
```diff
@@ -78,6 +78,34 @@ func (r *reqstatusWebAPI) handleWaitRequestProcessed(c echo.Context) error {
 	}
 	rec := <-ch.AwaitRequestProcessed(c.Request().Context(), reqID, true)
 	return r.resolveReceipt(c, ch, rec)
+
+	// retCh := make(chan *blocklog.RequestReceipt, 1)
```
dessaya: This left intentionally commented out?

packages/publisher/events.go
jorgemmsilva: The idea is that it will be very easy to consume the channel, filter the events, and forward them to all the clients that are subscribed to certain criteria.

If this returns a slice, we need to hold it in memory and iterate it again to filter/publish.

We could also do the publishing as we get the events, without the channel, but this seems to me like an easy abstraction, and keeps the logic separated. We can get rid of it, sure

dessaya: The main concern with this approach is that it may cause a goroutine leak if the caller does not fully consume the chans.

dessaya (approved):

packages/publisher/events.go
jorgemmsilva: Fair points. I'll reconsider this tomorrow :+1:

jorgemmsilva: removed the go routine :ok_man:

PR #1589, opened 2022-12-20
Publish a off-ledger-request to access nodes, if committee is not known.

kape1395:


jorgemmsilva (approved):

packages/chain/mempool/distSync/distSync.go
```diff
@@ -122,7 +123,13 @@ func (dsi *distSyncImpl) handleInputAccessNodes(input *inputAccessNodes) gpa.Out
 //   - Just send a message to all the committee nodes.
```
muXxer, line 123: Shouldn't we change that comment as well?

kape1395: OK, will change. With another PR I'm working on.

PR #1595, opened 2022-12-21
Fix: webapi chainrecords

jorgemmsilva:


packages/webapi/model/address.go
```diff
@@ -19,6 +19,9 @@ func (a Address) MarshalJSON() ([]byte, error) {
 }
 
 func (a *Address) UnmarshalJSON(b []byte) error {
+	if len(b) <= 2 { // empty string means len(b) == 2 // ("")
+		return nil
+	}
```
dessaya: Why is an empty string considered a valid address? :thinking:

jorgemmsilva: Our API is sending empty strings when the fields don't exist, I think it's a consequence of changing stuff from pointers to objects (strings)

dessaya: And can't we fix the API instead? Otherwise IMO this seems like inviting hard-to-find bugs. For example:

```go
var a Address
err := a.UnmarshalJSON([]byte(""))
require.NoError(err)
// since no error I expect a to be valid
addr := a.Address() // this panics
```

jorgemmsilva: yup, I realized this was the correct way to go after writing the previous comment. Thanks for the repair! :ok_hand:

dessaya (approved):

muXxer (approved):

PR #1602, opened 2022-12-22
misc fixes

dessaya:
After DAYS of trying to find out why the nodes were getting stuck... turns out the problem was in `wasp-cluster` all along :man_facepalming: . When a wasp node printed a very long line the `bufio.Scanner` in `wasp-cluster` errored and everything went to a halt.

Good news is that in the process I ended up fixing two other unrelated small issues.

* fix(wasp-cluster): properly handle long lines in wasp log
* fix(vm): enforce maximum error size in receipt
* fix: properly handle timeout in api/waitRequestProcessed

### DIFF
--- a/packages/webapi/reqstatus/reqstatus.go
+++ b/packages/webapi/reqstatus/reqstatus.go
@@ -76,10 +76,16 @@ func (r *reqstatusWebAPI) handleWaitRequestProcessed(c echo.Context) error {
 			return httperrors.BadRequest("Invalid request body")
 		}
 	}
+
+	delay := time.NewTimer(req.Timeout)
 	select {
 	case rec := <-ch.AwaitRequestProcessed(c.Request().Context(), reqID, true):
+		if !delay.Stop() {
+			// empty the channel to avoid leak
+			<-delay.C
+		}
 		return r.resolveReceipt(c, ch, rec)
-	case <-time.After(req.Timeout):
+	case <-delay.C:
 		return httperrors.Timeout("Timeout")
 	}
 }
--- a/packages/webapi/reqstatus/reqstatus.go
+++ b/packages/webapi/reqstatus/reqstatus.go
@@ -76,8 +76,12 @@ func (r *reqstatusWebAPI) handleWaitRequestProcessed(c echo.Context) error {
 			return httperrors.BadRequest("Invalid request body")
 		}
 	}
-	rec := <-ch.AwaitRequestProcessed(c.Request().Context(), reqID, true)
-	return r.resolveReceipt(c, ch, rec)
+	select {
+	case rec := <-ch.AwaitRequestProcessed(c.Request().Context(), reqID, true):
+		return r.resolveReceipt(c, ch, rec)
+	case <-time.After(req.Timeout):
+		return httperrors.Timeout("Timeout")
+	}
 }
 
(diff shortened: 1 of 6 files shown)


packages/webapi/reqstatus/reqstatus.go
```diff
@@ -76,8 +76,12 @@ func (r *reqstatusWebAPI) handleWaitRequestProcessed(c echo.Context) error {
 			return httperrors.BadRequest("Invalid request body")
 		}
 	}
-	rec := <-ch.AwaitRequestProcessed(c.Request().Context(), reqID, true)
-	return r.resolveReceipt(c, ch, rec)
+	select {
+	case rec := <-ch.AwaitRequestProcessed(c.Request().Context(), reqID, true):
+		return r.resolveReceipt(c, ch, rec)
+	case <-time.After(req.Timeout):
```
jorgemmsilva: this can create leaks, as the `ticker` from `time.After` is only GC'd after the timeout expires :confused: 

btw, I think we could remove this "user timeout" and just rely on the "generic timeout" middleware, that will drop any long running request

tools/cluster/cluster.go
```diff
@@ -645,13 +630,31 @@ func waitForAPIReady(initOk chan<- bool, apiURL string) {
...
+			if err != io.EOF {
+				fmt.Printf("[cluster] error reading output for %s: %v\n", tag, err)
+			}
+			break
+		}
+
+		if isLineStart {
+			fmt.Printf("[%s] ", tag)
 		}
+		fmt.Printf("%s", line)
```
jorgemmsilva, line 652:
```suggestion
		fmt.Print(line)
```

jorgemmsilva (approved):

packages/webapi/reqstatus/reqstatus.go
dessaya: I'm not sure about removing the user-defined timeout... in this case I think it is part of the endpoint interface and it is useful for the client to be able to customize it. e.g.: "I want to wait until my request is processed, but not more than x seconds". Anyway, if we decide to remove it, I think we should do that in a separate PR since it will be a larger change (since it involves changing the client as well).

Also, about `time.After`... how severe is that ticker not being gc'ed if the timeout is smallish? Maybe we should limit the timeout value, like 2 minutes max or so?

tools/cluster/cluster.go
dessaya: Tried that at first, but `line` is `[]byte` here, so it doesn't work :)

packages/webapi/reqstatus/reqstatus.go
jorgemmsilva: If we get many "waits per second" it can snowball. Probably not too bad, but something I keep in mind after learning that "select time.after" leaks lol.

I don't think user timeout is much needed, I mean the client can just cancel the request if they want to.

tools/cluster/cluster.go
jorgemmsilva: ahah 👍

packages/webapi/reqstatus/reqstatus.go
dessaya: Ok, for now I replaced time.After with NewTimer. We can remove the timeout in another PR.

PR #1617, opened 2023-01-04
ServerNodes = Inverse(AccessNodes)

kape1395:
A chain has to use ServerNodes to query requests of blocks instead of AccessNodes.
The ServerNodes is the inverse of AccessNodes, considering the latter as a relation.
Now the access nodes get connected to the chain automatically.


packages/webapi/testutil/mockchain.go
```diff
@@ -74,6 +74,11 @@ func (*MockChain) ConfigUpdated(accessNodes []*cryptolib.PublicKey) {
 	panic("unimplemented")
 }
 
+// ConfigUpdated implements chain.Chain
```
muXxer:
```suggestion
// ServersUpdated implements chain.Chain
```

packages/solo/chain.go
```diff
@@ -625,6 +625,11 @@ func (*Chain) ConfigUpdated(accessNodes []*cryptolib.PublicKey) {
 	panic("unimplemented")
 }
 
+// ConfigUpdated implements chain.Chain
```
muXxer:
```suggestion
// ServersUpdated implements chain.Chain
```

packages/peering/lpp/lppNetImpl.go
```diff
@@ -472,6 +472,11 @@ func (n *netImpl) TrustedPeers() ([]*peering.TrustedPeer, error) {
 	return n.trusted.TrustedPeers()
 }
 
+// TrustedPeers implements the peering.TrustedNetworkManager interface.
```
muXxer:
```suggestion
// TrustedPeersListener implements the peering.TrustedNetworkManager interface.
```

packages/chains/accessMgr/amDist/msg.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/iotaledger/wasp/packages/gpa"
+)
+
+const (
+	msgTypeAccess byte = iota
+)
+
+func (ami *accessMgrDist) UnmarshalMessage(data []byte) (gpa.Message, error) {
+	if len(data) < 1 {
+		return nil, fmt.Errorf("accessMgrImpl::UnmarshalMessage: data to short")
```
muXxer:
```suggestion
		return nil, fmt.Errorf("accessMgrImpl::UnmarshalMessage: data too short")
```

packages/chains/accessMgr/amDist/msg.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+		return m, nil
+	}
+	var logData []byte
+	if len(data) <= 20 {
+		logData = data
+	} else {
+		logData = data[0:20]
+	}
+	return nil, fmt.Errorf("accessMgrImpl::UnmarshalMessage: cannot parse message starting with: %v", logData)
```
muXxer: IMO its better to encode the logData to hex.

packages/gpa/ack_handler.go
```diff
@@ -47,6 +48,12 @@ func NewAckHandler(me NodeID, nested GPA, resendPeriod time.Duration) AckHandler
 	}
 }
 
+func (a *ackHandler) DismissPeer(peerID NodeID) {
+	delete(a.counters, peerID)
```
muXxer, line 52: We should use the generic `ShrinkingMap` from hive.go for all long living maps that get modified over time, otherwise this will leak memory.

packages/chains/accessMgr/accessMgr_test.go
```diff
@@ -0,0 +1,108 @@
...
+		am.ChainAccessNodes(chainID, peerPubKeys)
+	}
+	//
+	// Wait for everyone to get the server nodes.
+	for {
+		for i := range nodeServers {
+			if !util.Same(nodeServers[i], peerPubKeys) {
+				t.Logf("Wait for node %v", i)
+				time.Sleep(100 * time.Millisecond)
+				continue
```
muXxer: I don't get that logic of these two for loops.
Should this line continue the outer for loop instead? If so, you need to label it.

muXxer: How does this test even fail? After a certain timeout?

packages/chains/accessMgr/accessMgr.go
```diff
@@ -0,0 +1,234 @@
...
+	TrustedNodes(trusted []*cryptolib.PublicKey)
+	ChainAccessNodes(chainID isc.ChainID, accessNodes []*cryptolib.PublicKey)
+	ChainDismissed(chainID isc.ChainID)
+}
+
+type accessMgrImpl struct {
+	dist                gpa.AckHandler
+	dismissPeerBuf      []*cryptolib.PublicKey
+	reqTrustedNodesPipe pipe.Pipe
+	reqChainAccessNodes pipe.Pipe
```
muXxer: why not also call these variables `Pipe` like `reqTrustedNodesPipe` ?

packages/chains/accessMgr/accessMgr.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+type reqChainDismissed struct {
+	chainID isc.ChainID
+}
+
+var _ AccessMgr = &accessMgrImpl{}
+
+const (
+	msgTypeAccessMgr byte = iota
```
muXxer, line 54: why is `iota` used here if it is only 1 constant anyway? I would set it directly to the specific number instead.

packages/chains/accessMgr/accessMgr.go
```diff
@@ -0,0 +1,234 @@
...
+)
+
+func New(
+	ctx context.Context,
+	serversUpdatedCB func(chainID isc.ChainID, servers []*cryptolib.PublicKey),
+	nodeIdentity *cryptolib.KeyPair,
+	net peering.NetworkProvider,
+	log *logger.Logger,
+) AccessMgr {
+	netPeeringID := peering.PeeringIDFromBytes([]byte("AccessMgr"))
```
muXxer: Is this correct here? Should that be the same value for all nodes?

BTW: I saw that we sometimes hash the chainID plus some additional bytes, and sometimes not.
Maybe this is also wrong? See `mempool.go` and `node.go`.

packages/chains/accessMgr/accessMgr.go
```diff
@@ -0,0 +1,234 @@
...
+
+// Implements the AccessMgr interface.
+func (ami *accessMgrImpl) ChainDismissed(chainID isc.ChainID) {
+	ami.reqChainDismissed.In() <- &reqChainDismissed{chainID: chainID}
+}
+
+// A callback for amDist.
+func (ami *accessMgrImpl) dismissPeerCB(peerPubKey *cryptolib.PublicKey) {
+	// Dismiss them after the messages are sent.
+	ami.dismissPeerBuf = append(ami.dismissPeerBuf, peerPubKey)
```
muXxer, line 120: Is `dismissPeerBuf` threadsafe? Shouldn't we at least add some mutex?

packages/chains/accessMgr/accessMgr.go
```diff
@@ -0,0 +1,234 @@
...
+			MsgReceiver: peering.PeerMessageReceiverAccessMgr,
+			MsgType:     msgTypeAccessMgr,
+			MsgData:     msgData,
+		}
+		ami.net.SendMsgByPubKey(ami.netPeerPubs[m.Recipient()], pm)
+	})
+}
+
+func (ami *accessMgrImpl) pubKeyAsNodeID(pubKey *cryptolib.PublicKey) gpa.NodeID {
+	nodeID := gpa.NodeID(pubKey.String())
```
muXxer: Why do we use a string for the `NodeID` type btw? Why don't we use the `PublicKeyKey` type instead? :)
That should use less memory and should be faster to compare, in map lookups etc.

muXxer: grafik

packages/chains/accessMgr/amDist/accessMgrDist.go
```diff
@@ -0,0 +1,403 @@
...
+
+type Output interface {
+	ChainServerNodes(chainID isc.ChainID) []*cryptolib.PublicKey
+}
+
+type accessMgrDist struct {
+	nodes            map[gpa.NodeID]*accessMgrNode             // State for each peer.
+	chains           map[isc.ChainID]*accessMgrChain           // State for each chain.
+	pubKeyToNodeID   func(*cryptolib.PublicKey) gpa.NodeID     // Convert PubKeys to NodeIDs.
+	serversUpdatedCB func(isc.ChainID, []*cryptolib.PublicKey) // Called when a set fo servers has changed for a chain.
```
muXxer:
```suggestion
	serversUpdatedCB func(isc.ChainID, []*cryptolib.PublicKey) // Called when a set of servers has changed for a chain.
```

packages/chains/accessMgr/amDist/accessMgrDist.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// Access node list has updated for a particular chain.
+//
+// > Send disabled for nodes not in the access list anymore.
+// > Send enabled for new access nodes.
+func (amd *accessMgrDist) handleInputAccessNodes(input *inputAccessNodes) gpa.OutMessages {
+	//
+	// Update the info from the chain perspective.
+	chain, ok := amd.chains[input.chainID]
```
muXxer, line 126: Can't all these functions be executed in parallel by different threads?
If so, the amd.chains access needs to be locked with mutexes.

packages/chains/accessMgr/amDist/accessMgrDist.go
```diff
@@ -0,0 +1,403 @@
...
+
+func (amn *accessMgrNode) handleMsgAccess(msg *msgAccess) gpa.OutMessages {
+	// This has to be checked before updating the state.
+	// > IF /\ m.access = serverForChains(n, m.src)    \* Peer's info hasn't changed, so we don't need to ack it.
+	// >    /\ m.server = H(accessForChains(n, m.src)) \* Our info echoed, so that was an ack.
+	// >    /\ m.src_lc >= lClock[n][m.src]            \* Peer's clock is not outdated, we don't need to push it forward.
+	// >    /\ m.dst_lc <= lClock[n][n]                \* And the echoed clock don't exceed our clock, so we don't need to push it.
+	// > THEN sendAndAck(m, {})
+	// > ELSE sendAndAck(m, accessMsgs(n))
+	sendDone := true &&
```
muXxer, line 328:
```
	sendDone := true &&
```
the true is not needed.

packages/chains/accessMgr/amDist/accessMgrDist.go
```diff
@@ -0,0 +1,403 @@
...
+		msgServerFor.FromSlice(msg.serverForChains)
+		if !amn.accessFor.Equals(msgServerFor) {
+			amn.ourLC++
+		}
+	}
+	//
+	// Send message back, if needed.
+	if !sendDone {
+		return gpa.NoMessages().Add(
+			newMsgAccess(msg.Sender(), amn.ourLC, amn.peerLC, amn.accessFor.AsSlice(), amn.serverFor.AsSlice()),
```
muXxer, line 353: is this also no problem that we access all these slices without mutexes?

muXxer (changes requested):

packages/chains/accessMgr/accessMgr.go
kape1395: Thanks for pointing this out. Will fix.

packages/chains/accessMgr/accessMgr.go
kape1395: All the distributed / GPA algorithms are now designed in the same way. I kept the constant assignment like this to have the implementation uniform with other algorithms. I don't know if setting it to 0 would make less of a surprise in the future.

packages/chains/accessMgr/accessMgr.go
kape1395: Yes. We have 1 AccessMgr per Wasp node. It manages access for all the chains. Thus the identifier is a constant. Mempools are per chain, chain nodes - as well. And the CmtLog instances are per chain/committee, etc. That's why the identifiers are constructed in different ways.

packages/chains/accessMgr/accessMgr.go
kape1395: This attribute is only accessed from a single thread, i.e., from the run() function. I was trying to control the goroutines with the GPA approach, primarily by running a component on a dedicated routine. Other components then communicate with the component's routine via channels and pipes. The code is organized this way to avoid using mutexes too much.

packages/chains/accessMgr/accessMgr.go
kape1395: The initial reason was to handle a hierarchy of algorithms. I.e. consensus uses RBC, ACS, and others. The idea was to have different node identifiers for each protocol in the hierarchy. Later the hierarchy was resolved differently.
Also, I wanted to keep the GPA framework independent of other parts in the wasp code base. So that's less of an argument.
Now a lot of tests are using string-based node ids. Mainly to have them short, which is useful when analyzing logs.

packages/chains/accessMgr/accessMgr_test.go
kape1395: The intention was to have a failure on a timeout. That's because the tested function works asynchronously; thus, there is no clear way in the test to determine when the result is reached. I do that here by repeatedly checking the conditions with delays until the condition passes or a timeout occurs.

The loop is strange. Rewrote it.

packages/chains/accessMgr/amDist/accessMgrDist.go
kape1395: No, this object is always used from a single thread. That's designed intentionally to avoid complicating the code by mixing the distributed algorithms with threading.

packages/chains/accessMgr/amDist/accessMgrDist.go
kape1395: It is here only to have the conditions below aligned and more readable. I hope that `true &&` would be optimized out by the compiler anyway.

packages/chains/accessMgr/amDist/accessMgrDist.go
kape1395: No, not a problem, as mentioned above --  the object is accessed from a single thread.

packages/chains/accessMgr/amDist/msg.go
kape1395: Thanks, updated is all the similar places.

packages/gpa/ack_handler.go
kape1395: Probably yes. But maybe we should do that in a separate PR and for all the uses of maps? There will be a lot of them. Also, it seems like more functions would be needed for the `ShrinkingMap` to keep the code readable, e.g. `MustGet`.

kape1395 (commented):
Responses to comments.

packages/gpa/ack_handler.go
muXxer: I created an issue #1627

packages/chains/accessMgr/accessMgr_test.go
muXxer: But still, shouldn't we have a defined timeout? Like a context with Timeout with 1 minute.
Otherwise it will run as long as the timeout parameter for the tests is set (I put there 1h because we have so many tests).

packages/chains/accessMgr/accessMgr.go
muXxer: Can you add a new `const` section if you use the `iota` keyword?

```

const (
	msgTypeAccessMgr byte = iota
)

const (
	resendPeriod          = 3 * time.Second
	distDebugTick         = 10 * time.Second
	distTimeTick          = 1 * time.Second
)
```

packages/chains/accessMgr/accessMgr.go
muXxer: I created another PR #1626 to fix what I mean.

packages/chains/accessMgr/accessMgr.go
muXxer: ok perfect.

packages/chains/accessMgr/accessMgr.go
muXxer: Ok but still wouldn't it be better to have `type NodeID [32]byte` then?
So we don't have the string comparisons everywhere.

muXxer: I added a PR #1630 to do these changes.

packages/chains/accessMgr/accessMgr_test.go
muXxer: I created a PR #1631 to address that.

packages/chains/accessMgr/accessMgr.go
muXxer: I created a PR #1631 to address that.

muXxer (approved):

PR #1629, opened 2023-01-09
Fix: larger webapi timeout

jorgemmsilva:
I spent some hours trying to figure out why the webapi was closing http requests (with EOF) on the spam tests.

Turns out that when the node is overloaded, it takes a while to respond to the requests. This PR alleviates this issue by increasing the webapi read/write timeouts to more forgiving values

### DIFF
--- a/plugins/webapi/component.go
+++ b/plugins/webapi/component.go
@@ -102,8 +102,9 @@ func provide(c *dig.Container) error {
 			nil,
 			ParamsWebAPI.DebugRequestLoggerEnabled,
 		)
-		e.Server.ReadTimeout = 20 * time.Second
-		e.Server.WriteTimeout = 30 * time.Second
+
+		e.Server.ReadTimeout = ParamsWebAPI.ReadTimeout
+		e.Server.WriteTimeout = ParamsWebAPI.WriteTimeout
 
 		e.HidePort = true
 		e.HTTPErrorHandler = httperrors.HTTPErrorHandler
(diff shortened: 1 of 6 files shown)


plugins/webapi/component.go
```diff
@@ -102,8 +102,8 @@ func provide(c *dig.Container) error {
 			nil,
 			ParamsWebAPI.DebugRequestLoggerEnabled,
 		)
-		e.Server.ReadTimeout = 5 * time.Second
-		e.Server.WriteTimeout = 10 * time.Second
+		e.Server.ReadTimeout = 20 * time.Second
```
muXxer: Should we maybe add config parameters for these two?

tools/cluster/tests/missing_requests_test.go
```diff
@@ -52,12 +52,6 @@ func TestMissingRequests(t *testing.T) {
 	err = clu.WaspClient(1).PostOffLedgerRequest(chainID, req)
 	require.NoError(t, err)
 
-	// TODO try to send to only 2 nodes
-	err = clu.WaspClient(2).PostOffLedgerRequest(chainID, req)
```
muXxer, line 56: This was removed on purpose, right? :sweat_smile:

jorgemmsilva: yeah, the TODO was to do exactly that :smile:

plugins/webapi/component.go
jorgemmsilva: sounds reasonable, I'll add it

dessaya (approved):

PR #1635, opened 2023-01-10
Consensus redesign - rotation tests fixed.

kape1395:


packages/chain/mempool/mempool.go
```diff
@@ -777,10 +782,16 @@ func (mpi *mempoolImpl) pubKeyAsNodeIDMap(nodePubKey *cryptolib.PublicKey, _ int
...
 
 // Have to have it as a separate function to be able to use type params.
-func unprocessedPredicate[V isc.Request](chainState state.State) func(V, time.Time) bool {
+func unprocessedPredicate[V isc.Request](chainState state.State, log *logger.Logger) func(V, time.Time) bool {
 	return func(request V, ts time.Time) bool {
 		requestID := request.ID()
-		if processed, err := blocklog.IsRequestProcessed(chainState, &requestID); err != nil && processed {
+		processed, err := blocklog.IsRequestProcessed(chainState, &requestID)
+		if err != nil {
+			log.Debugf("Cannot check if request %v is processed at state.TrieRoot=%v, err=%v", requestID, chainState.TrieRoot(), err)
```
jorgemmsilva: should log a warn or error here?

tools/cluster/tests/publisher_test.go
```diff
@@ -61,6 +61,8 @@ func assertMessages(t *testing.T, messages []string, expectedFinalCounter int) {
 
 // TODO the TODOs on this test indicate that there is a race condition with the "await request endpoint", needs to be debugged
 func TestNanoPublisher(t *testing.T) {
+	t.Skip("Nano publisher is going to be replaced, so the test is skipped for now.") // TODO: Re-enable it!
```
jorgemmsilva: this skip will trigger some linting warns, better leave it un-skipped. It should pass with the "TODO blocks" that were removed (or I forgot to commit :trollface: ).

jorgemmsilva (approved):

tools/cluster/tests/publisher_test.go
jorgemmsilva: this is what I mean btw:
```diff
diff --git a/tools/cluster/tests/publisher_test.go b/tools/cluster/tests/publisher_test.go
index 301846a85..e3d5735c3 100644
--- a/tools/cluster/tests/publisher_test.go
+++ b/tools/cluster/tests/publisher_test.go
@@ -94,6 +94,12 @@ func TestNanoPublisher(t *testing.T) {
 		req, err := myClient.PostOffLedgerRequest(inccounter.FuncIncCounter.Name, chainclient.PostRequestParams{Nonce: uint64(i + 1)})
 		require.NoError(t, err)
 
+		// ---
+		// TODO shouldn't be needed
+		_, err = env.Chain.CommitteeMultiClient().WaitUntilRequestProcessedSuccessfully(env.Chain.ChainID, req.ID(), 30*time.Second)
+		require.NoError(t, err)
+		// ---
+
 		reqIDs[i] = req.ID()
 	}
 

```

jorgemmsilva: with that the test will pass

packages/chain/mempool/mempool.go
kape1395: Good point. Will change to Warn.

tools/cluster/tests/publisher_test.go
kape1395: OK, will fix this test instead of skipping it.
The problem is not only in the case of removal of that "TODO shouldn't be needed". 2 last blocks are produced twice, maybe L1 rejects one of them. Still checking.

muXxer (approved):

Issue #1638, opened 2023-01-10
Optimization of "receiveRequest" endpoint


jorgemmsilva:
![Image](https://user-images.githubusercontent.com/1187222/211587200-06c6ffca-adca-4a30-91f0-20c162df0444.png)

the way we fetch the state can be improved, also 3 views are called, when a single one should suffice

PR #1643, opened 2023-01-10
Fix: optimize receive req endpoint

jorgemmsilva:
https://github.com/iotaledger/wasp/issues/1638

Selection_003


dessaya (approved):

packages/webapi/v1/request/request.go
```diff
@@ -57,34 +51,33 @@ func AddEndpoints(
...
 
 	if o.requestsCache.Get(reqID) != nil {
 		return httperrors.BadRequest("request already processed")
 	}
+	o.requestsCache.Set(reqID, true)
 
 	// check req signature
-	if err := offLedgerReq.VerifySignature(); err != nil {
+	if err := req.VerifySignature(); err != nil {
 		o.requestsCache.Set(reqID, true)
```
muXxer, line 75:
```suggestion
```

muXxer (approved):

muXxer (changes requested):
in `EnqueueOffLedgerRequest` in `offledger.go` there is a `c.requestCache.Set(reqID, true)` missing in line 61.

line 64 can be removed as well.

packages/webapi/v1/request/request.go
kape1395: Please wait for me with this.

kape1395 (changes requested):
Let's use the latest state from the chain.

packages/vm/core/blocklog/external.go
```diff
@@ -38,7 +38,7 @@ func GetRequestIDsForBlock(stateReader kv.KVStoreReader, blockIndex uint32) ([]i
 
 func GetRequestReceipt(stateReader kv.KVStoreReader, requestID isc.RequestID) (*RequestReceipt, error) {
 	partition := subrealm.NewReadOnly(stateReader, kv.Key(Contract.Hname().Bytes()))
-	return isRequestProcessedInternal(partition, requestID)
+	return IsRequestProcessedInternal(partition, requestID)
```
kape1395: Doesn't it sound suspicious to have the "*Internal" function public?

packages/webapi/v1/endpoints.go
```diff
@@ -53,9 +52,6 @@ func Init(
 	request.AddEndpoints(
 		pub,
 		chainsProvider.ChainProvider(),
-		chainutil.GetAccountBalance,
```
kape1395, line 56: Maybe we should leave these endpoints in the API even if we don't use them internally?

kape1395 (approved):
Several minor comments.

packages/webapi/v1/endpoints.go
jorgemmsilva: those are not endpoints :joy: 

the function name is misleading. those were just utility functions that were used before, not anymore.

packages/vm/core/blocklog/external.go
jorgemmsilva: you're right, addressing that :+1:

packages/chain/chain.go
```diff
@@ -22,15 +22,24 @@ type NodeConnection interface {
 	Run(ctx context.Context)
 }
 
+type StateFreshness byte
+
+const (
+	LatestState    StateFreshness = iota // ActiveState, if exist; Confirmed state otherwise.
```
dessaya: Maybe rename to something like `StateFreshnessAuto` or `Default`? Because this looks repetitive and not very clear:

```
Chain.LatestState(chain.LatestState)
```

dessaya (approved):

packages/chain/chain.go
kape1395: Renamed it to ActiveOrCommittedState.

packages/webapi/v1/endpoints.go
muXxer: got confused the same way yesterday hahaha :D

packages/chain/node.go
```diff
@@ -839,6 +852,35 @@ func (cni *chainNodeImpl) LatestAliasOutput() (confirmed, active *isc.AliasOutpu
...
+	case ActiveState:
+		if latestActiveState == nil {
+			return nil, fmt.Errorf("chain %v has no active state", cni.chainID)
+		}
+		return latestActiveState, nil
+	case ConfirmedState:
+		if latestConfirmedState == nil {
+			return nil, fmt.Errorf("chain %v has no confirmed state", cni.chainID)
+		}
+		return latestConfirmedState, nil
```
muXxer: Why don't we call the store here like we do in `ActiveOrCommittedState` ?

packages/webapi/v2/services/vm.go
```diff
@@ -3,11 +3,13 @@ package services
 import (
 	"errors"
 
+	"github.com/iotaledger/wasp/packages/chain"
```
muXxer: Its imported twice. See next line.

packages/webapi/v2/services/offledger.go
```diff
@@ -71,32 +77,44 @@ func (c *OffLedgerService) EnqueueOffLedgerRequest(chainID isc.ChainID, binaryRe
...
-	}
+	return nil
+}
 
-	// check user has on-chain balance
-	assets, err := chainutil.GetAccountBalance(chain, request.SenderAccount())
+// implemented this way so we can re-use the same state, and avoid the overhead of calling views
+// TODO exported just to be used by V1 API until that gets deprecated at once.
+func ShouldBeProcessed(ch chain.ChainCore, req isc.OffLedgerRequest) error {
+	// state, err := ch.GetStateReader().LatestState()
```
muXxer: Why is that commented code still here?

packages/webapi/v2/services/chain.go
```diff
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/iotaledger/wasp/packages/chain"
```
muXxer: imported twice, see next line.

packages/solo/chain.go
```diff
@@ -661,6 +662,23 @@ func (ch *Chain) LatestAliasOutput() (confirmed *isc.AliasOutputWithID, active *
 	return ao, ao
 }
 
+// LatestState implements chain.Chain
+func (ch *Chain) LatestState(freshness chain.StateFreshness) (state.State, error) {
+	ao := ch.GetAnchorOutput()
+	if ao == nil {
+		return ch.store.LatestState()
+	}
+	l1c, err := state.L1CommitmentFromAliasOutput(ao.GetAliasOutput())
```
muXxer, line 671: Do we really want to panic the whole wasp node if one chain it follows has a state with invalid metadata?

packages/solo/evm.go
```diff
@@ -44,12 +46,20 @@ func (b *jsonRPCSoloBackend) EVMGasPrice() *big.Int {
...
+func (b *jsonRPCSoloBackend) ISCLatestState() state.State {
+	latestState, err := b.Chain.LatestState(chain.ActiveOrCommittedState)
+	if err != nil {
+		panic(err)
+	}
+	return latestState
+}
+
+func (b *jsonRPCSoloBackend) ISCStateByBlockIndex(blockIndex uint32) (state.State, error) {
+	return b.Chain.store.StateByIndex(blockIndex)
```
muXxer, line 62: Should we also add the same logic like in the jsonrpc waspevmbackend?

```go
	latestState, err := b.chain.LatestState(chain.ActiveOrCommittedState)
	if err != nil {
		return nil, fmt.Errorf("couldn't get latest state: %s", err.Error())
	}
	if latestState.BlockIndex() == blockIndex {
		return latestState, nil
	}
	return b.chain.Store().StateByIndex(blockIndex)
```

packages/solo/chain.go
jorgemmsilva: its not the "wasp node" its just solo, that's okay

packages/solo/evm.go
jorgemmsilva: solo is just the test framework, this optimization wouldn't be a big benefit, rather keep the simpler code

packages/webapi/v2/services/chain.go
jorgemmsilva: well spotted!

packages/solo/chain.go
muXxer: ah, true. ok feel free to ignore :D

packages/chain/node.go
jorgemmsilva: this was written by Karolis, but my interpretation is that we might not have an "active state", but we must always have a `latestConfirmedState`

packages/solo/evm.go
muXxer: Yeah I thought the benefit is not the optimization, but the fact that we "test" the exact same "optimization code" as we use in production.

packages/chain/node.go
muXxer: But if we always must have a `latestConfirmedState`, why is there a fallback in `ActiveOrCommittedState` but not in `ConfirmedState`. Thats what I don't get :sweat_smile:

jorgemmsilva: good point, maybe we can remove the fallback, or add it on the commented line.
@kape1395 wdyt?

muXxer: @kape1395 can you elaborate? then we can merge this PR. :+1:

kape1395: Maybe that's a good point.
  - I was assuming the ConfirmedState case should be more strict, but probably it is correct enough to fall back to the store here.
  - Changed to use a fallback.
  - The situation, when the ConfirmedState is not yet available (apart from the Store) is the startup when the chain is started and the state is not yet fetched from the store.
  - Reorganized code a bit.

muXxer (approved):

PR #1647, opened 2023-01-11
Statemanager BugFix

Juliusan:
# Description of change

State manager was not directly committing to the store state drafts coming from consensus. This bug is fixed, tests written to demonstrate that. And other minor changes:
* adapting state manager logging to `NodeID` change from string to byte array: logging only part of the array
* `NodeID.ShortString()` returns first four bytes of byte array with `0x` in front for consistency with other `ShortString` methods (`ChainID`); for test purposes `NodeID` is generated in 4 byte BigEndian format to get a human understandable IDs "0x00000001", "0x00000002" instead of "0x01000000".
* function to obtain short strings slice from a slice of objects that implement `ShortString() string` and test for this function
* test to make sure that commit to the store of the same block twice doesn't panic
* some fogotten `pipe` documentation

## Type of change

- Documentation Fix

The new tests pass as well as all of the ones in `make test` except those, that do not pass in `develop` too.

## Change checklist

### DIFF
--- a/packages/gpa/interface.go
+++ b/packages/gpa/interface.go
@@ -6,13 +6,15 @@ package gpa
 
 import (
 	"encoding"
-	"strings"
 
 	iotago "github.com/iotaledger/iota.go/v3"
 	"github.com/iotaledger/wasp/packages/cryptolib"
+	"github.com/iotaledger/wasp/packages/util"
 )
 
 type NodeID [32]byte
+
+var _ util.ShortStringable = NodeID{}
 
 func NodeIDFromPublicKey(pubKey *cryptolib.PublicKey) NodeID {
 	nodeID := NodeID{}
@@ -37,7 +39,7 @@ func (niT NodeID) String() string {
 }
 
 func (niT NodeID) ShortString() string {
-	return strings.TrimPrefix(iotago.EncodeHex(niT[:8]), "0x")
+	return iotago.EncodeHex(niT[:4]) // 4 bytes - 8 hexadecimal digits
 }
 
 type Message interface {
--- a/packages/gpa/test_utils.go
+++ b/packages/gpa/test_utils.go
@@ -11,9 +11,9 @@ import (
 func MakeTestNodeIDFromIndex(index int) NodeID {
 	nodeID := NodeID{}
 
-	indexBytes := make([]byte, 8)
-	binary.LittleEndian.PutUint64(indexBytes, uint64(index))
-	copy(nodeID[:8], indexBytes)
+	indexBytes := make([]byte, 4)
+	binary.BigEndian.PutUint32(indexBytes, uint32(index))
+	copy(nodeID[:4], indexBytes)
 
 	return nodeID
 }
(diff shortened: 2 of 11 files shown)


kape1395 (approved):

packages/util/strutil.go
```diff
@@ -27,3 +27,18 @@ func TimeOrNever2(t time.Time, never string) string {
 func TimeOrNever(t time.Time) string {
 	return TimeOrNever2(t, "never")
 }
+
+type ShortStringable interface {
+	ShortString() string
+}
+
+func SliceShortString[E ShortStringable](slice []E) string {
```
muXxer, line 35: maybe we can use
```suggestion
func SliceShortString[E ShortStringable](args ...E) string {
```
instead?

packages/gpa/test_utils.go
```diff
@@ -11,9 +11,9 @@ import (
 func MakeTestNodeIDFromIndex(index int) NodeID {
 	nodeID := NodeID{}
 
-	indexBytes := make([]byte, 8)
-	binary.LittleEndian.PutUint64(indexBytes, uint64(index))
-	copy(nodeID[:8], indexBytes)
+	indexBytes := make([]byte, 4)
+	binary.BigEndian.PutUint32(indexBytes, uint32(index))
+	copy(nodeID[:4], indexBytes)
```
muXxer, line 16: Can we also copy the bytes at the end of "nodeID" ?
```suggestion
	copy(nodeID[NodeIDLength-4:NodeIDLength], indexBytes)
```
And also change the ShortString method to use these bytes?
Otherwise we have  `0x000000010000000000000000000000000000000000000000000000000000000` as the "full node ID" instead of `0x000000000000000000000000000000000000000000000000000000000000001`

packages/gpa/interface.go
```diff
@@ -37,7 +39,7 @@ func (niT NodeID) String() string {
 }
 
 func (niT NodeID) ShortString() string {
-	return strings.TrimPrefix(iotago.EncodeHex(niT[:8]), "0x")
+	return iotago.EncodeHex(niT[:4]) // 4 bytes - 8 hexadecimal digits
```
muXxer, line 42: See other comment in `test_utils.go` or ignore if you want :)

packages/util/strutil.go
Juliusan: Is there any benefit? It would be easier to call if you have several separate objects but you would have to use `...` for a slice. I see no real difference.

muXxer: It was only a suggestion because in the golang standard libs its often used that way for strings. Example:
```go
// Sprint formats using the default formats for its operands and returns the resulting string.
// Spaces are added between operands when neither is a string.
func Sprint(args ...any) string {
```
But feel free to ignore it.

dessaya (approved):

packages/gpa/test_utils.go
Juliusan: Well... `ShortString` always takes first 4 bytes. (see e.g. `ChainID`). To be consistent, `ShortString` for `NodeID` also takes first 4 bytes. And this `NodeID` generation is just for tests, so it adapts to the real code (`ShortString`). Besides, the full ID is usually not logged.

PR #1649, opened 2023-01-11
A bit of cleanup around the crypto part.

kape1395:


packages/dkg/node.go
```diff
@@ -114,7 +114,7 @@ func (n *Node) GenerateDistributedKey(
 	if peerCount < 1 || threshold < 1 || threshold > peerCount {
 		return nil, invalidParams(fmt.Errorf("wrong DKG parameters: N = %d, T = %d", peerCount, threshold))
 	}
-	if threshold < peerCount/2+1 {
+	if threshold < peerCount/2+1 { // TODO: Incorrect.
```
jorgemmsilva, line 117: what would be the correct formula? Could we just update it?

jorgemmsilva (approved):

packages/dkg/node.go
kape1395: I will unify those formulas in one place. Because now they are in a lot of places, sometimes differ a bit.

PR #1659, opened 2023-01-12
Add optional tips to doPostTx

muXxer:
We need this to be able to chain transactions on L1.

### DIFF
--- a/packages/nodeconn/nodeconn.go
+++ b/packages/nodeconn/nodeconn.go
@@ -364,9 +364,29 @@ func (nc *nodeConnection) GetMetrics() nodeconnmetrics.NodeConnectionMetrics {
 	return nc.nodeConnectionMetrics
 }
 
-func (nc *nodeConnection) doPostTx(ctx context.Context, tx *iotago.Transaction) (iotago.BlockID, error) {
+func (nc *nodeConnection) doPostTx(ctx context.Context, tx *iotago.Transaction, tipsAdditional ...iotago.BlockID) (iotago.BlockID, error) {
+	var parents iotago.BlockIDs
+
+	// if no tips are given, we use empty parents, the node will do the tipselection in that case.
+	// if tips are given, we should have at least BlockMaxParents/2 parents to have a healthy tangle.
+	if len(tipsAdditional) > 0 {
+		// add the tips to parents we want to reference.
+		parents = append(parents, tipsAdditional...)
+
+		if len(parents) < (iotago.BlockMaxParents / 2) {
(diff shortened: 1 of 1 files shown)


packages/nodeconn/nodeconn.go
```diff
@@ -364,9 +364,29 @@ func (nc *nodeConnection) GetMetrics() nodeconnmetrics.NodeConnectionMetrics {
 	return nc.nodeConnectionMetrics
 }
 
-func (nc *nodeConnection) doPostTx(ctx context.Context, tx *iotago.Transaction) (iotago.BlockID, error) {
+func (nc *nodeConnection) doPostTx(ctx context.Context, tx *iotago.Transaction, tipsAdditional ...iotago.BlockID) (iotago.BlockID, error) {
+	var parents iotago.BlockIDs
+
+	// if no tips are given, we use empty parents, the node will do the tipselection in that case.
+	// if tips are given, we should have at least BlockMaxParents/2 parents to have a healthy tangle.
```
jorgemmsilva, line 371: are these details documented anywhere?
Its awesome that you know about them, but this feels really obscure

jorgemmsilva (approved):

packages/nodeconn/nodeconn.go
muXxer: See the official Tangle improvement proposals repo :sweat_smile: 
https://github.com/iotaledger/tips/blob/main/tips/TIP-0025/tip-0025.md
grafik

kape1395:
That's something intermediate? Who will pass the additional parameters to doPostTx?

muXxer:
> That's something intermediate? Who will pass the additional parameters to doPostTx?

yes, intermediate but non breaking.

packages/nodeconn/nodeconn.go
jorgemmsilva: it doesn't talk about:
>we should have at least BlockMaxParents/2 parents to have a healthy tangle.

muXxer: This info doesn't exist anywhere AFAIK. Maybe research has some simulations on this topic.
The idea is that we only pass 1 single additional tip because we want to reference the last transaction of the chain.
But 1 tip is bad, because then we basically form a chain and if every participant in the tangle would do this, the global confirmation rate would drop significantly.

For this reason IOTA uses also 4 parents as default value in the node since chrysalis part 1.
4 was decided by me basically back then based on spam tests. :stuck_out_tongue:

PR #1668, opened 2023-01-13
Show notification to login if a request fails due to failed authorization

lmoe:


client/client.go
```diff
@@ -45,6 +48,10 @@ func processResponse(res *http.Response, decodeTo interface{}) error {
 	}
 	defer res.Body.Close()
 
+	if res.StatusCode == http.StatusUnauthorized {
```
muXxer, line 50: Why not move the `defer res.Body.Close()` to the top of the function, and then check the statuscode, and then do the `ReadAll`? Isn't that possible? Can and should we do the `defer res.Body.Close()` even if the `ReadAll` fails?

lmoe: The client will be removed soon, I've just modified it quickly to report a checkable error to the cli. Don't think we should invest more time than needed on it.

muXxer: Instead of writing that comment, you could have fixed it :trollface:

muXxer: Fixed it for you :)

muXxer (approved):

jorgemmsilva (approved):

PR #1677, opened 2023-01-17
feat(evm): ERC721 contract for IRC27 NFT collection

dessaya:
Some TODOs for following PRs:

* Querying the amount of owned NFTs in a given collection is inefficient; this probably requires a change in the accounts core contract.

* Minting NFTs from L2 is not supported yet; for now they have to be minted in L1, then deposited on the chain.

* The ERC721 contract is not automatically deployed. The collection NFT has to be deposited on the chain in order to register the ERC721 contract (this is necessary to extract its metadata). Afterwards it may be withdrawn to L1.


packages/solo/solo.go
```diff
@@ -547,17 +547,34 @@ type NFTMintedInfo struct {
...
+	}
+	return nfts[0], infos[0], nil
+}
+
+// MintNFTsL1 mints len(immutableMetadata) NFTs with the `issuer` account and sends them
+// to a `target` account.
+//
+// If collectionOutputID is not nil, it must be an outputID of an NFTOutput owned by the issuer.
+// All minted NFTs will belong to the given collection.
+// See: https://github.com/Kami-Labs/tips/blob/main/tips/TIP-0027/tip-0027.md
```
jorgemmsilva: any reason in particular to link to the Kami-Labs repo, instead of:
https://github.com/iotaledger/tips/blob/main/tips/TIP-0027/tip-0027.md
?

packages/transaction/nfttransaction.go
```diff
@@ -1,38 +1,66 @@
...
-		NFTID: iotago.NFTID{},
-		Conditions: iotago.UnlockConditions{
-			&iotago.AddressUnlockCondition{Address: par.Target},
-		},
-		ImmutableFeatures: iotago.Features{
-			&iotago.IssuerFeature{Address: issuerAddress},
-			&iotago.MetadataFeature{Data: par.ImmutableMetadata},
-		},
+		out := collectionOutput.Clone().(*iotago.NFTOutput)
+		out.NFTID = collectionID
```
jorgemmsilva, line 46: is this correct? shouldn't the output have its own unique ID? (be empty at this point)

packages/vm/core/accounts/internal.go
```diff
@@ -432,6 +433,19 @@ func getAccountNFTs(account *collections.ImmutableMap) []iotago.NFTID {
 	return ret
 }
 
+func getAccountNFTsInCollection(state kv.KVStoreReader, account *collections.ImmutableMap, collectionID iotago.NFTID) []iotago.NFTID {
+	// TODO: this may be inefficient if the account owns many NFTs not in the collection
```
jorgemmsilva, line 437: This should be fine, no? how would we optimize this? maintaining an index by collection id?

packages/vm/core/evm/iscmagic/ERC721NFTCollection.sol
```diff
@@ -0,0 +1,151 @@
...
+
+pragma solidity >=0.8.11;
+
+import "@iscmagic/ISCTypes.sol";
+import "@iscmagic/ISCSandbox.sol";
+import "@iscmagic/ISCAccounts.sol";
+import "@iscmagic/ISCPrivileged.sol";
+
+// The ERC721 contract for a L2 collection of ISC NFTs, as defined in IRC27:
+// https://github.com/Kami-Labs/tips/blob/main/tips/TIP-0027/tip-0027.md
```
jorgemmsilva: another link to Kami-Labs repo

jorgemmsilva (approved):

packages/solo/solo.go
dessaya: No reason, other than it appears first when doing a google search :stuck_out_tongue: 
I'll change it :+1:

packages/transaction/nfttransaction.go
dessaya: It's correct. This sets up the output of the *collection NFT* which is already minted and is taken as an input of the tx.

packages/vm/core/accounts/internal.go
dessaya: I don't know what would be the best way. Perhaps maintaining an index, but I'd like to avoid duplicating information if possible.

I also don't know the usage patterns, but this is O(n) with the amount of NFTs and native tokens which are virtually unlimited. IMO something as simple as querying the amount of owned NFTs should be O(1).

packages/transaction/nfttransaction.go
jorgemmsilva: okay, that makes sense, I was reading this wrong, thanks!

PR #1682, opened 2023-01-18
Displaying commit on wasp start

Juliusan:
Earlier on starting wasp the exact commit was logged. This change adds this information to current Wasp implementation. The exact commit is very useful in debugging the errors. Moreover, it might be used to check if the code is run on the code you are expecting it to be run. The version with commit is extracted using `git describe --tags`. Now the start of node looks like this:
```>>>>> Starting Wasp vv0.4.0-alpha.2-265-gbc29416c2 <<<<<```
instead of this:
```>>>>> Starting Wasp v0.4.0-alpha.2 <<<<<```
Note the double "v" in front of the number: one comes from `app` package in `hive.go`, another - from tag, used in `wasp`. [SemVer standard does not allow "v" in front of a number](https://semver.org/#is-v123-a-semantic-version), although `github.com/hashicorp/go-version` package allows it. I suggest dropping "v" from tags in `wasp`.
Also note than in a case where commit matches the tag only the tag is included, so for builds from tags the view will be (almost) unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 GIT_COMMIT_SHA := $(shell git rev-list -1 HEAD)
+GIT_VERSION := $(shell git describe --tags)
 BUILD_TAGS = rocksdb
-BUILD_LD_FLAGS = "-X=github.com/iotaledger/wasp/packages/wasp.VersionHash=$(GIT_COMMIT_SHA)"
+BUILD_LD_FLAGS = "-X=github.com/iotaledger/wasp/packages/wasp.VersionHash=$(GIT_COMMIT_SHA)\
+				  -X=github.com/iotaledger/wasp/packages/wasp.Version=$(GIT_VERSION)"
 
 #
 # You can override these e.g. as
--- a/core/app/app.go
+++ b/core/app/app.go
@@ -27,7 +27,7 @@ import (
 )
 
 func App() *app.App {
-	return app.New(wasp.Name, wasp.Version+"+"+wasp.VersionHash,
+	return app.New(wasp.Name, wasp.Version,
 		app.WithVersionCheck("iotaledger", "wasp"),
 		app.WithInitComponent(InitComponent),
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by jorgemmsilva — COMMENTED** [review]
there is a linting issue.

If we want to do it like this, we need to update `docker-push.yml`,  and 2 `waypoint.hcl` files.

**Comment by muXxer** [comment]
Can we do it the way we do it in hornet? :D
There we always use the Version number if we compile it without the script in the repo (for example via github CI), or we replace the version number with the commit hash when compiling via script or make build (because the version number makes no sense in this context then).
This also removes the leading "v", see the following code in app.go:

```go
	versionString := a.Info().Version
	if _, err := goversion.NewSemver(a.Info().Version); err == nil {
		// version is a valid SemVer => release version
		versionString = "v" + versionString
	} else {
		// version is not a valid SemVer => maybe self-compiled
		versionString = "commit: " + versionString
	}
```

muXxer:
```
./wasp
>>>>> Starting Wasp commit: 78be23490ddd5aa140098a9b4c9bcd2c8b41bedd <<<<<
```

I can push the needed changes if you agree.

Juliusan:
I've seen that code and was wondering, what it is used for. Thanks for the explanation. I agree that if we use hive, we should do it the way hive wants. In fact I don't care much, how it is done, as far as I can be sure that if I get a log from community, I know that it was produced by this exact commit. Github CI is used simply to run tests and it is usually clear, what is the exact commit of these runs. But how are we going to build the production releases? Is it possible that the production release is going to be built in the repo but on tag+2 commits thus pointing to a tag but actually being tag+2 commits? My solution solves these problems. But if your solution is equivalent, except it is how hive app library expects it to be done, then you are welcome to overwrite my changes.

muXxer:
I tried to address all your comments in #1696 and also use the nice `git describes --tags` idea.

PR #1686, opened 2023-01-19
Wal

Juliusan:
# Description of change

Wal is used to obtain a block. Also `rawBlocks` config parameter renamed to `wal` and made `wal` enabled by default.

## Type of change

`make test`. New tests added and old ones improved to test the changes.

## Change checklist


packages/util/equatable.go
```diff
@@ -34,6 +34,16 @@ func RemoveAll[E Equatable](objectsToRemove, objects []E) []E {
...
+func Intersection[E Equatable](objects1, objects2 []E) []E {
+	result := make([]E, 0)
+	for _, o := range objects1 {
+		if Contains(o, objects2) {
+			result = append(result, o)
+		}
+	}
+	return result
+}
+
```
jorgemmsilva: could this be used instead?
https://github.com/samber/lo#intersect

jorgemmsilva: either that or `Contains`/`ContainsBy`

packages/chain/statemanager/smGPA/smGPAUtils/interface.go
```diff
@@ -16,6 +16,13 @@ type BlockWAL interface {
 	Write(state.Block) error
 	Contains(state.BlockHash) bool
 	Read(state.BlockHash) (state.Block, error)
+	Contents() []state.BlockHash
+}
+
+// May be used only for tests; deleting in production should not be available.
+type TestBlockWAL interface {
+	BlockWAL
+	Delete(state.BlockHash) bool
```
jorgemmsilva: can this be moved to a test file?

packages/chain/statemanager/smGPA/smGPAUtils/block_wal_mocked.go
```diff
@@ -34,3 +41,21 @@ func (mbwT *mockedBlockWAL) Read(blockHash state.BlockHash) (state.Block, error)
...
+
+func (mbwT *mockedBlockWAL) Contents() []state.BlockHash {
+	result := make([]state.BlockHash, len(mbwT.walContents))
+	i := 0
+	for blockHash := range mbwT.walContents {
+		result[i] = blockHash
+		i++
+	}
+	return result
+}
```
jorgemmsilva: can this be moved to a test file?

packages/chain/statemanager/smGPA/smGPAUtils/block_wal_db.go
```diff
@@ -0,0 +1,50 @@
+package smGPAUtils
+
+import (
+	"fmt"
+
+	"github.com/iotaledger/wasp/packages/state"
+)
+
+// Function is not used currently
```
jorgemmsilva: remove it?

packages/chain/statemanager/smGPA/smGPAUtils/block_wal.go
```diff
@@ -95,6 +98,39 @@ func (bwT *blockWAL) Read(blockHash state.BlockHash) (state.Block, error) {
...
+		blockHashString := fileName[:len(fileName)-len(constFileSuffix)]
+		blockHash, err := state.BlockHashFromString(blockHashString)
+		if err != nil {
+			bwT.LogWarnf("Wal folder file %s is not named after block hash: %w", fileName, err)
+			continue
+		}
+		result = append(result, blockHash)
+	}
+	return result
+}
```
jorgemmsilva: should this return an error, instead of just logging warns?

jorgemmsilva (changes requested):

packages/util/equatable.go
dessaya: +1 for reusing `lo` whenever possible. In fact, I think the `Equatable` interface can be dropped because it is implemented by two types that are arrays, so they already implement the `comparable` constraint. So `lo.Intersect` and friends should work.

packages/chain/statemanager/smGPA/smGPAUtils/block_wal_mocked.go
Juliusan: No. These methods describe `mockedBlockWAL` struct and they belong together. It would be less clear if we scatter the methods all over the code. And `Contents` is in the main interface anyway.

packages/chain/statemanager/smGPA/smGPAUtils/block_wal_db.go
Juliusan: Will do. I spent on it plenty of time and committed it, so it was a shame to remove it :-)

packages/chain/statemanager/smGPA/smGPAUtils/block_wal.go
Juliusan: This function was needed for the method, which is not used and will be removed. For it it was enough to have a method, that silently returns what it can. I think, I'll simply drop this method.

packages/util/equatable.go
Juliusan: Yes, `Equatable` interface is only used for array types. But it is more flexible than `comparable` constraint. It could be used for any struct, which implement `Equals` method telling how to decide which two objects are equal. I can remove it for now though.

dessaya: If it is unused I'd vote for removing it, just so we have less code to maintain.

If you decide to keep it, I suggest:

* Renaming the interface, since I think "equatable" is not a real word (is it?). Maybe `Comparable`, or `Eq` like Haskell.

* Changeing the interface to:

```
type Eq[T any] interface {
  Equals(T other) bool
}
```

so that the `Equals` method can accept an object of the same type. Then:

```
func Intersection[T Eq[T]](objects1, objects2 []T) []T {
  ...
}
```

packages/chain/statemanager/smGPA/smGPAUtils/block_cache_rapid_no_wal_test.go
```diff
@@ -107,7 +109,7 @@ func (bcnwtsmT *blockCacheNoWALTestSM) Restart(t *rapid.T) {
 }
 
 func (bcnwtsmT *blockCacheNoWALTestSM) invariantAllBlocksInCacheDifferent(t *rapid.T) {
-	require.True(t, util.AllDifferent(bcnwtsmT.blocksInCache))
+	require.Equal(t, len(bcnwtsmT.blocksInCache), len(lo.Uniq(bcnwtsmT.blocksInCache)))
```
jorgemmsilva, line 112: I know this is from before, but this is such a weird name: `bcnwtsmT`, why not just `b`? :joy:

jorgemmsilva (approved):

packages/chain/statemanager/smGPA/smGPAUtils/block_cache_rapid_no_wal_test.go
Juliusan: Shortened from "BlockCacheNoWalTestStateMachine" :-D I would prefer "this", but it is prohibited :-) I rarely use one letter variables.

jorgemmsilva: agree using one letter variables is bad in general, but for receivers is very intuitive. Also easier to write than `zlxkTcNziolkj` :trollface:

PR #1699, opened 2023-01-23
refactor: avoid possible memory leaks with time.After

### LINKED ISSUE #1464: Refactor: avoid potentially timer memory leaks
- avoid using time.After in select cases (https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/)
- drain stopped timer channels using https://github.com/iotaledger/hive.go/blob/548bc381ee8c6ce39f36ffc71f4aed83b1f46ba3/core/timeutil/timeutil.go#L8

ivange94:
# Description of change

Avoid potential memory leaks with `timer.After`. 

Fixes https://github.com/iotaledger/wasp/issues/1464


tools/cluster/cluster.go
```diff
@@ -450,9 +451,12 @@ func (clu *Cluster) start() error {
 	}
 
 	for i := 0; i < len(clu.Config.Wasp); i++ {
+		delay := time.NewTimer(10 * time.Second)
+		defer timeutil.CleanupTimer(delay)
```
muXxer: you don't need to change the logic in the cluster tests or test cases, because its irrelevant for the test cases and is harder to read.

this change here introduces a resource leak because of the defer in the loop, please revert.

tools/cluster/cluster.go
```diff
@@ -526,13 +533,16 @@ func (clu *Cluster) RestartNodes(nodeIndex ...int) error {
 		if err != nil {
 			return err
 		}
+		delay := time.NewTimer(5 * time.Second)
```
muXxer: you don't need to change the logic in the cluster tests or test cases, because its irrelevant for the test cases and is harder to read.

this change here introduces a resource leak because of the defer in the loop, please revert.

packages/peering/lpp/lppNetImpl.go
```diff
@@ -493,8 +494,11 @@ func (n *netImpl) usePeer(remotePubKey *cryptolib.PublicKey) (peering.PeerSender
 
 func (n *netImpl) maintenanceLoop(stopCh chan bool) {
 	for {
+		delay := time.NewTimer(maintenancePeriod)
```
muXxer, line 500: This introduces a resource leak.
IMO you should put the whole code in an inline func, and then call that in the for loop.
This way you can do the `defer` in the inline func.

muXxer (changes requested):
Can you also fix the missing `time.After` in:
```
packages/peering/group/group.go
packages/dkg/proc.go
packages/chain/statemanager/smGPA/smGPAUtils/block_cache.go
packages/chain/statemanager/smGPA/smGPAUtils/time_provider_default.go
packages/chain/cons/gr/gr.go
```

packages/peering/lpp/lppNetImpl.go
ivange94: After looking at this piece again, the `defer` was not necessary. Right now I just clean up the timer inside 
`case <-stopCh`. So if the timer fires before the stop signal is sent, the `timer.C` is read. And since a stop signal has not being sent, the loop restarts and a new timer is initialized which is how the original logic was. 

Using `defer` in an inline function, I ended up with this 
````
for {
	stop := func() bool {
		delay := time.NewTimer(maintenancePeriod)
                defer timeutil.CleanupTimer(delay)

		select {
		case <-delay.C:
			n.peersLock.RLock()
			for _, p := range n.peers {
				p.maintenanceCheck()
			}
			n.peersLock.RUnlock()
			return false
		case <-stopCh:
			timeutil.CleanupTimer(delay)
			return true
		}
	}()
	if stop {
		return
	}
}
````
To prevent the loop from running indefinitely even after the stop signal has been received, I had to introduce a new variable to check the return status of the inline function. I find the former version much more readable and unless I missed something, the both have the same effect.

ivange94:
@muXxer concerning
>Can you also fix the missing time.After in:

>packages/peering/group/group.go
packages/dkg/proc.go
packages/chain/statemanager/smGPA/smGPAUtils/block_cache.go
packages/chain/statemanager/smGPA/smGPAUtils/time_provider_default.go
packages/chain/cons/gr/gr.go

1. `block_cache.go` does't use the package level function `time.After` but the struct function `Time.After` which returns a boolean and it's not even used in a `select` block so I'm not sure what you mean.

2. `time_provider_default.go` implements [TimeProvider](https://github.com/iotaledger/wasp/blob/develop/packages/chain/statemanager/smGPA/smGPAUtils/interface.go#L21) and changing the method implementation forces me to change the interface definition which intend forces me to modify a lot of test cases. And as you mentioned, fixing this in tests adds little value but makes the code much harder to read.

ivange94:
Closing this. Will open a new one with a cleaner commit history.

PR #1742, opened 2023-01-25
feat(evm): ERC721NFTs implements IERC721Metadata

dessaya:
Also reduce code duplication between ERC721NFTs and ERC721NFTCollection.


packages/vm/core/evm/iscmagic/ERC721NFTs.sol
```diff
@@ -22,14 +22,22 @@ contract ERC721NFTs { // is IERC721, IERC165
     event Approval(address indexed owner, address indexed approved, uint256 indexed tokenId);
     event ApprovalForAll(address indexed owner, address indexed operator, bool approved);
 
+    function _balanceOf(ISCAgentID memory owner) internal virtual view returns (uint256) {
+        return __iscAccounts.getL2NFTAmount(owner);
+    }
+
+    function _isInCollection(ISCNFT memory) internal virtual view returns (bool) {
+        return true;
```
jorgemmsilva, line 32: should the virtual func return false?
"vanilla ERC721" don't belong to a collection by default ,right?

packages/vm/core/evm/iscmagic/ERC721NFTs.sol
```diff
@@ -22,14 +22,22 @@ contract ERC721NFTs { // is IERC721, IERC165
...
     function balanceOf(address owner) public view returns (uint256) {
-        ISCAgentID memory ownerAgentID = ISCTypes.newEthereumAgentID(owner);
-        return __iscAccounts.getL2NFTAmount(ownerAgentID);
+        return _balanceOf(ISCTypes.newEthereumAgentID(owner));
     }
 
     function ownerOf(uint256 tokenId) public view returns (address) {
         ISCNFT memory nft = __iscSandbox.getNFTData(tokenId.asNFTID());
         require(nft.owner.isEthereum());
+        require(_isInCollection(nft));
```
jorgemmsilva: why do we check the collection here?
if its the owner, isn't that enough?

packages/vm/core/evm/iscmagic/ERC721NFTs.sol
```diff
@@ -119,6 +130,22 @@ contract ERC721NFTs { // is IERC721, IERC165
...
+        return "";
+    }
+
+    function symbol() external pure returns (string memory) {
+        return ""; // not defined in IRC27
+    }
+
+    function tokenURI(uint256 tokenId) external view returns (string memory) {
+        IRC27NFT memory nft = __iscSandbox.getIRC27NFTData(tokenId.asNFTID());
+        require(_isInCollection(nft.nft));
```
jorgemmsilva: why the collection check here?
just for `ERC721NFTCollection` to not have to override this, right?

packages/vm/core/evm/iscmagic/ERC721NFTs.sol
dessaya: Maybe the name of the function is confusing, but it is correct to return true here, meaning that the NFT is managed by this contract. What would be a better name? `_isManagedByThisContract`? :stuck_out_tongue: 

The ERC721NFTs contract allows access to all NFTs, regardless of whether they belong to a collection or not, so we return true here. The ERC721NFTCollection allows access to NFTs in a specific collection.

packages/vm/core/evm/iscmagic/ERC721NFTs.sol
dessaya: Yes, that's why it always returns true. Its only purpose is being overridden by a derived contract (namely ERC721NFTCollection).

packages/vm/core/evm/iscmagic/ERC721NFTs.sol
dessaya: That's right. It was the best way I found to have these two contracts that share 95% of the logic without code duplication.

Again, maybe it will be clearer if I change the name of that function.

jorgemmsilva: yeah, `_isManagedByThisContract` seems more descriptive :+1:

jorgemmsilva (approved):

PR #1790, opened 2023-01-31
 perf: improve EVM estimate gas performance 

dessaya:
Added benchmark. Before:

```
cpu: Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
BenchmarkRPCEstimateGas   676   17944494 ns/op   8640961 B/op   136120 allocs/op
```

After:

```
BenchmarkRPCEstimateGas  3540    3429016 ns/op   1705300 B/op    25347 allocs/op
```

Specific improvements:

* VM: Storage deposit estimation is costly. Calculate it just once.
* VM: Iterating stuff in the state is costly. Use array instead of map
  for block context subscriptions.
* VM: skip CloseBlockContexts() if estimating gas
* EVM: getBalance() called getFeePolicy() each time. Call it just once
  when creating the emulator instead.
* Ad caching to the trie store (I think this time it works :sweat_smile:  )


packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -83,22 +83,24 @@ func (bctx *blockContext) mintBlock() {
...
-		getSubBalanceFunc(ctx),
-		getAddBalanceFunc(ctx),
+		getBalanceFunc(ctx, feePolicy),
+		getSubBalanceFunc(ctx, feePolicy),
+		getAddBalanceFunc(ctx, feePolicy),
 	)
 }
 
 func createEmulatorR(ctx isc.SandboxView) *emulator.EVMEmulator {
+	feePolicy := getFeePolicy(ctx)
```
jorgemmsilva: won't this keep the wrong value if the fee policy is changed mid block?

jorgemmsilva (approved):
just 1 comment :ok_hand:

packages/vm/core/evm/evmimpl/internal.go
dessaya: That's right, good call! I'll fix it before merging.

PR #1793, opened 2023-02-01
Simplify chain deployment command

### LINKED ISSUE #1770: Chain deploy argument defaults: `--gov-controller` should default to the deployer address, defaults should be set to single validator chain (quorum=1, committee=0)
Would be good to make these changes to make sure the chain can be rotated again if needed and that we always start with a single node committee and rotate from there on out.

`chain deploy NAME` should be enough to deploy a single node chain with the deploying address.0 as the governor.

ivange94:
# Description of change
Fixes https://github.com/iotaledger/wasp/issues/1770
>Would be good to make these changes to make sure the chain can be rotated again if needed and that we always start with a single node committee and rotate from there on out.
chain deploy NAME should be enough to deploy a single node chain with the deploying address.0 as the governor.

### DIFF
--- a/tools/wasp-cli/chain/deploy.go
+++ b/tools/wasp-cli/chain/deploy.go
@@ -47,6 +47,13 @@ func isEnoughQuorum(n, t int) (bool, int) {
 	return t >= (n - maxF), maxF
 }
 
+func controllerAddressStr(addr string) string {
+	if addr != "" {
+		return addr
+	}
+	return viper.GetString("address.0")
+}
+
 func initDeployCmd() *cobra.Command {
 	var (
 		committee        []int
@@ -59,17 +66,9 @@ func initDeployCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "deploy",
 		Short: "Deploy a new chain",
-		Args:  cobra.NoArgs,
+		Args:  cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			l1Client := config.L1Client()
-			alias := GetChainAlias()
-
-			if committee == nil {
-				committee = getAllWaspNodes()
-			}
-			if quorum == 0 {
-				quorum = defaultQuorum(len(committee))
-			}
 
 			if ok, _ := isEnoughQuorum(len(committee), quorum); !ok {
 				log.Fatal("quorum needs to be bigger than 1/3 of committee size")
@@ -83,7 +82,7 @@ func initDeployCmd() *cobra.Command {
 			}
 
 			var govControllerAddr iotago.Address
-			if govControllerStr != "" {
+			if controllerAddressStr(govControllerStr) != "" {
 				var err error
 				var prefix iotago.NetworkPrefix
 				prefix, govControllerAddr, err = iotago.ParseBech32(govControllerStr)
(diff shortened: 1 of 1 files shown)


tools/wasp-cli/chain/deploy.go
```diff
@@ -83,7 +82,7 @@ func initDeployCmd() *cobra.Command {
 			}
 
 			var govControllerAddr iotago.Address
-			if govControllerStr != "" {
+			if controllerAddressStr(govControllerStr) != "" {
```
dessaya: controllerAddressStr() will never return an empty string? :thinking:

ivange94: It will return an empty string if `address.0` is not set.

dessaya: But wait... I don't see an "address" section in the config. What is it supposed to return?

To obtain your address just do `wallet.Load().Address()`. And that is guaranteed to return a valid address.

ivange94: I wasn't really sure which address was being talked about as I missed that part of the meeting on Friday but when the issue mentioned `address.0`, I presumed it's something you set with maybe `wasp-cli set address.0 {address}`. @fijter can you help clarify that requirement?

tools/wasp-cli/chain/deploy.go
```diff
@@ -59,17 +71,9 @@ func initDeployCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "deploy",
```
dessaya:
```suggestion
		Use:   "deploy [<alias>]",
```

tools/wasp-cli/chain/deploy.go
```diff
@@ -113,12 +106,18 @@ func initDeployCmd() *cobra.Command {
 			})
 			log.Check(err)
 
+			var alias string
+			if len(args) == 0 {
+				alias = GetChainAlias()
+			} else {
+				alias = args[0]
+			}
```
dessaya, line 118: Does it make sense to deploy a chain with a previously used alias? I think it would be better to make it a required argument.

Anyway, I think this change might break documentation and tests. Did you run the wasp-cli cluster tests?

tools/wasp-cli/chain/deploy.go
```diff
@@ -113,12 +106,18 @@ func initDeployCmd() *cobra.Command {
...
 			AddChainAlias(alias, chainid.String())
 		},
 	}
 
-	cmd.Flags().IntSliceVarP(&committee, "committee", "", nil, "peers acting as committee nodes (ex: 0,1,2,3) (default: all nodes)")
-	cmd.Flags().IntVarP(&quorum, "quorum", "", 0, "quorum (default: 3/4s of the number of committee nodes)")
+	cmd.Flags().IntSliceVarP(&committee, "committee", "", []int{0}, "peers acting as committee nodes (ex: 0,1,2,3) (default: all nodes)")
```
dessaya:
```suggestion
	cmd.Flags().IntSliceVarP(&committee, "committee", "", []int{0}, "peers acting as committee nodes (ex: 0,1,2,3) (default: 0)")
```

tools/wasp-cli/chain/deploy.go
```diff
@@ -113,12 +106,18 @@ func initDeployCmd() *cobra.Command {
...
+	cmd.Flags().IntVarP(&quorum, "quorum", "", 1, "quorum (default: 3/4s of the number of committee nodes)")
```
dessaya: This change is not correct: the default quorum must be calculated from the committee list; that's what `defaultQuorum()` did.

dessaya (changes requested):

tools/wasp-cli/chain/deploy.go
ivange94: So my design was to make sure you can deploy a chain with `wasp-cli chain deploy <alias>` but also have it working as it's always been. For example you should still be able to deploy with 
`wasp-cli chain deploy --chain=alias --committee=0 --quorum=1 --gov-controller=addr`

tools/wasp-cli/chain/deploy.go
ivange94: But if the default committee is always made up of 1 node(as per the requirement), what's the point of calculating the quorum? If someone is choosing to use the fault committee of `0`, I assume they'll also be working with quorum of 1 by default. Or is it possible to have `--committee=0` and a quorum other than `1`?

tools/wasp-cli/chain/deploy.go
ivange94: But if the default committee is always made up of 1 node(as per the requirement), what's the point of calculating the quorum? If someone is choosing to use the fault committee of 0, I assume they'll also be working with quorum of 1 by default. Or is it possible to have --committee=0 and a quorum other than 1?

dessaya: The requirement says that the default values should be committee=[0] and quorum=1.

But if the user specifies a different quorum, e.g. [0,1,2,3], then the default quorum should NOT be 1, since the minimum is 1/3 committee size. See `isEnoughQuorum()`

That is the current behavior, and I think it should be kept as is. Just call `defaultQuorum()` if the user does not specify the quorum.

dessaya (approved):

PR #1795, opened 2023-02-01
refactor: allowance assets

jorgemmsilva:
- removes fungible tokens / allowance structs - use `Assets` everywhere
- deprecate "sendAsNFT", now we have a single SEND function on the sandbox
- misc: avoid running tests in parallel so we don't get weird exceptions due to wasm not playing well with parallelism


packages/vm/core/evm/iscmagic/ISCPrivileged.sol
```diff
@@ -8,9 +8,23 @@ import "@iscmagic/ISCTypes.sol";
 // The ISC magic contract has some extra methods not included in the standard ISC interface:
 // (only callable from privileged contracts)
 interface ISCPrivileged {
-    function moveBetweenAccounts(address sender, address receiver, ISCAllowance memory allowance) external;
-    function addToAllowance(address from, address to, ISCAllowance memory allowance) external;
-    function moveAllowedFunds(address from, address to, ISCAllowance memory allowance) external;
+    function moveBetweenAccounts(
```
dessaya, line 11: Are you using an autoformatter? Can it be configured to allow longer lines? I feel like it is a bit overzealous :stuck_out_tongue:

dessaya (approved):

packages/vm/core/evm/iscmagic/ISCPrivileged.sol
jorgemmsilva: yeah, I just auto format everything :smile: . Think it uses `prettier` by default don't think its configurable without adding a config file to the repo.

should we add `solhint`, or `prettier` for solidity files? I rather have a tool doing all those decisions for me than manually formatting files :joy:

jorgemmsilva: but regarding this change in particular I do think multiple lines are easier to read :shrug:

dessaya: I have mixed feelings. It's a matter of taste, of course. But automatically breaking long lines IMO is not good because then you lose consistency: now to my eyes some lines look too long and others look too short. That makes it worse. I think it is more readable when all lines look the same. For example here:

```
    event Approval(
        address indexed tokenOwner,
        address indexed spender,
        uint256 tokens
    );
    event Transfer(address indexed from, address indexed to, uint256 tokens);
```

Also, autoformatters are most useful when everyone in the team use them. Otherwise you get random changes across commits, and `git blame` becomes less precise. Case in point the example above, where there is nothing related to the assets change.

Anyway, just ranting. Feel free to ignore :stuck_out_tongue:

PR #1801, opened 2023-02-01
fix: calculate storage deposit estimate after getting L1 params

dessaya:


jorgemmsilva (commented):
I've also noticed this is done too many times.
Not sure if the fix is right, tho. In theory the L1params could change during runtime, and this would need to be recalculated

muXxer:
I think we need to pass `L1().Protocol.RentStructure` to `NewStorageDepositEstimate()` and recalculate `MinRent` every time.
The used outputs for the estimation never change, right? So maybe we can reuse these. The calculation of MinRent shouldn't be too heavy I guess.
Or we check if the RentParameters changed since last execution, and recalculate in case they were modified?

dessaya:
Sounds good, I will change it so that only MinRent is called dynamically.

jorgemmsilva (approved):

PR #1840, opened 2023-02-07
perf(evm): improve binary search performance in EVM estimate gas

dessaya:
Before:

```
BenchmarkRPCEstimateGas   3494  3413750 ns/op  1695188 B/op  24989 allocs/op
```

After:

```
BenchmarkRPCEstimateGas  30376   392667 ns/op   151316 B/op   2305 allocs/op
```


packages/chainutil/estimategas_evm.go
```diff
@@ -75,46 +75,63 @@ func EstimateGas(ch chain.ChainCore, call ethereum.CallMsg) (uint64, error) {
...
+		return false, nil, res.Receipt.GasBurned
 	}
+
 	// Execute the binary search and hone in on an executable gas limit
+	var lastUsed uint64
 	for lo+1 < hi {
 		mid := (hi + lo) / 2
-		failed, err := executable(mid)
+		if lastUsed > lo && lastUsed != mid && lastUsed < hi {
+			// use the last used gas as a better estimation to home in faster
```
jorgemmsilva, line 108:
```suggestion
			// use the last used gas as a better estimation to hone in faster
```

packages/chainutil/estimategas_evm.go
```diff
@@ -75,46 +75,63 @@ func EstimateGas(ch chain.ChainCore, call ethereum.CallMsg) (uint64, error) {
 	gasCap = hi
 
 	// Create a helper to check if a gas allowance results in an executable transaction
-	executable := func(gas uint64) (failed bool, err error) {
+	executable := func(gas uint64) (failed bool, err error, used uint64) {
```
jorgemmsilva: isn't it standard style in go to return the error as the last value?

packages/chainutil/estimategas_evm.go
```diff
@@ -75,46 +75,63 @@ func EstimateGas(ch chain.ChainCore, call ethereum.CallMsg) (uint64, error) {
...
 			return 0, err
 		}
 		if failed {
 			lo = mid
 		} else {
 			hi = mid
+			if lastUsed == mid {
+				// if used gas == gas limit, then use this as the estimation.
+				// It may not be the most precise estimation (e.g. lowering the gas
+				// limit may end up using less gas), but it's "good enough" and
```
jorgemmsilva, line 125: >lowering the gas limit may end up using less gas)

this shouldn't be the case :thinking:. Unless the execution is tied to the gas limit (contract uses the gas limit value), it shouldn't use less gas if a lower value is provided, it should fail.

packages/chainutil/estimategas_evm.go
```diff
@@ -75,46 +75,63 @@ func EstimateGas(ch chain.ChainCore, call ethereum.CallMsg) (uint64, error) {
...
+		var err error
+		failed, err, lastUsed = executable(mid)
 		if err != nil {
 			return 0, err
 		}
 		if failed {
 			lo = mid
 		} else {
 			hi = mid
+			if lastUsed == mid {
```
jorgemmsilva, line 122: this will only work for the cases where "exactimation" is acceptable (gas used = request gas limit).

for calls that need a gas limit higher than the gas that will be charged, this will basically be equivalent to the existing solution (which is good, it won't fail).

However, I'm not 100% sure if its a good idea to optimize for "exactimations" like this, it means the request will fail if an extra operation is executed for some reason (say 1 object is added to an array that is iterated in the contract execution between the time the estimation is made and request is executed on the real network, the request would fail by 1 gas unit or something like that).

maybe its okay to provide the exactimation + some extra "safeguard gas" like 1000 units or something.

What do you think?

dessaya: Do we have a test for that border case? Note that this is not just returning the gas used. I just tried doing that, and one of our tests fails.

Instead, it is running the tx with some amount of gas, taking note of the gas used, then running it **again** with the gas used previously, and only if the gas used stays the same, then it breaks the loop. This way at least all our tests pass.

packages/chainutil/estimategas_evm.go
dessaya: There is a `gasleft()` function that in theory could be used to produce this effect: use only the amount of gas provided. Although I didn't test it, maybe I should add a test for that.

packages/chainutil/estimategas_evm.go
dessaya: Haha yes, don't know how that one escaped me :man_facepalming:

packages/chainutil/estimategas_evm.go
dessaya: Actually I think "home in" makes more sense here: https://prowritingaid.com/home-in-vs-home-in

jorgemmsilva: Ahah, TIL about "home in"

Still, a bit ambiguous:
>Hone in means to advance toward a goal

😛

packages/chainutil/estimategas_evm.go
jorgemmsilva: > it is running the tx with some amount of gas, taking note of the gas used, then running it again

you're right! :thinking: This solves that issue


I remember users were having problems with exactimations so changed this to use the binary search method, after that there were no more reports of bad gas estimations. 
I didn't specifically create a test to reproduce this edge case, but it's a good idea. I believe it can be reproduced by using the `new` keyword:
https://trufflesuite.com/blog/ethereum-gas-exactimation/#setup-an-example-smart-contract

Its also possible that our test suite already includes a case like this, that's why it failed initially :smile:

packages/chainutil/estimategas_evm.go
jorgemmsilva: sorry, I don't follow :sweat_smile:

dessaya:
```
    event LoopEvent();

    function loopWithGasLeft() public {
        while (gasleft() >= 10000) {
            emit LoopEvent();
        }
    }
```

This one uses as much gas as you give it. So, calling `estimateGas` could give any value. unless we used the "proper" binary search, which would return the smallest value that succeeds.

jorgemmsilva: yeah, that's what I meant by:

>Unless the execution is tied to the gas limit

but that shouldn't be 99.999% of the cases :smile:

jorgemmsilva (approved):

Issue #1846, opened 2023-02-09
linter: remove exclusion of :shadow: declaration of "err"


jorgemmsilva:
right now this error is being ignored:
```
shadow: declaration of "err"
```

this is bad, we should remove this "exclude rule", and fix the issues that will pop-up

Issue #1845, opened 2023-02-09
Linter task (make link) does not execute on wasp-cli


jorgemmsilva:
easy to reproduce: 
 - create a linting issue in wasp-cli code
 - run `make lint`, notice the wasp-cli lint error is not reported

its not trivial to resolve, adding `./tools/wasp-cli` in the lint task will result in:
```
[linters_context] typechecking error: main module (github.com/iotaledger/wasp) does not contain package github.com/iotaledger/wasp/tools/wasp-cli 

```

PR #1848, opened 2023-02-09
chore(lint): lint task not running on wasp-cli, remove bad lint exclude

jorgemmsilva:
fixes:
https://github.com/iotaledger/wasp/issues/1846
https://github.com/iotaledger/wasp/issues/1845


dessaya (approved):

muXxer (approved):
Mhh not sure I like the "err2" thingy because its not really shadowing the other "err" variable if it is only in the scope of the if function, but yeah. Lets merge it :trollface:

PR #1855, opened 2023-02-09
feat(evm): ERC20 contract for off-chain native tokens

### LINKED ISSUE #1837: Add support for registering Foundries not created by the chain (DX improvement for native asset support in EVM)

dessaya:
closes #1837


documentation/docs/guide/core_concepts/core_contracts/evm.md
```diff
@@ -72,6 +72,71 @@ Only the foundry owner can call this endpoint.
...
+- `t` (`string`): The ticker symbol
+- `d` (`uint8`): The token decimals
+- `A` (`uint8`): The target chain address, where the ERC20 contract will be
+  registered.
+
+### `registerERC20ExternalNativeToken`
+
+Registers an ERC20 contract to act as a proxy for the native tokens.
+
+Only an alias address can call this endpoint. End users should call
```
fijter: How does this work from an end-user perspective? Do we need to ask for additional functionality and examples to the core client libs to support this? A common question would be "I have this token on L1 I created, how do I get it on a chain as ERC20?", do you have a summary of the steps this user should take to get there?

documentation/docs/guide/core_concepts/core_contracts/evm.md
```diff
@@ -72,6 +72,71 @@ Only the foundry owner can call this endpoint.
 - `t` (`string`): The ticker symbol
 - `d` (`uint8`): The token decimals
 
+### `registerERC20NativeTokenOnChain`
+
+Registers an ERC20 contract to act as a proxy for the native tokens **on another
+chain**.
+
+The foundry must be controlled by the chain. Only the foundry owner can call
```
fijter: Is there a way to transfer the ownership of a foundry easily for L1 to a SC chain? Have we ever tried this?

documentation/docs/guide/core_concepts/core_contracts/evm.md
```diff
@@ -72,6 +72,71 @@ Only the foundry owner can call this endpoint.
...
+this endpoint.
+
+After a few minutes, call
+[`getERC20ExternalNativeTokensAddress`](#geterc20externalnativetokensaddress)
+on the **target chain** to find out the address of the ERC20 contract.
+
+#### Parameters
+
+- `fs` (`uint32`): The foundry serial number
```
fijter, line 93: Should this be Foundry ID instead of serial number? I thought the serial number was just a integer in combination with the controlling alias per the TIP: https://github.com/iotaledger/tips/blob/main/tips/TIP-0018/tip-0018.md#foundry-output

documentation/docs/guide/core_concepts/core_contracts/evm.md
```diff
@@ -72,6 +72,71 @@ Only the foundry owner can call this endpoint.
...
+
+Registers an ERC20 contract to act as a proxy for the native tokens.
+
+Only an alias address can call this endpoint. End users should call
+[`registerERC20NativeTokenOnChain`](#registererc20nativetokenonchain)
+on the chain that controls the foundry.
+
+#### Parameters
+
+- `fs` (`uint32`): The foundry serial number
```
fijter, line 113: Same question as above :)

fijter (commented):
Added some remarks regarding how this works, no in-depth code analysis

documentation/docs/guide/core_concepts/core_contracts/evm.md
dessaya: No idea. But even if it's possible:
1. we would have to implement that feature in the VM
2. it feels "wrong" to have to transfer ownership of the foundry just to register it on another chain

documentation/docs/guide/core_concepts/core_contracts/evm.md
dessaya: The foundry ID is calculated as: `alias address + serial number + token scheme type`

The `alias address` is known; it is the same as the chain ID of the chain that receives this call.

The `token scheme type` is assumed to be `SimpleTokenScheme`.

So, the only unknown bit is the `serial number`. That's why it is received as a parameter. We could receive the Fondry ID but it would be redundant.

documentation/docs/guide/core_concepts/core_contracts/evm.md
dessaya: Firstly, on L1 a foundry can only be controlled by an alias address. The only implementation of an alias address I know is an ISC Chain. So, the only way an end-user has to create a new fondry is by "asking" an ISC chain to do so.

So, let's say we have two chains, A and B.
1. Alice creates a new foundry on chain A, and wants it to be available on chain B as ERC20.
2. Alice then calls `registerERC20NativeTokenOnChain` on chain A with `targetChain = B`
3. Chain A calls `registerERC20ExternalNativeToken` on chain B (this is probably our first real application of an inter-chain call?). This is an async operation since it is an on-ledger request.
4. Alice must wait a few minutes and then call `getERC20ExternalNativeTokensAddress` on chain B, in order to find out the new ERC20 address.

I don't know of another way to create a foundry. If there is, then probably step 1 can be replaced with that.

documentation/docs/guide/core_concepts/core_contracts/evm.md
alexsporn: that does not work, the chain would need to support the control of other aliases

fijter: > No idea. But even if it's possible:
> 
> 1. we would have to implement that feature in the VM
> 2. it feels "wrong" to have to transfer ownership of the foundry just to register it on another chain

I agree it's wrong to do that for just this feature, it might however be useful at one point in time to transfer management including minting and burning permissions of a native asset to a chain. Not a high priority though and we likely should just accept that as the limitation of non-chain managed native assets and instead have people migrate to chain managed tokens instead if that occurs.

documentation/docs/guide/core_concepts/core_contracts/evm.md
```diff
@@ -72,6 +72,78 @@ Only the foundry owner can call this endpoint.
 - `t` (`string`): The ticker symbol
 - `d` (`uint8`): The token decimals
 
+### `registerERC20NativeTokenOnRemoteChain`
+
+Registers an ERC20 contract to act as a proxy for the native tokens **on another
+chain**.
```
jorgemmsilva, line 78:
```suggestion
Registers an ERC20 contract to act as a proxy for the native tokens **created by another
chain**.
```

jorgemmsilva (approved):

documentation/docs/guide/core_concepts/core_contracts/evm.md
dessaya: No, the foundry is created by this chain, and the purpose of this endpoint is to register the token on some other chain. This is explained in the paragraph below.

I agree it is confusing though. Accepting ideas to make it clearer.

PR #1868, opened 2023-02-11
Missing build version makes wasp to fail on startup in local setup

ivange94:
# Description of change

Wasp panics in local setup due to missing version during build


ivange94:
This is a user error. Looks like the before running `docker compose up`, you need to run one of the build scripts depending on your platform version. So running `build_container.sh` before `docker compose up` fixed the issue for me. I'll update the documentation in a different PR.

PR #1872, opened 2023-02-11
Update local setup docs

ivange94:
# Description of change
Update local setup docs

### DIFF
--- a/tools/local-setup/README.md
+++ b/tools/local-setup/README.md
@@ -1,26 +1,34 @@
 # Local Setup
 
-This Directory contains a Docker-based setup to run your own Wasp development setup.
+This Directory contains a Docker-based setup to run your own Wasp development
+setup.
 
 ## Usage
 
 ### Starting
 
+**Note** You need to build the wasp container using `build_container.sh` if
+you're on Mac/Linux or `build_container.cmd` on windows so that the `wasp`
+version is properly set.
+
 Run `docker-compose up` to start the setup.
 
 ### Stopping
 
-Press `Ctrl-C` to shut down the setup, but don't press it twice to force it. Otherwise, you can corrupt the Hornet database.
+Press `Ctrl-C` to shut down the setup, but don't press it twice to force it.
+Otherwise, you can corrupt the Hornet database.
 
 You can also shut down the setup with `docker-compose down` in a new terminal.
 
 ### Reset
 
-Run `docker-compose down --volumes` to shut down the nodes and to remove all databases.
+Run `docker-compose down --volumes` to shut down the nodes and to remove all
+databases.
 
 ### Recreation
 
-If you made changes to the Wasp code and want to use it inside the setup, you need to recreate the Wasp image.
+If you made changes to the Wasp code and want to use it inside the setup, you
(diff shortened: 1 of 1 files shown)


tools/local-setup/README.md
```diff
@@ -1,26 +1,34 @@
...
+setup.
 
 ## Usage
 
 ### Starting
 
+**Note** You need to build the wasp container using `build_container.sh` if
+you're on Mac/Linux or `build_container.cmd` on windows so that the `wasp`
+version is properly set.
+
```
jorgemmsilva, line 13: this is not true. the local setup should use an image from dockerhub.

tools/local-setup/README.md
```diff
@@ -44,7 +52,8 @@ To configure a new wasp-cli you can use the following commands:
 
 :::note
 
-You can either use a wasp-cli installed on your system, or use the one built-in to the wasp docker container by doing: `docker exec  wasp /app/wasp-cli init`
+You can either use a wasp-cli installed on your system, or use the one built-in
+to the wasp docker container by doing: `docker exec  wasp /app/wasp-cli init`
 
 :::
```
jorgemmsilva, line 59: we should remove this since we took wasp-cli out of the main image.
Rather there should be a link to down the latest release from github

jorgemmsilva (changes requested):

tools/local-setup/README.md
ivange94: Then you might want to check what's on dockerhub. I can reproduce this error every time if I clean up all images and containers with `docker system prune -a` and just do a `docker compose up` without building first and this error pops up. Here's the log
```
   self.warnings.warn(
local-setup-hornet-nest-1  | 2023-02-11 02:36:59,238 CRIT Server 'unix_http_server' running without any HTTP authentication checking
wasp                       | panic: unable to initialize app: no version given
wasp                       |
wasp                       | goroutine 1 [running]:
wasp                       | github.com/iotaledger/hive.go/core/app.New({0x2dad45d, 0x4}, {0x0, 0x0}, {0xc000c43e78, 0x4, 0x1a0?})
wasp                       |    /go/pkg/mod/github.com/iotaledger/hive.go/core@v1.0.0-rc.2.0.20230120131747-b65c9cb22ea9/app/app.go:72 +0x618
wasp                       | github.com/iotaledger/wasp/core/app.App()
wasp                       |    /scratch/core/app/app.go:38 +0x2f7
wasp                       | main.main()
wasp                       |    /scratch/main.go:8 +0x19
wasp exited with code 2
wasp                       | panic: unable to initialize app: no version given
```
But I will close this as a user error just incase the problem is my system.

jorgemmsilva: I'm not saying the image is working :trollface: 

Just saying:
>You need to build the wasp container using...

is not correct.
We want this to be plug-and-play, no need for the users to rebuild wasp image.

we should fix the dockerhub image if that is broken.

ivange94:
This problem no longer exists. I just did a `docker system prune -a` and then `docker compose up` and the version is set properly so closing this now.

PR #1891, opened 2023-02-14
Accept pub keys apart from names when specifying trusted keys.

kape1395:
That's done for the DKG as well as chain access node management.

Also:
  - the node adds self to trusted peers on startup.
  - stacktrace is printed on error in the api.

### DIFF
--- a/packages/peering/peering.go
+++ b/packages/peering/peering.go
@@ -11,10 +11,13 @@ package peering
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/samber/lo"
 
 	"github.com/iotaledger/wasp/packages/cryptolib"
 )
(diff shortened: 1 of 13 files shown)


packages/peering/peering.go
```diff
@@ -59,12 +62,58 @@ type TrustedNetworkManager interface {
...
+			peer, ok := lo.Find(trustedPeers, func(p *TrustedPeer) bool {
+				return p.Name == pubKeyOrName
+			})
+			if !ok {
+				return nil, fmt.Errorf("cannot find trusted peer by name=%v", pubKeyOrName)
+			}
+			result[i] = peer
+		}
+	}
+	return result, nil
```
jorgemmsilva, line 114: could be simplified by using lo.filter

jorgemmsilva (approved):
:ok_hand:

packages/peering/peering.go
kape1395: The `lo.Filter()` would allow you to return an incomplete set. E.g., you asked 3 peers, but DB has 1. The result would be 0 or 1 peers, but not 3, if you would filter the DB. Therefore I do mapping over the nodes in the query. Don't look like lo.map would work much better.

jorgemmsilva: after filtering you can just assert that the length matches the expected and return an error otherwise. But sure, this way you can exit earlier and tell in the error the first peer that is missing.

PR #1892, opened 2023-02-14
Update l2 api

lmoe:
# Description of change

* Updates the EVM toolkit
* Adds dockerized apiclient generator and updates Makefile `make apiclient-docker` 
* Restricts deletion of the last user
* Sanitizes users permissions
* Shows own node in node list as always trusted


packages/users/user_manager.go
```diff
@@ -25,6 +30,22 @@ func NewUserManager(storeCallback func([]*User) error) *UserManager {
...
+	sanitizedPermissions := map[string]struct{}{}
+
+	for _, permission := range maps.Keys(permissions) {
+		if isPermissionAllowed(permission) {
+			sanitizedPermissions[permission] = struct{}{}
+		}
+	}
+
+	return sanitizedPermissions
+}
```
jorgemmsilva, line 47: I'm a bit confused here, why is this a `map[string]struct{}`, and why does the sanitized version is just  map of strings to empty stucts?

packages/webapi/interfaces/interfaces.go
```diff
@@ -30,6 +30,7 @@ var (
 	ErrNoBalanceOnAccount     = errors.New("no balance on account")
 	ErrInvalidNonce           = errors.New("invalid nonce")
 	ErrChainNotFound          = errors.New("chain not found")
+	ErrCantDeleteLastUser     = errors.New("you can't delete the last user")
```
jorgemmsilva, line 33: this shouldn't be needed, since a user can't delete himself, right?

jorgemmsilva (approved):

packages/webapi/interfaces/interfaces.go
lmoe: True, unless the authentication scheme is set to `none` and there is no "user" to be checked against.

packages/users/user_manager.go
lmoe: That's how the user manager was built. The emptry struct could also have been a boolean but that was not my decision. 

I just sanitize the map properly so you can't add some bogus permissions to the user. :)

PR #1893, opened 2023-02-15
Recovery should be counted after mempool proposal is received.

kape1395:


jorgemmsilva (approved):
there is a linting error on an unrelated file :shrug:

kape1395:
@jorgemmsilva , added a useless change to make the linter happy.

PR #1897, opened 2023-02-15
fix: graceful shutdown when INX connection breaks

jorgemmsilva:


tools/cluster/tests/env.go
```diff
@@ -18,13 +18,15 @@ import (
 
 // TODO remove this?
 func setupWithNoChain(t *testing.T, opt ...waspClusterOpts) *ChainEnv {
-	return &ChainEnv{t: t, Clu: newCluster(t, opt...)}
+	clu, _ := newCluster(t, opt...)
+	return &ChainEnv{t: t, Clu: clu}
```
muXxer, line 22: any reason why you didn't pass the dataPath here, but did in another changed code part?

clients/multiclient/node.go
```diff
@@ -0,0 +1,18 @@
+package multiclient
+
+import (
+	"context"
+
+	"github.com/iotaledger/wasp/clients/apiclient"
+)
+
+// ActivateChain sends a request to activate a chain in all wasp nodes
```
muXxer: Copy paste error :trollface:

packages/nodeconn/nodeconn.go
```diff
@@ -707,7 +726,8 @@ func (nc *nodeConnection) AttachChain(
 	}()
 
 	if err := chain.SyncChainStateWithL1(ctx); err != nil {
-		nc.LogPanicf("synchronizing chain state %s failed: %s", chainID, err.Error())
+		nc.LogError("synchronizing chain state %s failed: %s", chainID, err.Error())
+		nc.shutdownHandler.SelfShutdown(fmt.Sprintf("Cannot sync chain %s with L1", chain.chainID), true)
```
muXxer: Maybe also add the error message here?

tools/cluster/tests/env.go
jorgemmsilva: so it is available in `chainEnv`.
a lot of tests just start with `setupNativeInccounterTest` that return a `chainEnv`, this makes the chainEnv aware of the data path

muXxer (approved):

PR #1916, opened 2023-02-15
feat(api): add endpoint for getAccountFoundries

dessaya:


tools/cluster/tests/wasp-cli_test.go
```diff
@@ -454,14 +454,13 @@ func TestWaspCLILongParam(t *testing.T) {
...
-		"chain", "post-request", "accounts", accounts.FuncFoundryCreateNew.Name,
-		"string", accounts.ParamTokenScheme, "bytes", "0x00d107000000000000000000000000000000000000000000000000000000000000d207000000000000000000000000000000000000000000000000000000000000d30700000000000000000000000000000000000000000000000000000000000000", "-l", "base:1000000",
-		"-t", "base:1000000",
-		"--node=0",
-	)
+	w.CreateL2Foundry(&iotago.SimpleTokenScheme{
+		MaximumSupply: big.NewInt(1000000),
+		MeltedTokens:  big.NewInt(0),
+		MintedTokens:  big.NewInt(0),
+	})
```
jorgemmsilva, line 463: doesn't this remove the existing test? `TestWaspCLILongParam` feels like that long value was important to prevent a regression.

packages/webapi/corecontracts/accounts.go
```diff
@@ -91,6 +92,24 @@ func (a *Accounts) GetAccountNFTs(chainID isc.ChainID, agentID isc.AgentID) ([]i
...
+	}
+	sns := make([]uint32, 0, len(ret))
+	for k := range ret {
+		sn, err := codec.DecodeUint32([]byte(k))
+		if err != nil {
+			return nil, err
+		}
+		sns = append(sns, sn)
+	}
+	return sns, nil
```
jorgemmsilva, line 110: it is still impossible to know what is the newly created foundry SN, right?
if an account has 10 foundries, and after creation 11, it needs to figure out which is the new one

jorgemmsilva (approved):

tools/cluster/tests/wasp-cli_test.go
dessaya: This is just cleaning up the code so that the call to `foundryCreateNew` is more readable. The "long param" bit is right below.

packages/webapi/corecontracts/accounts.go
dessaya: That's right, but at least you can call this function before and after and do a diff? :grin:

PR #1918, opened 2023-02-15
Fix invalid authentication, update documentation, refactor wasp-cli

lmoe:
# Description of change
* Fixes public API appearing as protected
* Change `Distrust Peer` from body request to param only (invalid HTTP standard) 
* Put all param path keys and descriptions into a separate file to ensure consistency between route descriptions
* Adjust param path descriptions types
* Refactor wasp-cli distrust peer command

## Type of change

- Documentation Fix

### DIFF
--- a/packages/webapi/params/keys.go
+++ b/packages/webapi/params/keys.go
@@ -1,0 +1,1 @@
+package params
--- a/packages/webapi/params/keys.go
+++ b/packages/webapi/params/keys.go
@@ -18,7 +18,7 @@ const (
 )
 
 const (
-	DescriptionAgentID       = "AgentID (Bech32 for WasmVM | Hex for EVM | '000000@Bech32' Addresses require urlencode)"
+	DescriptionAgentID       = "AgentID (Bech32 for WasmVM | Hex for EVM)"
 	DescriptionBlobHash      = "BlobHash (Hex)"
 	DescriptionChainID       = "ChainID (Bech32)"
 	DescriptionContractHName = "The contract hname (Hex)"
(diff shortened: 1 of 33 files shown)


jorgemmsilva (approved):

packages/webapi/params/keys.go
```diff
@@ -0,0 +1,33 @@
...
+	ParamPublicKey     = "publicKey"
+	ParamRequestID     = "requestID"
+	ParamSharedAddress = "sharedAddress"
+	ParamStateKey      = "stateKey"
+	ParamTxHash        = "txHash"
+	ParamUsername      = "username"
+)
+
+const (
+	DescriptionAgentID       = "AgentID (Bech32 for WasmVM | Hex for EVM | '000000@Bech32' Addresses require urlencode)"
```
dessaya:
```suggestion
	DescriptionAgentID       = "AgentID (Bech32 for L1 | Hex for EVM | 'hname@Bech32' for contracts)"
```

dessaya: I didn't understand the urlencode bit?

packages/webapi/params/keys.go
```diff
@@ -0,0 +1,33 @@
...
+	DescriptionBlobHash      = "BlobHash (Hex)"
+	DescriptionChainID       = "ChainID (Bech32)"
+	DescriptionContractHName = "The contract hname (Hex)"
+	DescriptionFieldKey      = "FieldKey (String)"
```
dessaya, line 25:
```suggestion
	DescriptionFieldKey      = "Key (String)"
```

packages/webapi/params/keys.go
lmoe: The AgentID is used in GET urls. The problem is that you can't call http://api/chain/accounts/000000@bech32Address without doing a urlencode. The browser should do it automatically, but some rest clients do not. The echo server will do weird stuff with it otherwise and the api will parse a wrong address.

dessaya: AFAIK the `@` is valid in an URL and does not require encoding: https://stackoverflow.com/questions/7109143/what-characters-are-valid-in-a-url

lmoe: So I've tried to figure out why exactly I have put that in. 
I definitely had the issue that the AgentID could not be decoded properly, and then I have added the urldecode function which made it work.  

Now I can't reproduce it. Removing the urldecode works fine too. Maybe it was something else.

I have removed it for now.
If some things don't work anymore, we should revisit I guess. :D

PR #1919, opened 2023-02-15
fix: Hname calculation edge case

### LINKED ISSUE #1576: [audit] edge case in hname computation
In ISC each smart contract is referenced by an 'hname' which is a 32-bit value computed from the contract name. It is mentioned that the hname cannot be 0 or ^0. This is implemented in the code https://github.com/iotaledger/wasp/blob/dd1954aed902a0ef3706969ea89b707684efecd8/packages/isc/hname.go#L44 by checking if the first 4 bytes of the Blake2b hash are 0 or ^0 if so bytes 4-8 are returned. Since bytes 4-8 can also be 0 or ^0 this could still return 0 or ^0. Note that it would require substantive effort to compute a string for which the Blake2b hash starts with 8 0 or ^0 bytes but this is not infeasible given hardware is available that can compute multiple TH/s of Blake2b computations. A potential fix would be to return a constant such as '1' when bytes 4-8 are also 0 or ^0.

dessaya:
closes #1576

### DIFF
--- a/packages/isc/hname.go
+++ b/packages/isc/hname.go
@@ -15,8 +15,8 @@ import (
 	"github.com/iotaledger/wasp/packages/hashing"
 )
 
-// Hname is 4 bytes of blake2b hash of any string interpreted as little-endian uint32.
-// 0 and not ^0 are reserved values and the isc.Hn ensures it is not returned
+// Hname is calculated as the first 4 bytes of the blake2b hash of a string, interpreted as
+// a little-endian uint32.
 type Hname uint32
 
 const HnameLength = 4
@@ -29,6 +29,9 @@ var (
 	EntryPointInit = Hn(FuncInit)
 )
 
+// HnameNil is the value used to represent a non-existent Hname.
+const HnameNil = Hname(0)
+
 // HnameFromBytes constructor, unmarshalling
 func HnameFromMarshalUtil(mu *marshalutil.MarshalUtil) (ret Hname, err error) {
 	err = ret.ReadFromMarshalUtil(mu)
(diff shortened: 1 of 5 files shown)


packages/isc/hname.go
```diff
@@ -40,15 +43,24 @@ func HnameFromBytes(data []byte) (ret Hname, err error) {
...
-// Hn create hname from arbitrary string.
+// Hn calculates the hname for the given string.
+// For any given string s, it is guaranteed that Hn(s) != HnaneNil.
 func Hn(name string) (ret Hname) {
-	h := hashing.HashStrings(name)
-	_ = ret.Read(bytes.NewReader(h[:HnameLength]))
-	if ret == 0 || ret == Hname(^uint32(0)) {
-		// ensure 0 and ^0 are impossible
-		_ = ret.Read(bytes.NewReader(h[HnameLength : 2*HnameLength]))
+	for i := byte(0); i < 10; i++ {
```
BugFreeSoftware: Question: why rehashing? Why not simply take the next bytes of h and only when h's bytes are exhausted return 1?

dessaya: I figured it would be harder to exploit this way? But yes, I can change it, I have no real preference.

BugFreeSoftware: I guess it does not matter much, since the zero hash should occur only about 1 in 4 billion times. So rehashing would only happen that infrequent. But anyway, I don't see how anyone could exploit this. hNames identify contracts and functions. Anyone can call them with the actual public names. Forging a name to get to a hash to call a certain function seems overkill in that regard. You can simply call that function, You won't get any better access to the function or anything.
I have a slight preference for just getting the next bytes from the original hash, because that means less overhead on each call in the Rust version and a simpler loop. Due to ownership rules Rust will have to create a mutable copy of the input string first to be able to append the asterisk later, unless I unroll the loop once.

BugFreeSoftware (approved):

packages/isc/hname.go
```diff
@@ -40,15 +43,24 @@ func HnameFromBytes(data []byte) (ret Hname, err error) {
...
+		_ = ret.Read(bytes.NewReader(h[:HnameLength]))
+		if ret != HnameNil {
+			return ret
+		}
+		// add something to the end of the string to generate a different hash
+		name += "*"
 	}
-	return ret
+	// astronomically unlikely to end up here after 10 attempts
+	return 1
```
jorgemmsilva: should it panic here?

jorgemmsilva (approved):

packages/isc/hname.go
dessaya: I'll change it then, if it makes the Rust implementation easier :+1:

packages/isc/hname.go
dessaya: IMO no, because that could hang the node. It is better to just return a constant value as the hname.

PR #1936, opened 2023-02-16
fix(evm): Remove the EVM blockGasLimit parameter

dessaya:
The EVM blockGasLimit parameter was being ignored anyway. What's more, when estimating gas, the ISC block gas limit was being used instead of the MaxGasPerCall parameter. This ensures that:

* The EVM block gas limit is always equal to ISC MaxGasPerBlock (after ratio conversion).

* The maximum EVM tx gas limit is equal to ISC MaxGasPerRequest (after ratio conversion).

Also a few more improvements to the gas estimation algorithm.

### DIFF
--- a/documentation/docs/guide/core_concepts/core_contracts/evm.md
+++ b/documentation/docs/guide/core_concepts/core_contracts/evm.md
@@ -43,7 +43,6 @@ Some parameters of the `evm` contract can be specified by passing them to the
 [`root` contract `init` entry point](root.md#init):
 
 - `evmg` (optional [`GenesisAlloc`](#genesisalloc)): The genesis allocation. The balance of all accounts must be 0.
-- `evmgl` (optional `uint64` - default: 15000000): The EVM block gas limit (EVM gas units)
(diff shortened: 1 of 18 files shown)


documentation/docs/guide/core_concepts/core_contracts/governance.md
```diff
@@ -189,6 +189,16 @@ Changes the ISC : EVM gas ratio.
...
+### `setEVMBlockGasLimit`
+
+Sets the EVM block gas limit.
+
+#### Parameters
+
+- `l`: (`uint64`).
+
+It can only be invoked by the chain owner.
+
```
jorgemmsilva: I'm a bit confused, do these exist?

I don't see them added in this PR and they don't exist on `develop` :thinking:

jorgemmsilva (approved):

documentation/docs/guide/core_concepts/core_contracts/governance.md
dessaya: Nice catch... I added those and then decided to remove them :facepalm:

PR #1967, opened 2023-02-19
Change WAL parameter name and default value

muXxer:


packages/chains/chains.go
```diff
@@ -51,7 +51,7 @@ type Chains struct {
 	trustedNetworkListenerCancel     context.CancelFunc
 	chainStateStoreProvider          database.ChainStateKVStoreProvider
 	walEnabled                       bool
-	walDir                           string
+	walFolderPath                    string
```
kape1395, line 54: Why not to call it `walPath` to match the naming in other places?

muXxer: I can do that as well.
I just thought I name it `folderPath` like we did in the DKSharesRegistry for example.

kape1395 (approved):

PR #1970, opened 2023-02-19
LocalView redesigned, logging improvements, cleanup of consensus instances.

kape1395:


packages/chain/cmtLog/var_localView.go
```diff
@@ -28,123 +78,254 @@ type VarLocalView interface {
...
+	pendingSICount := uint32(lvi.pending.Size())
+	for i := uint32(0); i < pendingSICount; i++ {
+		entries, ok := lvi.pending.Get(confirmedSI + i + 1)
+		if !ok {
+			return nil // That's a gap.
+		}
+		if len(entries) != 1 {
+			return nil // Alternatives exist.
+		}
+		if entries[0].rejected {
```
muXxer, line 299: Do we need to loop over all entries here just to be safe that none of them is rejected? Or is this already clear if we only check the first entry?

packages/chain/cmtLog/var_localView.go
```diff
@@ -9,16 +9,66 @@
...
+//     be rejected eventually as well, because they use the rejected AO as an input. On the
+//     other hand, it is unclear if unconfirmed AOs before the rejected one will be confirmed
+//     or rejected, we will wait until L1 decides on all of them.
+//
+//   - If we get a confirmed AO, that is not one of the AOs we have posted (and still waiting
+//     for a decision), then someone from the outside of a committee transitioned the chain.
+//     In this case all our produced/pending transactions are not meaningful anymore and we
+//     have to start building all the chain from the newly received AO.
+//
+//   - Recovery notice is received from a consensus (with SI/LI...) a new new consensus will
```
muXxer:
```suggestion
//   - Recovery notice is received from a consensus (with SI/LI...) a new consensus will
```

packages/chain/cmtLog/var_localView.go
kape1395: The check above tests if there is only a single element in the entries list. So it is enough here to check that single element.

muXxer: hahaha true :D sorry my bad.

muXxer (approved):

PR #1972, opened 2023-02-20
Wasmclient update to websockets

BugFreeSoftware:


muXxer:
Can we split this in two PRs?
I really would like to review the golang stuff for the added websocket, but I have no clue about the rust code.

lmoe:
Yeah I will need to change a few things before I would like to merge the websockets. Especially the command handling should be optimizied. I should be able to finish that today, and then we can merge yours.

PR #1973, opened 2023-02-20
wasp-cli convinience method for registerERC20NativeToken

ivange94:
# Description of change

wasp-cli convinience method for registerERC20NativeToken


tools/wasp-cli/corecontractscmd/cmd.go
```diff
@@ -0,0 +1,30 @@
...
+func Init(rootCmd *cobra.Command) {
+	callCmd := initCallCmd()
+	rootCmd.AddCommand(callCmd)
+
+	callCmd.AddCommand(initEvmCmd())
+}
+
+func initCallCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "call <contract>",
```
dessaya: This `chain call` command is a bit wierd considering that most other subcommands under `chain` are already calling core contracts. I think instead of:

```
wasp-cli chain call evm register-erc20-native-token
```

it would be simpler and more consistent like:

```
wasp-cli chain register-erc20-native-token
```

tools/wasp-cli/corecontractscmd/evm.go
```diff
@@ -0,0 +1,58 @@
...
+	cmd := &cobra.Command{
+		Use:   "register-erc20-native-token",
+		Short: "Call evm core contract registerERC20NativeToken entry point",
+		Args:  cobra.NoArgs,
+		Run: func(cmd *cobra.Command, args []string) {
+			n := waspcmd.DefaultWaspNodeFallback(node)
+			c := chaincmd.DefaultChainFallback(chain)
+
+			params := chainclient.PostRequestParams{}
+			chaincmd.PostRequest(n, c, evm.Contract.Hname().String(), "registerERC20NativeToken", params, true, adjustStorageDeposit)
```
dessaya: Maybe use the api client like other commands, for consistency?

dessaya (approved):

tools/wasp-cli/corecontractscmd/evm.go
ivange94: There's no API for full entry point calls just view calls. I was going to add that then I came across the `postRequest` command and thought maybe it'll be best to reuse it. I can add the api endpoint for this and refactor this command in another PR.

tools/wasp-cli/chain/register-erc20-native-token.go
```diff
@@ -0,0 +1,46 @@
...
+	cmd := &cobra.Command{
+		Use:   "register-erc20-native-token",
+		Short: "Call evm core contract registerERC20Nativetoken entry point",
+		Args:  cobra.NoArgs,
+		Run: func(cmd *cobra.Command, args []string) {
+			node = waspcmd.DefaultWaspNodeFallback(node)
+			chain = defaultChainFallback(chain)
+
+			params := chainclient.PostRequestParams{}
+			postRequest(node, chain, evm.Contract.Hname().String(), "registerERC20NativeToken", params, true, adjustStorageDeposit)
```
muXxer: How are parameters like `foundrySerialNumber`, `tokenName`, `tickerSymbol`, `tokenDecimals` and `allowance` passed to the actual command? I don't get it :innocent:

muXxer (changes requested):

tools/wasp-cli/chain/register-erc20-native-token.go
ivange94: Nice catch. That should be fixed now.

muXxer (approved):

PR #1976, opened 2023-02-20
Fix node connection on startup if L1 node is not synced

muXxer:
We had to move parts of the init logic to the `Run` function of the node connection.

Otherwise the node connection would loop forever if L1 was not sync on startup, because the stream to update the L1 node status was not started yet, and it was only fetched once to get the initial value.


jorgemmsilva (approved):
did you run the cluster tests?

There could be some race condition where the protocol parameters are needed before those are set

muXxer:
Thats why I left it as a draft, until the cluster tests on my PC are done :trollface:

packages/parameters/l1parameters.go
```diff
@@ -80,7 +80,7 @@ func L1() *L1Params {
 func L1NoLock() *L1Params {
 	if l1Params == nil {
 		if isTestContext() {
-			l1Params = L1ForTesting
+			l1Params = l1ForTesting
```
jorgemmsilva, line 83: shouldn't this use the parameters from `testutil`?

jorgemmsilva (approved):

packages/parameters/l1parameters.go
muXxer: Yeah, but that would create a circular dependency. I will remove this stuff in another refactor PR.
Just wanted to split this up a bit to fix the bug for now before doing a big refactor.

Issue #1414, opened 2022-11-08
Replace Nanomsg


fijter:
Figure out what makes sense, use websockets we already implement for parts, or MQTT.

fijter:
Pending changes in wasm part as well, websocket changes primarily complete and working.

lmoe:
solved with #1981 and #1978

PR #1978, opened 2023-02-20
Add websockets v2

lmoe:
# Description of change

Adds a new websocket to the webapi. 

To be found on `ws://api.wasphost/ws`

Supports subscribing to message topics 
```
{ "command": "subscribe", "topic": "request" }
```
Supports subscribing to chains (either all, or selected)
```
{ "command": "subscribe", "topic": "chains" }
{ "command": "subscribe", "topic": "chains/<chainID>" }
``` 

fixes issue #1414


plugins/webapi/component.go
```diff
@@ -228,5 +249,13 @@ func run() error {
 		Plugin.LogPanicf("failed to start worker: %s", err)
 	}
 
+	if err := Plugin.Daemon().BackgroundWorker("WebsocketHub", func(ctx context.Context) {
+		go deps.WebsocketHub.Run(ctx)
```
muXxer: Remove this go routine and wait for the `Run` to finish.
Remove the `Done() `

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
 
-func New(log *logger.Logger, publisher *publisher.Publisher, msgTypes []string) *PublisherWebSocket {
+func New(log *logger.Logger, hub *websockethub.Hub, msgTypes []string, publisher *publisher.Publisher) *PublisherWebSocket {
 	msgTypesMap := make(map[string]bool)
 	for _, t := range msgTypes {
 		msgTypesMap[t] = true
 	}
 
+	subscriptionManager := subscriptionmanager.New(
+		subscriptionmanager.WithMaxTopicSubscriptionsPerClient[websockethub.ClientID, string](5),
```
muXxer, line 36: 5? Is that enough?
Should this maybe a config parameter?

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+		if !p.subscriptionManager.ClientSubscribedToTopic(client.ID(), event.Kind) {
 			return
 		}
 
-		select {
-		case ch <- ev.MsgType + " " + strings.Join(ev.Parts, " "):
-		default:
-			p.log.Warnf("dropping websocket message for %s", r.RemoteAddr)
+		if err := client.Send(ctx, event); err != nil {
+			p.log.Warnf("error sending message: %v", err)
```
muXxer, line 75: maybe add the info to which client ID the send failed.

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+	CommandClientWasUnsubscribed = "client_unsubscribed"
+)
+
+type SubscriptionCommand struct {
+	Command string `json:"command"`
+	Topic   string `json:"topic"`
+}
+
+func (p *PublisherWebSocket) handleSubscriptionCommand(ctx context.Context, client *websockethub.Client, message []byte) {
+	p.log.Info(string(message))
```
muXxer, line 100: what is that log???

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+
+	var command SubscriptionCommand
+	if err := json.Unmarshal(message, &command); err != nil {
+		p.log.Warnf("Could not deserialize message to type ControlCommand, msg: '%v'", message)
```
muXxer, line 104: you will print a byte slice here as a log message. better encode it to hex before.

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+	case CommandUnsubscribe:
+		p.subscriptionManager.Unsubscribe(client.ID(), command.Topic)
+		err = client.Send(ctx, SubscriptionCommand{
+			Command: CommandClientWasUnsubscribed,
+			Topic:   command.Topic,
+		})
+	}
+
+	if err != nil {
+		p.log.Warnf("error sending message: %v", err)
```
muXxer, line 126: maybe add the info about the client here as well

packages/publisher/publisherws/publisherws.go
muXxer: And add the info about the client.

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+	return eventClosure
+}
+
+type BaseCommand struct {
+	Command string `json:"command"`
+}
+
+const (
+	CommandSubscribe           = "subscribe"
+	CommandClientWasSubscribed = "client_subscribed"
```
muXxer, line 88: maybe just `subscribed` is fine?

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+
+	CommandUnsubscribe           = "unsubscribe"
+	CommandClientWasUnsubscribed = "client_unsubscribed"
```
muXxer, line 91: maybe just `unsubscribed` is fine?

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+		p.log.Warnf("Could not deserialize message to type ControlCommand, msg: '%v'", message)
+		return
 	}
 
+	var err error
+
+	switch command.Command {
+	case CommandSubscribe:
+		p.subscriptionManager.Subscribe(client.ID(), command.Topic)
+		err = client.Send(ctx, SubscriptionCommand{
```
muXxer, line 113: And I'm wondering if that is a command.... isn't it more an event?

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+			Command: CommandClientWasSubscribed,
+			Topic:   command.Topic,
+		})
+	case CommandUnsubscribe:
+		p.subscriptionManager.Unsubscribe(client.ID(), command.Topic)
+		err = client.Send(ctx, SubscriptionCommand{
```
muXxer, line 119: And I'm wondering if that is a command.... isn't it more an event?

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+
+	if err != nil {
+		p.log.Warnf("error sending message: %v", err)
+	}
+}
+
+func (p *PublisherWebSocket) handleNodeCommands(client *websockethub.Client, message []byte) {
+	var baseCommand BaseCommand
+	if err := json.Unmarshal(message, &baseCommand); err != nil {
+		p.log.Warnf("Could not deserialize message to type BaseCommand")
```
muXxer, line 133: maybe add the actual error into the `Warnf`

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+		return
+	}
+
+	switch baseCommand.Command {
+	case CommandSubscribe, CommandUnsubscribe:
+		p.handleSubscriptionCommand(client.Context(), client, message)
+	default:
+		p.log.Warnf("Could not deserialize message")
```
muXxer, line 141: The error message should be that the command is unknown. maybe add the value of the command into the `Warnf`

packages/publisher/publisherws/publisherws.go
muXxer: And the clientID

packages/publisher/publisherws/publisherws.go
muXxer: And the client ID

packages/publisher/publisherws/publisherws.go
```diff
@@ -4,82 +4,189 @@
...
+	p.publisher.Events.Published.Hook(eventWriter)
+	defer p.publisher.Events.Published.Detach(eventWriter)
+
 	for {
-		msg := <-ch
-		err := c.Write(ctx, websocket.MessageText, []byte(msg))
-		if err != nil {
-			c.Close(websocket.StatusInternalError, err.Error())
-			break
+		select {
```
muXxer, line 153: Maybe nest the ReceiveChan in another default case to prevent race conditions:

```go
				for {
					// we need to nest the client.ReceiveChan into the default case because
					// the select cases are executed in random order if multiple
					// conditions are true at the time of entry in the select case.
					select {
					case <-client.ExitSignal:
						// client was disconnected
						return
					default:
						select {
						case <-client.ExitSignal:
							// client was disconnected
							return
						case msg, ok := <-client.ReceiveChan:
							if !ok {
								// client was disconnected
								return
							}```

File: packages/publisher/publisherws/publisherws.go [hunk 14]
```diff
@@ -4,82 +4,189 @@
...
+
+// ServeHTTP serves the websocket.
+// Provide a chainID to filter for a certain chain, provide an empty chain id to get all chain events.
+func (p *PublisherWebSocket) ServeHTTP(c echo.Context) error {
+	p.log.Infof("ServeHTTP ctx: %v %v", c.Request().Context(), c.Request().Context().Err())
 
-	return nil
+	return p.hub.ServeWebsocket(c.Response(), c.Request(),
+		func(client *websockethub.Client) {
+			go p.OnClientCreated(client)
```
muXxer, line 186: I would call this synchronous, and start another go routine inside the `OnClientCreated` after the `client.ReceiveChan` was created. Otherwise there is a race condition where you could loose some messages from the client at the beginning.

muXxer (approved):
The requested changes will be made later on by @lmoe

PR #1988, opened 2023-02-21
fix: Add network check in wasp-cli

### LINKED ISSUE #1994: wasp-cli should check the hrp of the config and the target network.

howjmay:
Closes #1994

### DIFF
--- a/tools/wasp-cli/cli/config/config.go
+++ b/tools/wasp-cli/cli/config/config.go
@@ -124,7 +124,7 @@ func GetChain(name string) isc.ChainID {
 	log.Check(err)
 
 	if networkPrefix != parameters.L1().Protocol.Bech32HRP {
-		err = fmt.Errorf("target network not matches wasp-cli config")
+		err = fmt.Errorf("target network of the L1 node does not match the wasp-cli config")
 	}
 	log.Check(err)
 
--- a/tools/wasp-cli/cli/config/config.go
+++ b/tools/wasp-cli/cli/config/config.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/spf13/viper"
 
+	iotago "github.com/iotaledger/iota.go/v3"
 	"github.com/iotaledger/wasp/packages/isc"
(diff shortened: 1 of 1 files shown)


tools/wasp-cli/cli/config/config.go
```diff
@@ -118,7 +118,11 @@ func AddChain(name, chainID string) {
 }
 
 func GetChain(name string) isc.ChainID {
-	chainID, err := isc.ChainIDFromString(viper.GetString("chains." + name))
+	configChainID := viper.GetString("chains." + name)
+	chainID, err := isc.ChainIDFromString(configChainID)
```
muXxer: IMO we should do that check in a bit cleaner way, to understand whats going on.

I would use the networkPrefix parameter from the `networkPrefix, addr, err := iotago.ParseBech32(bech32)` , which is the HRP, and compare it with the `parameters.L1().Protocol.Bech32HRP`.

muXxer (changes requested):

tools/wasp-cli/cli/config/config.go
howjmay: Fixed. However, I am not sure if the current error message is a good one

tools/wasp-cli/cli/config/config.go
```diff
@@ -118,7 +119,16 @@ func AddChain(name, chainID string) {
 }
 
 func GetChain(name string) isc.ChainID {
-	chainID, err := isc.ChainIDFromString(viper.GetString("chains." + name))
+	configChainID := viper.GetString("chains." + name)
+	networkPrefix, _, err := iotago.ParseBech32(configChainID)
+	log.Check(err)
+
+	if networkPrefix != parameters.L1().Protocol.Bech32HRP {
+		err = fmt.Errorf("target network not matches wasp-cli config")
```
muXxer:
```suggestion
		err = fmt.Errorf("target network of the L1 node does not match the wasp-cli config")
```

muXxer (approved):

PR #1997, opened 2023-02-22
Fix committee nodes updates between components.

kape1395:
Also:
  - L1 retry of block cannot be fetched.
  - rundkg command updated, output made more useful, quorums fixed.
  - other commands with \n added...

### DIFF
--- a/packages/chain/node.go
+++ b/packages/chain/node.go
@@ -582,7 +582,7 @@ func (cni *chainNodeImpl) handleTxPublished(ctx context.Context, txPubResult *tx
 }
 
 func (cni *chainNodeImpl) handleAliasOutput(ctx context.Context, aliasOutput *isc.AliasOutputWithID) {
-	cni.log.Debugf("handleAliasOutput, aliasOutput[StateIndex=%v].ID=", aliasOutput.GetStateIndex(), aliasOutput.OutputID().ToHex())
+	cni.log.Debugf("handleAliasOutput: %v", aliasOutput)
 	cni.stateTrackerCnf.TrackAliasOutput(aliasOutput, true)
 	cni.stateTrackerAct.TrackAliasOutput(aliasOutput, false) // ACT state will be equal to CNF or ahead of it.
 	outMsgs := cni.chainMgr.Input(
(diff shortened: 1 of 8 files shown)


packages/l1connection/l1connection.go
```diff
@@ -125,18 +125,29 @@ func (c *l1client) postBlock(ctx context.Context, block *iotago.Block) (*iotago.
 			return nil, fmt.Errorf("failed during local PoW: %w", err)
 		}
 	}
-	block, err := c.nodeAPIClient.SubmitBlock(ctx, block, parameters.L1().Protocol)
-	if err != nil {
-		return nil, fmt.Errorf("failed to submit block: %w", err)
+
+	var postedBlock *iotago.Block
+	maxRetries := 10
+	for i := 0; ; i++ {
```
kape1395: This doesn't work either because it fails with an error similar to "duplicate" on the post part.

packages/chain/node.go
```diff
@@ -843,17 +842,25 @@ func (cni *chainNodeImpl) deriveActiveAccessNodes() {
...
+	if !anSame {
 		cni.log.Infof("Access nodes updated, active=%+v", activeAccessNodes)
+		cni.listener.AccessNodesUpdated(cni.chainID, activeAccessNodes)
+	}
+	if !anSame || !cnSame {
+		if !cnSame {
+			cni.log.Infof("Committee nodes updated, active=%+v", activeCommitteeNodes)
+		}
 		cni.mempool.AccessNodesUpdated(activeCommitteeNodes, activeAccessNodes)
 		cni.stateMgr.ChainNodesUpdated(serverNodes, activeAccessNodes, activeCommitteeNodes)
```
jorgemmsilva, line 863:
```suggestion
	anSame := util.Same(oldAccessNodes, activeAccessNodes)
	cnSame := util.Same(oldCommitteeNodes, activeCommitteeNodes)
	if !anSame {
		cni.log.Infof("Access nodes updated, active=%+v", activeAccessNodes)
		cni.listener.AccessNodesUpdated(cni.chainID, activeAccessNodes)
		cni.mempool.AccessNodesUpdated(activeCommitteeNodes, activeAccessNodes)
	}
	if !cnSame {
		cni.log.Infof("Committee nodes updated, active=%+v", activeCommitteeNodes)
		cni.stateMgr.ChainNodesUpdated(serverNodes, activeAccessNodes, activeCommitteeNodes)
```
couldn't we do it like this?

jorgemmsilva (approved):

packages/chain/node.go
kape1395: Because, for example, `cni.stateMgr.ChainNodesUpdated` has to be called if access nodes OR committee nodes are changed. I decided it is better to duplicate checks than the actual calls to make fewer mistakes.

PR #2004, opened 2023-02-22
gasPerToken -> Ratio32 + new state schema migration mechanism

dessaya:


packages/webapi/models/governance.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type GovChainInfoResponse struct {
+	ChainID         string       `json:"chainID" swagger:"desc(ChainID (Bech32-encoded).),required"`
+	ChainOwnerID    string       `json:"chainOwnerId" swagger:"desc(The chain owner address (Bech32-encoded).),required"`
+	Description     string       `json:"description" swagger:"desc(The description of the chain.),required"`
+	GasFeePolicy    GasFeePolicy `json:"gasFeePolicy" swagger:"desc(The gas fee policy),required"`
+	MaxBlobSize     uint32       `json:"maxBlobSize" swagger:"desc(The maximum contract blob size.),required"`
+	MaxEventSize    uint16       `json:"maxEventSize" swagger:"desc(The maximum event size.),required"`                      // TODO: Clarify
+	MaxEventsPerReq uint16       `json:"maxEventsPerReq" swagger:"desc(The maximum amount of events per request.),required"` // TODO: Clarify
```
jorgemmsilva, line 19: maybe these are silly and could be removed now that we charge gas for everything. Gas can act as the limiting factor by itself

packages/vm/vmcontext/migrations.go
```diff
@@ -0,0 +1,46 @@
...
+		return
+	}
+
+	var schemaVersion uint32
+	vmctx.callCore(root.Contract, func(s kv.KVStore) {
+		schemaVersion = root.GetSchemaVersion(s)
+	})
+	if schemaVersion < migrations.BaseSchemaVersion {
+		panic(fmt.Sprintf("inconsistency: schema version %d should be >= %d", schemaVersion, migrations.BaseSchemaVersion))
+	}
```
jorgemmsilva: this can never happen. `migrations.BaseSchemaVersion` is 0 and the `schemaVersion` is an uint

packages/vm/vmcontext/migrations.go
```diff
@@ -0,0 +1,46 @@
...
+	if schemaVersion > migrations.LatestSchemaVersion {
+		panic(fmt.Sprintf("inconsistency: schema version %d should be <= %d", schemaVersion, migrations.LatestSchemaVersion))
+	}
+
+	for schemaVersion < migrations.LatestSchemaVersion {
+		migration := migrations.Migrations[schemaVersion-migrations.BaseSchemaVersion]
```
jorgemmsilva: shouldn't this be
```suggestion
		schemaVersion++
		migration := migrations.Migrations[schemaVersion]
```
?

if latest schema is 10 and current 3 this line is doing 3-10, that's weird.

dessaya: Let's say `base = 5`, `current = 7`, `latest = 10`, this would happen because there are 5 migrations in the list, the first one corresponding to schema version 5.

```
for schemaVersion < migrations.LatestSchemaVersion { // 7 < 10
  migration := migrations.Migrations[schemaVersion-migrations.BaseSchemaVersion] // 7-5 = 2
```

I think it's ok.

packages/vm/vmcontext/migrations.go
dessaya: But BaseSchemaVersion may be > 0 in the future, when we decide to prune old migrations. This check would panic if the node is behind pruned migrations.

packages/webapi/models/governance.go
dessaya: Maybe... I'd rather we change this in another PR. I'll make an issue.

packages/vm/vmcontext/migrations.go
jorgemmsilva: oh okay,  good point :+1:

packages/vm/vmcontext/migrations.go
jorgemmsilva: I read `BaseSchemaVersion` as `latest` for some reason. :facepalm:

jorgemmsilva (approved):
I find the migrations pruning logic kinda weird, we need to be extra careful when changing that stuff.

otherwise, great change :rocket:

PR #2010, opened 2023-02-23
fix(evm)!: make sure block hash does not change after changing block gas limit

dessaya:
This is a breaking change, since it changes how EVM blocks are stored in the state.


muXxer:
Don't worry about the linter. I will fix it soon.
But currently I'm reviewing the websocket PR.

muXxer (approved):

PR #2011, opened 2023-02-23
Mark math/rand sources as pseudo RNG

muXxer:


packages/trie/trie_test.go
```diff
@@ -579,7 +580,9 @@ const letters = "abcdefghijklmnop"
 
 func genRnd3() []string {
 	ret := make([]string, 0, len(letters)*len(letters)*len(letters))
-	rnd := rand.New(rand.NewSource(1))
+
+	// TODO: should this really be seeded with 1?
+	rnd := util.NewPseudoRand(1)
```
muXxer, line 584: @dessaya can you maybe answer this? wouldn't it be better to have a random seed for the RNG? Or was this done for a specific reason?

dessaya: This code is originally by Evaldas I think.
I think the intention is to make the test as deterministic as possible, which I think is a good idea in general. So yes, seed it with anything that is constant.

dessaya (approved):
I'm not sure I like these "pseudo" all over the place. I mean, we already know random numbers are not really random... Even cryptographically secure RNGs are pseudo-random, so...

kape1395 (changes requested):
I will review it tomorrow.

packages/trie/trie_test.go
muXxer: Alright, then I will remove the TODO and leave it as it is!
I was just wondering because we once had the problem that we didn't found a bug in our codebase, because the test used a fixed seed for the RNG, so the tests were deterministic, but they never hit that one case where the error occurred :laughing:

muXxer:
> I'm not sure I like these "pseudo" all over the place. I mean, we already know random numbers are not really random... Even cryptographically secure RNGs are pseudo-random, so...

I thought `Pseudo` is a better word than `CryptoInsecureRand` :trollface:

kape1395 (approved):

PR #2017, opened 2023-02-24
Fix chain shutdown/deactivation handling

muXxer:


kape1395 (changes requested):
`TestRebootDuringTasks` fails for me. I am still trying to figure out why. Will investigate.

kape1395 (approved):
Today the `TestRebootDuringTasks` passes.
The sporadic failures may not be related to this change.

Issue #2033, opened 2023-02-28
controlAddresses in blocklog  can be removed


jorgemmsilva:
this information can be derived from the alias output, this is quite superfluous and should be removed for simplicity

Issue #2002, opened 2023-02-22
an increase in  L1 SD costs will crash WASP at the moment.


jorgemmsilva:
https://github.com/iotaledger/wasp/blob/develop/packages/vm/vmcontext/vmcontext.go#L112-L128

the way we calculate "storageDepositAssumption" at the start of the chain and use it forever needs to change, since the SD price is dynamic.

Any time an "accounting output" requires a SD increase, it should be covered either by the "common account" or by "the request sender"

---

special tests we need to be created for this

PR #2034, opened 2023-02-28
refactor: remove init tx

jorgemmsilva:
this removes the need for the "initTx", and by simplifying the process of chain deployment (to a single output), reduces the number of edge cases and allows requests to be sent right away to the chain.

- core contracts don't have constructors anymore, they instead init the state using information in the origin alias output
- removes descrition from the chain and corecontracts (not from from wasm contracts to not pile up more changes)
- refactors the "blockcontext" so that we don't need the "privileged sandbox methods" for that anymore

further work:
- chain Owner could be derived from the chain govenor - to be discussed
- SD assumptions need to be removed and we need to be able to handle protocol changes on L1, including min SD increase ([issue](https://github.com/iotaledger/wasp/issues/2002))
- controlAddresses in blocklog seem pretty useless and can be removed ([issue](https://github.com/iotaledger/wasp/issues/2033))
- remove description from wasm contracts

TODO:
@kape1395 please take a look at `(cni *chainNodeImpl) handleAliasOutput`. I think this is pretty okay, but please do confirm you agree with initializing the chain state there.


kape1395 (changes requested):
Do not merge until I'm done with

> @kape1395 please take a look at (cni *chainNodeImpl) handleAliasOutput. I think this is pretty okay, but please do confirm you agree with initializing the chain state there

packages/apilib/deploychain.go
```diff
@@ -92,60 +85,37 @@ func CreateChainOrigin(
 	originator *cryptolib.KeyPair,
 	stateController iotago.Address,
 	governanceController iotago.Address,
-	dscr string, initParams dict.Dict,
-) (isc.ChainID, *iotago.Transaction, error) {
+	dscr string,
```
dessaya: Is this `dscr` parameter still needed?

packages/evm/jsonrpc/evmchain.go
```diff
@@ -144,7 +144,7 @@ func (e *EVMChain) iscStateFromEVMBlockNumber(blockNumber *big.Int) (state.State
 		return nil, fmt.Errorf("block number is too large: %s", blockNumber)
 	}
 	// the first EVM block (number 0) is "minted" at ISC block index 1 (init chain)
```
dessaya: This comment is now outdated?

```suggestion
	// the first EVM block (number 0) is "minted" at ISC block index 0
```

packages/origin/origin.go
```diff
@@ -0,0 +1,75 @@
...
+	d.Set(kv.Key(coreutil.StatePrefixTimestamp), codec.EncodeTime(time.Unix(0, 0)))
+
+	contractState := func(contract *coreutil.ContractInfo) kv.KVStore {
+		return subrealm.New(d, kv.Key(contract.Hname().Bytes()))
+	}
+
+	chainOwner := codec.MustDecodeAgentID(initParams.MustGet(governance.ParamChainOwner), &isc.NilAgentID{})
+
+	blockKeepAmount := codec.MustDecodeInt32(initParams.MustGet(evm.FieldBlockKeepAmount), evm.BlockKeepAmountDefault)
+	evmChainID := evmtypes.MustDecodeChainID(initParams.MustGet(evm.FieldChainID), evm.DefaultChainID)
```
dessaya: Does this work correctly with `wasp-cli`? The EVM init params were prefixed with `root.ParamEVM` to avoid collisions with init params of other core contracts.

Maybe we should place all init param keys on the same file.

packages/state/db.go
```diff
@@ -15,10 +15,6 @@ const (
 	prefixBlockByTrieRoot = iota
 	prefixTrie
 	prefixLatestTrieRoot
-
-	// KeyChainID is the key used to store the chain ID in the state.
-	// It should not collide with any hname prefix (which are 32 bits long).
-	KeyChainID = "\x00"
```
dessaya, line 21: The chain ID is stored somewhere else now?

packages/state/state_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-package state
+package state_test
```
dessaya, line 4: Is this standard? There are now 2 different packages on the same folder :thinking: 

Also, why change this?

packages/transaction/origintx.go
```diff
@@ -39,14 +53,16 @@ func NewChainOriginTransaction(
...
-		if aliasOutput.Amount < aliasStorageDeposit {
-			aliasOutput.Amount = aliasStorageDeposit
-		}
-	}
+
+	// minSD := parameters.L1().Protocol.RentStructure.MinRent(aliasOutput)
+	// if aliasOutput.Deposit() < minSD {
+	// 	aliasOutput.Amount = minSD
+	// 	aliasOutput.StateMetadata = origin.L1Commitment(initParams, minSD).Bytes()
+	// }
```
dessaya: intentionally left commented out?

dessaya (commented):
Here: https://github.com/iotaledger/wasp/blob/develop/packages/vm/vmcontext/migrations.go#L14

Maybe the initialization of the schema version should be done when initializing the root contract instead?

packages/state/state_test.go
jorgemmsilva: this is standard, to test the current package as a blackbox. 
https://pkg.go.dev/testing

don't remember in particular now, but this is a good practice overall and I think we should do this where possible to avoid polluting the main package with testing functions

packages/state/db.go
jorgemmsilva: we don't need it. It can always be derived from somewhere else. the node knows what chain is running the task on.
Also the chainID is not known before issuing the originTx, meaning that if it is part of the state, we couldn't calculate the origin commitment to include it in the output.

packages/origin/origin.go
jorgemmsilva: good point :+1:

dessaya: I think this function should be removed now:
https://github.com/iotaledger/wasp/blob/develop/packages/vm/core/root/interface.go#L41

dessaya (approved):

packages/chain/node.go
```diff
@@ -584,6 +587,21 @@ func (cni *chainNodeImpl) handleTxPublished(ctx context.Context, txPubResult *tx
...
+			initParams, err = dict.FromBytes(originMetadata.Data)
+			if err != nil {
+				panic(fmt.Sprintf("invalid parameters on origin AO, %s", err.Error()))
+			}
+		}
+		aoSD := transaction.NewStorageDepositEstimate().AnchorOutput
+		origin.InitChain(cni.chainStore, initParams, aliasOutput.GetAliasOutput().Amount-aoSD)
+	}
+
+	cni.log.Debugf("handleAliasOutput, aliasOutput[StateIndex=%v].ID=", aliasOutput.GetStateIndex(), aliasOutput.OutputID().ToHex())
```
kape1395: This debug duplicates the one above. And AliasOutpitWithID.String() prints also the stateIndex, so this can be just removed.

packages/chain/node.go
```diff
@@ -584,6 +587,21 @@ func (cni *chainNodeImpl) handleTxPublished(ctx context.Context, txPubResult *tx
 
 func (cni *chainNodeImpl) handleAliasOutput(ctx context.Context, aliasOutput *isc.AliasOutputWithID) {
 	cni.log.Debugf("handleAliasOutput: %v", aliasOutput)
+	// TODO is this the best place to do this?
+	if aliasOutput.GetStateIndex() == 0 {
```
kape1395, line 588: I want to move the internals of this `if` to some function, but I am not sure in which package. `state` or `origin`?

packages/chain/node.go
```diff
@@ -584,6 +587,21 @@ func (cni *chainNodeImpl) handleTxPublished(ctx context.Context, txPubResult *tx
...
+		var initParams dict.Dict
+		if originMetadata := aliasOutput.GetAliasOutput().FeatureSet().MetadataFeature(); originMetadata != nil {
```
kape1395: This place is probably OK, but we have to make sure the origin state is also downloadable in the statemgr. Have you done something about that?

packages/chain/node_test.go
```diff
@@ -43,34 +43,37 @@ type tc struct {
...
+		{n: 4, f: 0, reliable: true, timeout: 40 * time.Second},   // Minimal robust config.
+		{n: 4, f: 1, reliable: true, timeout: 50 * time.Second},   // Minimal robust config.
+		{n: 10, f: 3, reliable: true, timeout: 130 * time.Second}, // Typical config.
 	}
 	if !testing.Short() {
 		tests = append(tests,
-			tc{n: 4, f: 1, reliable: false},  // Minimal robust config.
-			tc{n: 10, f: 3, reliable: false}, // Typical config.
-			tc{n: 31, f: 10, reliable: true}, // Large cluster, reliable - to make test faster.
+			// TODO these "unreliable" tests are crazy, they either succeed in 10~20s or run forever...
```
kape1395: Don't comment them out. This can hide error.
Have you tried to increase retry periods?

packages/chain/statemanager/smGPA/state_manager_gpa.go
```diff
@@ -478,7 +485,12 @@ func (smT *stateManagerGPA) completeRequests(alreadyCommittedL1C *state.L1Commit
 				var stateDraft state.StateDraft
 				previousCommitment := block.PreviousL1Commitment()
 				var err error
-				stateDraft, err = smT.store.NewEmptyStateDraft(previousCommitment)
+				if previousCommitment == nil {
+					// origin block
+					stateDraft = smT.store.NewOriginStateDraft()
```
kape1395, line 490: Is `NewOriginStateDraft` still valid and meaningful?
Should it take the AO as a parameter?

packages/evm/jsonrpc/evmchain.go
```diff
@@ -508,10 +508,10 @@ func iscBlockIndexByEVMBlockNumber(n uint64) (uint32, error) {
 	if n > math.MaxUint32-1 {
 		return 0, fmt.Errorf("block number is too large: %d", n)
 	}
-	return uint32(n) + 1, nil
+	return uint32(n), nil
```
kape1395, line 511: Why has this changed?

kape1395 (changes requested):

packages/chain/node.go
jorgemmsilva: I tried that, but started getting import cycles. This works and it's not too extensive imo

packages/chain/node.go
jorgemmsilva: Yep, done and tested

packages/chain/node_test.go
jorgemmsilva: These hang for a really really really long time. I saw the same issue happening in develop. Doesn't seem like a good test to have if we want to "speed up tests"

packages/evm/jsonrpc/evmchain.go
jorgemmsilva: Now Evm chain block number == ISC block number, they both start at origin.

packages/chain/statemanager/smGPA/state_manager_gpa.go
jorgemmsilva: This is just to apply the origin block when syncing

packages/chain/node_test.go
kape1395: But these tests are essential. They check how the chain copes with connection drops, message reorderings, duplicated messages, etc.

kape1395 (approved):

PR #2037, opened 2023-02-28
Register erc20 external

ivange94:
# Description of change

A few core contract convenience cli commands

### DIFF
--- a/tools/wasp-cli/chain/create-foundry.go
+++ b/tools/wasp-cli/chain/create-foundry.go
@@ -1,21 +1,38 @@
 package chain
 
 import (
+	"encoding/hex"
+	"math/big"
+
+	iotago "github.com/iotaledger/iota.go/v3"
+	"github.com/iotaledger/wasp/packages/kv/codec"
 	"github.com/iotaledger/wasp/packages/vm/core/accounts"
 	"github.com/spf13/cobra"
 )
 
 func initCreateFoundryCmd() *cobra.Command {
+	var maxSupply, mintedTokens, meltedTokens int64
+
 	return buildPostRequestCmd(
 		"create-foundry",
 		"Call accounts core contract foundryCreateNew to create a new foundry",
 		accounts.Contract.Name,
 		accounts.FuncFoundryCreateNew.Name,
 		func(cmd *cobra.Command) {
-			cmd.Flags().String("token-scheme", "", "Token scheme")
+			cmd.Flags().Int64Var(&maxSupply, "max-supply", 1000000, "Maximum token supply")
+			cmd.Flags().Int64Var(&mintedTokens, "minted-tokens", 0, "Minted tokens")
+			cmd.Flags().Int64Var(&meltedTokens, "melted-tokens", 0, "Melted tokens")
 		},
 		func(cmd *cobra.Command) []string {
-			return []string{"string", "t", "bytes", cmd.Flag("token-scheme").Value.String()}
+			tokenScheme := &iotago.SimpleTokenScheme{
+				MaximumSupply: big.NewInt(maxSupply),
+				MintedTokens:  big.NewInt(mintedTokens),
+				MeltedTokens:  big.NewInt(meltedTokens),
+			}
+
+			tokenSchemeBytes := codec.EncodeTokenScheme(tokenScheme)
+
+			return []string{"string", "t", "bytes", "0x" + hex.EncodeToString(tokenSchemeBytes)}
 		},
 	)
 }
--- a/tools/wasp-cli/chain/create-foundry.go
+++ b/tools/wasp-cli/chain/create-foundry.go
@@ -1,0 +1,21 @@
+package chain
+
+import (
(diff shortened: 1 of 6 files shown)


tools/wasp-cli/chain/create-foundry.go
```diff
@@ -0,0 +1,21 @@
...
+	return buildPostRequestCmd(
+		"create-foundry",
+		"Call accounts core contract foundryCreateNew to create a new foundry",
+		accounts.Contract.Name,
+		accounts.FuncFoundryCreateNew.Name,
+		func(cmd *cobra.Command) {
+			cmd.Flags().String("token-scheme", "", "Token scheme")
+		},
+		func(cmd *cobra.Command) []string {
+			return []string{"string", "t", "bytes", cmd.Flag("token-scheme").Value.String()}
```
dessaya: I don't understand, what format is used to specify the token scheme?

Also, can we add a test for this new command? (Or change existing tests to use it instead). Ideally every subcommand should have at least one test.

tools/wasp-cli/chain/register-erc20-native-token.go
```diff
@@ -12,62 +10,94 @@ import (
...
+		"register-erc20-native-token-on-remote-chain",
+		"Call evm core contract registerERC20NativeTokenOnRemoteChain entry point",
+		evm.Contract.Name,
+		evm.FuncRegisterERC20NativeTokenOnRemoteChain.Name,
+		func(cmd *cobra.Command) {
+			initRegisterERC20NativeTokenParams(cmd)
+			cmd.Flags().Uint8P("target", "A", 0, "Target chain ID")
+		},
+		func(cmd *cobra.Command) []string {
+			extraArgs := []string{"string", "A", "uint8", cmd.Flag("target").Value.String()}
```
dessaya: Is that `uint8` correct?

Also, can we add a test for this command as well?

ivange94: That's what's on the docs.

ivange94: So I actually had a test for this. It does calls the endpoint but it fails because of the `chainID`. It keeps reporting that the `chainID` is not a valid address type. I'll push the test so you see.

tools/wasp-cli/chain/create-foundry.go
ivange94: So for this I wasn't exactly sure myself. The `TokenScheme` param is not a primitive type but a user can only use primitive types(numbers, booleans and strings) with the command line. So I had to look at the implementation of the `util.EncodeParams` to see what types will best work for `TokenScheme`. I actually thought of adding an implementation for `TokenScheme` but it just looked exactly like whats there for `bytes`. 

Basically, to use this param, the token scheme needs to be sent as hex encoded bytes. This works from the command line but I'll add a test for it.

ivange94: Actually found the code I drew inspiration from. 

```go
func (w *WaspCLITest) CreateL2Foundry(tokenScheme iotago.TokenScheme) {
	tokenSchemeBytes := codec.EncodeTokenScheme(tokenScheme)
	out := w.PostRequestGetReceipt(
		"-o", "accounts", accounts.FuncFoundryCreateNew.Name,
		"string", accounts.ParamTokenScheme, "bytes", "0x"+hex.EncodeToString(tokenSchemeBytes),
		"--allowance", "base:1000000", "--node=0",
	)
	require.Regexp(w.T, `.*Error: \(empty\).*`, strings.Join(out, "\n"))
}
```

dessaya: Yes but the idea of these convenience commands is to make it easier for the user. Having to supply a hex-encoded value is not very friendly IMO.

Note that the only supported implementation of the `TokenScheme` interface is `SimpleTokenScheme` which is a straightforward struct with just 3 numbers.

tools/wasp-cli/chain/register-erc20-native-token.go
ivange94: @dessaya this part still isn't clear to me. What type should the `target address` be? The docs says `uint8` but also mentions that this is the target chain. So it's not clear to me if this is the chainID or some other address format.

dessaya: `target` is the chain ID of the chain you want to register the tokens into.

Let's say Bob creates a foundry on Chain A, and wants to register it on Chain B. Then, Bob calls `registerERC20NativeTokenOnRemoteChain` on Chain A, with `target = ChainID(B)`.

ivange94: I've updated my PR with some tests. The `onRemoteChain` test fails. Here's the error from the function call 
`Error: unknown address type: type 97`

dessaya: The endpoint expects the `target` parameter to be a ChainID

dessaya (approved):

PR #2038, opened 2023-03-01
feat: add gasfeepolicy and custom metadata to Alias Output

jorgemmsilva:


packages/vm/vmcontext/txbuilder.go
```diff
@@ -11,16 +11,28 @@ import (
...
-func (vmctx *VMContext) BuildTransactionEssence(stateData *state.L1Commitment) (*iotago.TransactionEssence, []byte) {
-	return vmctx.txbuilder.BuildTransactionEssence(stateData)
+func (vmctx *VMContext) StateMetadata(stateCommitment *state.L1Commitment) []byte {
+	stateMetadata := stateCommitment.Bytes()
+	if vmctx.chainInfo == nil {
+		// TODO this should only happen during originTx, can be removed after #2034 is merged
+		return stateMetadata
+	}
+	feePolicyBytes := vmctx.chainInfo.GasFeePolicy.Bytes()
+	stateMetadata = append(stateMetadata, feePolicyBytes...)
```
dessaya: Maybe make a struct for clarity? With `Bytes()` and `FromBytes()`

```
type StateMetadata struct {
  L1Commitment *state.L1Commitment
  GasFeePolicy *gas.FeePolicy
}
```

dessaya (approved):

packages/vm/vmcontext/txbuilder.go
```diff
@@ -11,16 +11,28 @@ import (
...
 
-func (vmctx *VMContext) BuildTransactionEssence(stateData *state.L1Commitment) (*iotago.TransactionEssence, []byte) {
-	return vmctx.txbuilder.BuildTransactionEssence(stateData)
+func (vmctx *VMContext) StateMetadata(stateCommitment *state.L1Commitment) []byte {
+	stateMetadata := stateCommitment.Bytes()
+	if vmctx.chainInfo == nil {
+		// TODO this should only happen during originTx, can be removed after #2034 is merged
+		return stateMetadata
+	}
+	feePolicyBytes := vmctx.chainInfo.GasFeePolicy.Bytes()
```
muXxer: We should add some sort of version byte or type byte, otherwise its hard to parse old versions if we change the format and add more infos in the future.

packages/vm/vmcontext/txbuilder.go
```diff
@@ -3,23 +3,78 @@ package vmcontext
...
+	L1Commitment   *state.L1Commitment
+	GasFeePolicy   *gas.GasFeePolicy
+	SchemaVersion  uint32
+	CustomMetadata string
+}
+
+func (s *StateMetadata) Bytes() []byte {
+	mu := marshalutil.New()
+	mu.WriteBytes(s.L1Commitment.Bytes())
+	mu.WriteUint32(s.SchemaVersion)
```
muXxer, line 28: While we are at it, I would move the SchemaVersion as the first byte.
In case our commitment size ever changes (for example different hash function).
WDYT?

dessaya, line 28: The schema version should be first? So that when unserializing one can inspect the first bytes to know the version.

packages/vm/vmcontext/txbuilder.go
```diff
@@ -3,23 +3,78 @@ package vmcontext
...
+	}
+	ret.SchemaVersion, err = mu.ReadUint32()
+	if err != nil {
+		return nil, err
+	}
+	ret.GasFeePolicy, err = gas.FeePolicyFromMarshalUtil(mu)
+	if err != nil {
+		return nil, err
+	}
+	ret.CustomMetadata = string(mu.ReadRemainingBytes())
```
muXxer: In my opinion there should be a "size" parameter in front of following metadata of unknown length.
Thats how we do it on L1 as well.
grafik

PR #2042, opened 2023-03-02
feat(evm): websocket transport for jsonrpc

dessaya:


packages/evm/evmtypes/logs.go
```diff
@@ -135,3 +135,35 @@ func DecodeFilterQuery(b []byte) (*ethereum.FilterQuery, error) {
...
+		for _, a := range addresses {
+			if log.Address == a {
+				found = true
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	if len(topics) > 0 {
```
jorgemmsilva: `if` could be inverted to decrease indentation

packages/evm/evmtypes/logs.go
```diff
@@ -135,3 +135,35 @@ func DecodeFilterQuery(b []byte) (*ethereum.FilterQuery, error) {
...
+		if len(topics) > len(log.Topics) {
+			return false
+		}
+		for i, sub := range topics {
+			match := len(sub) == 0 // empty rule set == wildcard
+			for _, topic := range sub {
+				if log.Topics[i] == topic {
+					match = true
+					break
```
jorgemmsilva: could just `return true` here?

packages/evm/evmtypes/receipt.go
```diff
@@ -88,3 +88,32 @@ func DecodeReceiptFull(receiptBytes []byte) (*types.Receipt, error) {
...
+		if !included {
+			return false
+		}
+	}
+
+	for _, sub := range topics {
+		included := len(sub) == 0 // empty rule set == wildcard
+		for _, topic := range sub {
+			if types.BloomLookup(bloom, topic) {
+				included = true
```
jorgemmsilva: could just `return true`?

jorgemmsilva (approved):

packages/evm/evmtypes/logs.go
dessaya: No:

```
	// The Topic list restricts matches to particular event topics. Each event has a list
	// of topics. Topics matches a prefix of that list. An empty element slice matches any
	// topic. Non-empty elements represent an alternative that matches any of the
	// contained topics.
	//
	// Examples:
	// {} or nil          matches any topic list
	// {{A}}              matches topic A in first position
	// {{}, {B}}          matches any topic in first position AND B in second position
	// {{A}, {B}}         matches topic A in first position AND B in second position
	// {{A, B}, {C, D}}   matches topic (A OR B) in first position AND (C OR D) in second position
```

I'll refactor using `lo` for clarity anyway.

PR #2043, opened 2023-03-02
fix: affordable gas calc

jorgemmsilva:


packages/webapi/models/governance.go
```diff
@@ -3,7 +3,7 @@ package models
 import "github.com/iotaledger/wasp/packages/util"
 
 type GasFeePolicy struct {
-	GasPerToken       util.Ratio32 `json:"gasPerToken" swagger:"desc(The gas per token ratio (fee = gas * A/B)),required"`
+	GasPerToken       util.Ratio32 `json:"gasPerToken" swagger:"desc(The gas per token ratio (gas/token)),required"`
```
dessaya: IMO the documentation needs to mention A & B, because the ratio converted to JSON looks like `{"A":42, "B":100}`.

dessaya (approved):

PR #2048, opened 2023-03-03
refcator: Remove peering port in local-setup docker-compose

howjmay:


muXxer:
Why do we want to remove the ability to connect to the network with additional wasps?

howjmay:
> Why do we want to remove the ability to connect to the network with additional wasps?

I think I misunderstood your response last time. Let me close this PR

Issue #2067, opened 2023-03-07
Handle deposits with not enough SDs (NTs and/or NFTs)


jorgemmsilva:
```
2023-03-07T10:23:14Z	DEBUG	Chains.0x793f158a.C-0xb7565301.LI-6.VM	debit from tst1qq2m4ck6sd5q0tenmp86y7w55938gt83yn3d0r9u28sx5c7xzcerzv9kze8: not enough funds
assets: base tokens: 4396000
2023-03-07T10:23:14Z	DEBUG	Chains.0x793f158a.C-0xb7565301.LI-6.VM	goroutine 614 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
github.com/iotaledger/wasp/packages/util/panicutil.CatchAllButDBError.func1.1()
	/scratch/packages/util/panicutil/panicutil.go:57 +0x270
panic({0x2b6d6a0, 0xc0023c37f0})
	/usr/local/go/src/runtime/panic.go:884 +0x213
github.com/iotaledger/wasp/packages/util/panicutil.CatchPanicReturnError.func1.1()
	/scratch/packages/util/panicutil/panicutil.go:30 +0x137
panic({0x2b6d6a0, 0xc0023c37f0})
	/usr/local/go/src/runtime/panic.go:884 +0x213
github.com/iotaledger/wasp/packages/util/panicutil.CatchPanicReturnError.func1.1()
	/scratch/packages/util/panicutil/panicutil.go:30 +0x137
panic({0x2b6d6a0, 0xc0023c37f0})
	/usr/local/go/src/runtime/panic.go:884 +0x213
github.com/iotaledger/wasp/packages/vm/core/accounts.DebitFromAccount({0x33770d8, 0xc001961fe0}, {0x3362b40?, 0xc0023c3350?}, 0xc0008ac900)
	/scratch/packages/vm/core/accounts/fungibletokens.go:57 +0x1da
github.com/iotaledger/wasp/packages/vm/core/accounts.AdjustAccountBaseTokens({0x33770d8, 0xc001961fe0}, {0x3362b40, 0xc0023c3350}, 0xffffffffffbcec20)
	/scratch/packages/vm/core/accounts/basetokens.go:26 +0x632
github.com/iotaledger/wasp/packages/vm/vmcontext.(*VMContext).adjustL2BaseTokensIfNeeded.func1.1({0x33770d8?, 0xc001961fe0?})
[...]

PR #2069, opened 2023-03-07
fix(vm): prevent VM crash when depositing NFT with min storage deposit

dessaya:
This addresses #2067. If the chain has enough assets in the common account, then the request is skipped instead of crashing.

TODO: If the chain does not have enough assets in the common account, the VM still panics, probably because the required storage deposit for the NFTOutput is not properly calculated.

### DIFF
--- a/packages/solo/req.go
+++ b/packages/solo/req.go
@@ -17,6 +17,7 @@ import (
 	"github.com/iotaledger/wasp/packages/kv"
 	"github.com/iotaledger/wasp/packages/kv/codec"
 	"github.com/iotaledger/wasp/packages/kv/dict"
+	"github.com/iotaledger/wasp/packages/parameters"
 	"github.com/iotaledger/wasp/packages/state"
 	"github.com/iotaledger/wasp/packages/transaction"
 	"github.com/iotaledger/wasp/packages/trie"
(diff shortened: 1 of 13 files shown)


packages/solo/req.go
```diff
@@ -415,16 +416,26 @@ func (ch *Chain) EstimateGasOffLedger(req *CallParams, keyPair *cryptolib.KeyPai
...
 func (ch *Chain) EstimateNeededStorageDeposit(req *CallParams, keyPair *cryptolib.KeyPair) (uint64, error) {
-	reqDeposit := uint64(0)
-	if req.ftokens != nil {
-		reqDeposit = req.ftokens.BaseTokens
+	if req.ftokens == nil {
+		req.ftokens = isc.NewEmptyAssets()
 	}
+
+	originalDeposit := req.ftokens
+	req.ftokens = originalDeposit.Clone().AddBaseTokens(1 * isc.Million) // must be > minSD
```
jorgemmsilva: is this needed?

jorgemmsilva (approved):

packages/solo/req.go
dessaya: Yes. I can't think of an elegant way of doing this without a much bigger refactor. In order to estimate the SD you need to create the tx, but if you provide 0 deposit, creating the tx fails because it is < minSD.

jorgemmsilva: would removing this check fix that?

https://github.com/iotaledger/wasp/blob/develop/packages/solo/req.go#L269-L271

dessaya: Not really, but I found another way by creating only the output instead of the whole tx.

PR #2073, opened 2023-03-08
fix: Panic if wasp-cli config not exist

howjmay:


tools/wasp-cli/waspcmd/cmd.go
```diff
@@ -60,7 +60,11 @@ func DefaultWaspNodeFallback(node string) string {
 }
 
 func getDefaultWaspNode() string {
-	waspSettings := viper.Sub("wasp").AllSettings()
+	waspConfigSubTree := viper.Sub("wasp")
+	if waspConfigSubTree == nil {
+		log.Fatalf("no wasp-cli config file at default location")
+	}
+	waspSettings := waspConfigSubTree.AllSettings()
```
dessaya: The error message can be misleading, since you can have a `wasp-cli.json` with a missing `wasp` section.

```suggestion
	waspSection := viper.Sub("wasp")
	waspSettings := map[string]interface{}{}
	if waspSection != nil {
		waspSettings = waspSection.AllSettings()
	}
```

dessaya (approved):

dessaya (commented):
Does `wasp-cli --help` still work?

howjmay:
@dessaya yes, `help` command are triggered by `Run`, but configure files are used by `PersistentPreRun`. Therefore, this PR won't block calling `help`.
See the following clip. 

```go
return &cobra.Command{
    Version: waspVersion,
    Use:     "wasp-cli",
    Short:   "wasp-cli is a command line tool for interacting with Wasp and its smart contracts.",
    Long: `wasp-cli is a command line tool for interacting with Wasp and its smart contracts.
NOTE: this is alpha software, only suitable for testing purposes.`,
    PersistentPreRun: func(cmd *cobra.Command, args []string) {
        config.Read()
    },
    Run: func(cmd *cobra.Command, args []string) {
        cmd.Help() //nolint:errcheck
    },
}
```

dessaya (changes requested):
Actually `PersistentPreRun` is always executed before `Run`. If you provide `--help`, neither `PreRun` nor `Run` are executed.

So these work:
* `wasp-cli --help`
* `wasp-cli chain --help`

But these do not work anymore (previously they showed the help message, now they just show the error):
* `wasp-cli`
* `wasp-cli chain`

Also, `--help` shows:

```
Available Commands:
  ...
  help              Help about any command
```

but `wasp-cli help` does not work anymore.

Also, some subcommands like `decode-metadata` do not require any configuration, and now they do not work anymore unless you do `init` first.

howjmay:
@dessaya May I ask what are the commands that don't need config file?
I think if there are multiple commands, then may be I just check each of them everytime they call `viper` that will bring less change

dessaya:
Sorry, I don't know every command and subcommand.

What is the root issue that you are trying to solve? Maybe there is a simpler approach.

howjmay:
@dessaya The root issue was a simple one. I just found if `wasp-cli.json` doesn't exist then a `nil pointer` error will be returned. And I thought this should not be a proper behavior.  So I can protect all the `viper.Sub()` from this condition. This solution may not look pretty, and the error message is not the best (it will only say a certain field in `wasp-cli.json` doesn't exist), but at least remove the aformentioned error

dessaya:
Then IMO it is better to just check for nil every time we call `viper.Sub()`. It's only 3 places, so it is not a big change. I understand it is uglier, but this way we don't break functionality.

tools/wasp-cli/chain/flag.go
```diff
@@ -19,7 +19,11 @@ func defaultChainFallback(chainName string) string {
 }
 
 func getDefaultChain() string {
-	chainSettings := viper.Sub("chains").AllSettings()
+	configSubTree := viper.Sub("chains")
+	if configSubTree == nil {
+		log.Fatal("`chains` field or config file doesn't exist")
+	}
+	chainSettings := configSubTree.AllSettings()
```
dessaya:
```suggestion
    chainSettings := map[string]interface{}{}
	if viper.Sub("chains") != nil {
		chainSettings = viper.Sub("chains").AllSettings()
	}
```

This way we fall back nicely into the `case 0` below. The error message there is good enough.

tools/wasp-cli/cli/init/checkversions.go
```diff
@@ -17,7 +17,11 @@ func initCheckVersionsCmd(waspVersion string) *cobra.Command {
 		Args:  cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
 			// query every wasp node info endpoint and ensure the `Version` matches
-			waspConfig := viper.Sub("wasp").AllSettings()
+			configSubTree := viper.Sub("wasp")
+			if configSubTree == nil {
+				log.Fatal("`wasp` field or config file doesn't exist")
```
dessaya:
```suggestion
				log.Fatal("no wasp node configured, you can add a node with `wasp-cli wasp add <name> <api url>`")
```

tools/wasp-cli/waspcmd/cmd.go
```diff
@@ -60,7 +60,11 @@ func DefaultWaspNodeFallback(node string) string {
 }
 
 func getDefaultWaspNode() string {
-	waspSettings := viper.Sub("wasp").AllSettings()
+	configSubTree := viper.Sub("wasp")
+	if configSubTree == nil {
+		log.Fatal("`wasp` field or config file doesn't exist")
+	}
+	waspSettings := configSubTree.AllSettings()
```
dessaya:
```suggestion
    waspSettings := map[string]interface{}{}
	if viper.Sub("wasp") != nil {
		waspSettings = viper.Sub("chains").AllSettings()
	}
```

falling back to `case 0` below.

dessaya (approved):

dessaya (approved):

jorgemmsilva (approved):

BugFreeSoftware:
Closed as duplicate after I encountered and fixed the nil pointer issue and more.
This fix still contains bugs that are fixed in my PR #2128

PR #2074, opened 2023-03-08
wasp-cli: only print stack trace when -d is provided

dessaya:
I don't know why this was changed recently? IMO stack traces are not friendly and also not useful to an end user, so if you want to see them you should add `-d` to the command.


jorgemmsilva (approved):
It was me. The rationale was that showing stack traces would enable us do debug user problems with ease, avoiding having to run potentially dangerous commands a second time

dessaya:
Then maybe we should add a log file for that? And ask the users to provide the logfile when reporting bugs.

PR #2075, opened 2023-03-08
feat: Implement NativeTokens (evm-toolkit) 

lmoe:
# Description of change

This PR is part 1 of 2 as it's growing in size already.

Stuff to do still: 
* NFTs are missing 
* Native token amount being BigInt brings up some issues that need to be solved
* Dependencies should be checked and unused ones be cleaned up
* Range slider should be replaced with a text input 
* NativeToken amount display is currently unformatted

Done:

Considering that we do more and more changes here, the frontend was partially refactored. Most changes in the Withdraw component, the Faucet will follow. Some comments were added and functions/types were seperated into different classes.

Furthermore, the new component NetworkSettings was added as Withdraw and Faucet share the same configuration and it was unclear before, that changes in the Faucet changes things in the Withdraw component.

NativeTokens are implemented in the Withdraw component. 
The name of each token is shown instead of ID to make it more clear what is what.

ESLint and Prettifier was included to get a cohesive codebase (run `npm run lint` / `npm run format`).

Users will now connect automatically to the page if they were connected once before. So no need to click "Connect Wallet" each visit.

Layout was updated a bit too.

Some bugfixes. 

## Type of change


tools/evm-toolkit/src/components/faucet/send_funds_transaction.ts
```diff
@@ -31,33 +32,38 @@ export class SendFundsTransaction {
...
+    metadata.writeInt8('a'.charCodeAt(0));
     metadata.writeUInt32LE(evmAddressBuffer.buffer.length);
     metadata.writeBytes(evmAddressBuffer.buffer);
 
     // Write account creation flag (arg2)
     metadata.writeUInt16LE(1);
-    metadata.writeInt8("c".charCodeAt(0))
+    metadata.writeInt8('c'.charCodeAt(0));
     metadata.writeUInt32LE(1);
     metadata.writeUInt8(1);
```
jorgemmsilva, line 69: This parameter is not needed anymore, we should remove it

jorgemmsilva (approved):

tools/evm-toolkit/src/components/faucet/send_funds_transaction.ts
lmoe: Oh right! Will remove it with the next PR. Thanks!

PR #2080, opened 2023-03-08
remove MaxEventSize, MaxEventsPerReq & MaxBlobSize

### LINKED ISSUE #2008: governance: consider removing MaxEventSize and MaxEventsPerReq
These parameters seem superfluous now that we charge gas for everything. Gas can act as the limiting factor by itself.

dessaya:
These parameters are not needed anymore since the sender pays for gas instead.

Also add CustomMetadata to the getChainInfo view call.

closes: #2008


packages/vm/core/testcore/events_test.go
```diff
@@ -1,132 +1,18 @@
...
-	events, err := ch.GetEventsForContract(manyEventsContractName)
-	require.NoError(t, err)
-	require.Len(t, events, n)
-
-	events, err = ch.GetEventsForRequest(reqid)
-	require.NoError(t, err)
-	require.Len(t, events, n)
-}
-
-func TestManyEvents(t *testing.T) {
```
jorgemmsilva, line 63: should we have similar tests that fail because of the gas limit instead?
just to assert that gas limit does indeed limit the amount/size of events (in a acceptable way, maybe we need to increase the gas cost of publishing an event, or the cost per bytes)

jorgemmsilva (approved):
linter is not pleased :smile:

packages/vm/core/testcore/events_test.go
dessaya: Done. Adjusted the tests so that they assert that gas burned increases with both amount and size of events.

The linter is not pleased about autogenerated files, so I'll ignore it.

jorgemmsilva: >The linter is not pleased about autogenerated files, so I'll ignore it.

but that will break CI, no? better run the auto-fix, or ignore the generated files from the linter task

PR #2083, opened 2023-03-10
FoundryCreateNew emit foundry serial number as event

ivange94:
# Description of change
Emit foundry serial number after creation

### DIFF
--- a/packages/vm/core/accounts/impl.go
+++ b/packages/vm/core/accounts/impl.go
@@ -179,7 +179,7 @@ func foundryCreateNew(ctx isc.Sandbox) dict.Dict {
 
 	ret := dict.New()
 	ret.Set(ParamFoundrySN, util.Uint32To4Bytes(sn))
-	ctx.Event(fmt.Sprintf("Foundry serial number = %d", sn))
+	ctx.Event(fmt.Sprintf("Foundry created, serial number = %d", sn))
 	return ret
 }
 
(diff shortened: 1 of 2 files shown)


packages/vm/core/accounts/impl.go
```diff
@@ -178,6 +179,7 @@ func foundryCreateNew(ctx isc.Sandbox) dict.Dict {
 
 	ret := dict.New()
 	ret.Set(ParamFoundrySN, util.Uint32To4Bytes(sn))
+	ctx.Event(fmt.Sprintf("Foundry serial number = %d", sn))
```
jorgemmsilva:
```suggestion
	ctx.Event(fmt.Sprintf("Foundry created, serial number = %d", sn))
```

jorgemmsilva (approved):
nice!

packages/vm/core/accounts/impl.go
dessaya: Shouldn't we use something more efficient? for the event payload? E.g.:

```suggestion
	ctx.Event(string(util.Uint32To4Bytes(sn)))
```

So that:
* it is more space efficient
* the client can decode the value more easily

dessaya: Hmmm but then we would need to include an "event type" code.

I think we should do this at some point. Let's merge this PR as is, and I'll add an issue to standardize the format of all events

dessaya (approved):
Maybe mention the event in the documentation of `foundryCreateNew` (`documentation/docs/guide/core_concepts/core_contracts/accounts.md`)

PR #2087, opened 2023-03-10
Use ShrinkingMaps for long living maps to avoid memleaks

muXxer:


packages/kv/buffered/mutation.go
```diff
@@ -14,14 +15,14 @@ import (
 // Mutations is a set of mutations: one for each key
 // It provides a deterministic serialization
 type Mutations struct {
-	Sets map[kv.Key][]byte
-	Dels map[kv.Key]struct{}
+	Sets *shrinkingmap.ShrinkingMap[kv.Key, []byte]
+	Dels *shrinkingmap.ShrinkingMap[kv.Key, struct{}]
```
dessaya: I don't think there are any long-lasting instances of `Mutation`, so why this change?
Isn't this premature optimization? (Which we all know is the root of all evil?).

Is there a measureable improvement? Are there no drawbacks (apart from the increased complexity)? For example I see that `for range` loops now involve a function call + a closure. Isn't that less efficient?

packages/trie/kvimpl.go
```diff
@@ -9,92 +9,91 @@ import (
...
-		store  InMemoryKVStore
+		store  *InMemoryKVStore
 		prefix []byte
 	}
 )
 
-func NewInMemoryKVStore() InMemoryKVStore {
-	return make(InMemoryKVStore)
+type InMemoryKVStore struct {
+	m *shrinkingmap.ShrinkingMap[string, []byte]
```
dessaya: This thing is only used in tests, so no need for the change.

dessaya (changes requested):

packages/trie/kvimpl.go
muXxer: Cool, I can revert that change. Can we move all "test-only" related packages maybe to `packages/testutil` ?

packages/kv/buffered/mutation.go
muXxer: I wasn't 100% sure about the lifetime of the `Mutations` struct, so I also changed it.
I can revert that change if you want.

packages/trie/kvimpl.go
dessaya: I don't think it would make sense here, because this `InMemoryKVStore` only makes sense for the `trie` package; it's not useful anywhere else. It could even be package private.

packages/kv/buffered/mutation.go
dessaya: I'd say let's revert unless there is a measureable improvement.

The same goes for the rest of the PR: in general it is increasing complexity, so IMO it should be justified. but I'm not familiar with every package so I'd ask Jorge, Karolis, Julius, etc.

dessaya (approved):

packages/kv/buffered/mutation.go
muXxer: Reverted :+1:

packages/trie/kvimpl.go
muXxer: Reverted and moved to `trie/test` :+1:

packages/vm/processors/cache.go
```diff
@@ -15,14 +16,14 @@ import (
 type Cache struct {
 	mutex      *sync.Mutex
 	Config     *Config
-	processors map[hashing.HashValue]isc.VMProcessor
```
jorgemmsilva, line 18: a new processors Cache created every time a new consensus instance starts, so I wouldn't say this is exactly "long living"

packages/chain/chainMgr/chainMgr.go
```diff
@@ -154,16 +158,16 @@ type cmtLogInst struct {
...
-	consensusStateRegistry  cmtLog.ConsensusStateRegistry           // Persistent store for log indexes.
-	latestActiveCmt         *iotago.Ed25519Address                  // The latest active committee.
-	latestConfirmedAO       *isc.AliasOutputWithID                  // The latest confirmed AO (follows Active AO).
-	activeAccessNodes       []*cryptolib.PublicKey                  // All the nodes authorized for being access nodes (for the ActiveAO).
-	activeAccessNodesCB     func([]*cryptolib.PublicKey)            // Called, when a list of access nodes has changed.
-	needConsensus           *NeedConsensus                          // Query for a consensus.
-	needPublishTX           map[iotago.TransactionID]*NeedPublishTX // Query to post TXes.
-	dkShareRegistryProvider registry.DKShareRegistryProvider        // Source for DKShares.
+	chainID                 isc.ChainID                                                      // This instance is responsible for this chain.
+	cmtLogs                 map[iotago.Ed25519Address]*cmtLogInst                            // All the committee log instances for this chain.
```
jorgemmsilva, line 163: shouldn't this also be a shrinking map?

packages/chain/cmtLog/var_log_index.go
```diff
@@ -114,16 +115,16 @@ type VarLogIndex interface {
...
+	nodeIDs    []gpa.NodeID                                                 // All the peers in this committee.
+	n          int                                                          // Total number of nodes.
+	f          int                                                          // Maximal number of faulty nodes to tolerate.
+	latestAO   *isc.AliasOutputWithID                                       // Latest known AO, as reported by the varLocalView, can be nil (means we suspend proposing LIs).
+	proposedLI LogIndex                                                     // Highest LI send by this node with a ⟨NextLI, li⟩ message.
+	agreedLI   LogIndex                                                     // LI for which we have N-F proposals (when reached, consensus starts, the LI is persisted).
+	minLI      LogIndex                                                     // Minimal LI at which this node can participate (set on boot).
+	maxPeerLIs map[gpa.NodeID]*msgNextLogIndex                              // Latest peer indexes received from peers.
+	consAggrAO *shrinkingmap.ShrinkingMap[LogIndex, *isc.AliasOutputWithID] // Recent outputs, to filter L1ReplacedBaseAliasOutput.
```
jorgemmsilva: think this `varLogIndexImpl` obj is only live during 1 consensus "log index", meaning it shouldn't be "long living"

packages/chains/accessMgr/amDist/accessMgrDist.go
```diff
@@ -32,11 +31,11 @@ type Output interface {
 }
 
 type accessMgrDist struct {
-	nodes            map[gpa.NodeID]*accessMgrNode             // State for each peer.
-	chains           map[isc.ChainID]*accessMgrChain           // State for each chain.
-	pubKeyToNodeID   func(*cryptolib.PublicKey) gpa.NodeID     // Convert PubKeys to NodeIDs.
-	serversUpdatedCB func(isc.ChainID, []*cryptolib.PublicKey) // Called when a set of servers has changed for a chain.
-	dismissPeerCB    func(*cryptolib.PublicKey)                // To stop redelivery at the upper layer.
+	nodes            *shrinkingmap.ShrinkingMap[gpa.NodeID, *accessMgrNode]   // State for each peer.
+	chains           *shrinkingmap.ShrinkingMap[isc.ChainID, *accessMgrChain] // State for each chain.
```
jorgemmsilva, line 35: these shouldn't hold more than a handful of elements, I doubt its worth it to make the change

jorgemmsilva (approved):
In a handful of places this change makes sense, but using this everywhere is kinda extreme.
We're adding a bunch of mutexes and cleanup logic where it is not even clear that there is a leak

let's wait for Karolis and Julius input before merging

packages/chain/chainMgr/chainMgr.go
muXxer: I couldn't find a `delete` call for this map, so I didn't change to a `shrinkingmap`

jorgemmsilva: I guess this turns into a question to @kape1395 : shouldn't we delete the cmtLogs at some point? :sweat_smile:

PR #2124, opened 2023-03-16
feat(evm): accept non-view calls in eth_call

### LINKED ISSUE #2105: gas estimation with EVM `gasleft()` or a static call on magic contracts fails
Reproducible code: 

```
pragma solidity ^0.8.3;

import "@iscmagic/ISC.sol";

contract Gasleft {

    function test() public returns (uint256 gasLeft) {
        bytes32 e = ISC.sandbox.getEntropy();
        gasLeft = gasleft();
    }
}
```

because functions like gasLeft (or callStatic in ethers.js) do a view call on a call function to see how much gas that function would use and the magic contracts do not expect a view function it currently fails to use this functionality on the magic contract. This is giving some issues with some dApp deployments that use this functionality including `multicall`.

dessaya:
eth_call and eth_estimateGas now work in a similar fashion, executing a VM task as if it was run by the consensus. As a consequence, non-view calls can be executed, and also the block index/hash parameter is respected.

Also, the BlockInfo object is refactored to store the current and previous AliasOutput. In order to avoid resetting the testnet, the old schema is supported; the new schema is in effect only for newly created blocks.

closes: #2105


packages/evm/jsonrpc/evmchain.go
```diff
@@ -234,7 +240,55 @@ func (e *EVMChain) iscStateFromEVMBlockNumberOrHash(blockNumberOrHash rpc.BlockN
 	return e.iscStateFromEVMBlockNumber(block.Number())
 }
 
-func (e *EVMChain) Balance(address common.Address, blockNumberOrHash rpc.BlockNumberOrHash) (*big.Int, error) {
+func (e *EVMChain) iscAliasOutputFromEVMBlockNumber(blockNumber *big.Int) (*isc.AliasOutputWithID, error) {
```
jorgemmsilva, line 240: we could probably deprecate the notion of ISC/EVM block number since they are the same now, just call it "block number" everywhere

packages/evm/jsonrpc/evmchain.go
```diff
@@ -234,7 +240,55 @@ func (e *EVMChain) iscStateFromEVMBlockNumberOrHash(blockNumberOrHash rpc.BlockN
...
+	if blockNumber == nil {
+		return e.backend.ISCLatestAliasOutput()
+	}
+	if !blockNumber.IsUint64() {
+		return nil, fmt.Errorf("block number is too large: %s", blockNumber)
+	}
```
jorgemmsilva: its too large if its bigger then max uint32 :smile:

packages/state/l1commitment.go
```diff
@@ -61,6 +62,15 @@ func (bh BlockHash) Equals(other BlockHash) bool {
 	return bh == other
 }
 
+//nolint:revive,shadow
```
jorgemmsilva: why the nolint here?

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -1,26 +1,44 @@
...
+	Timestamp             time.Time
+	TotalRequests         uint16
+	NumSuccessfulRequests uint16 // which didn't panic
+	NumOffLedgerRequests  uint16
+
+	// schema < BlockInfoAliasOutputSchemaVersion
+	// TODO remove at some point?
+	previousL1CommitmentOld *state.L1Commitment  // if not nil => old schema version
+	l1CommitmentOld         *state.L1Commitment  // if old schema => nil when not known yet for the current state
+	anchorTransactionIDOld  iotago.TransactionID // of the input state
```
jorgemmsilva, line 36: these will also be removed in another PR:
- TransactionSubEssenceHash
- TotalBaseTokensInL2Accounts
- TotalStorageDeposit

I'll make a single migration to cut everything at once

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -70,119 +112,143 @@ func (bi *BlockInfo) String() string {
...
-	}
-	if err := util.WriteUint64(w, bi.GasFeeCharged); err != nil {
-		return err
-	}
-	return nil
+	mu.WriteUint64(bi.TotalBaseTokensInL2Accounts)
+	mu.WriteUint64(bi.TotalStorageDeposit)
+	mu.WriteUint64(bi.GasBurned)
+	mu.WriteUint64(bi.GasFeeCharged)
+	return mu.Bytes()
```
jorgemmsilva, line 159: would be better to split into separate functions, even if the code is repeated, otherwise this turns into a nice spaghetti when we try to add another new version

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -70,119 +112,143 @@ func (bi *BlockInfo) String() string {
...
 }
 
-func (bi *BlockInfo) Read(r io.Reader) error {
-	if err := util.ReadTime(r, &bi.Timestamp); err != nil {
-		return err
-	}
-	if err := util.ReadUint16(r, &bi.TotalRequests); err != nil {
-		return err
+//nolint:gocyclo,revive,shadow,govet
```
jorgemmsilva: why the nolint? 
its better to ignore in the particular line, rather than for the entire func IMO.

jorgemmsilva (approved):

packages/evm/jsonrpc/evmchain.go
dessaya: This only checks that the BigInt can be converted to uint64, so that then it can be converted to uint32. I don't think there is a more elegant way (without sacrificing a little performance).
I will put both checks in the same function though.

packages/evm/jsonrpc/evmchain.go
dessaya: Idk, maybe. They are still different types: BigInt vs uint32

packages/state/l1commitment.go
dessaya: `indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)`

It's a false positive. If you do the suggester refactor, then the `buf` variable is out of scope.

packages/vm/core/blocklog/blockinfo.go
dessaya: `govet/shadow`: `declaration of "err" shadows declaration at line 165`

I don't care about shadowing `err` here. Fixing this would make the code even longer than it actually is.

`revive`: `indent-error-flow` false positive

`gocyclo`: yes, it's a long ass function but it's mostly flat so I think it's ok. Anyway, I will separate it in two `fromBytesOld` and `fromBytesNew`.

dessaya: Change of plans: I tried separating it in two functions, and I still get the `gocyclo` for both of them. I'll leave it as is, we can always refactor in future PRs.

packages/evm/jsonrpc/evmchain.go
jorgemmsilva: I was thinking something like this
```
var maxBlockNumber = new(big.Int).SetUint64(uint64(math.MaxUint32))

.........

if blockNumber.Cmp(maxBlockNumber) == 1 {
		return nil, fmt.Errorf("block number is too large: %s", blockNumber)
}
```

packages/state/l1commitment.go
jorgemmsilva: something like this would work, tho:
```
	buf, err := mu.ReadBytes(L1CommitmentSize)
	if err != nil {
		return nil, err
	}
 	return L1CommitmentFromBytes(buf)
```


it literally says:
>move short variable declaration to its own line if necessary

:stuck_out_tongue:

packages/evm/jsonrpc/evmchain.go
dessaya: I like that! I'll make that change in a future PR.

packages/state/l1commitment.go
dessaya: Yeah, well, I disagree that that is better, because in that case the `buf` variable ends up having a much larger scope.

This is one of those linter warnings that I think are not really valuable. It's just cosmetic, it does not prevent bugs. Does it?

PR #2141, opened 2023-03-20
Refactor: remove sd assumptions, handles L1 SD cost changes

jorgemmsilva:


packages/util/panicutil/panicutil.go
```diff
@@ -53,8 +53,8 @@ func CatchAllButDBError(f func(), log *logger.Logger, prefix ...string) (err err
 			default:
 				err = fmt.Errorf("%s%v", s, err1)
 			}
-			log.Debugf("%s%v", s, err)
-			log.Debugf(string(debug.Stack()))
+			// wrapp the error and add the stack trace
+			err = fmt.Errorf("%w: %s", err, string(debug.Stack()))
```
dessaya: Is it ok to include the stack trace in the error object? e.g. this will never reach an end-user, or stored in a receipt?

dessaya (approved):

packages/util/panicutil/panicutil.go
jorgemmsilva: you're right. I added this because it was very useful to debug stuff in the tests. 
Will remove for now, might add later just for solo

muXxer (approved):

PR #2142, opened 2023-03-21
refactor: Remove Cargo.lock in cleanup.sh

howjmay:


contracts/wasm/scripts/cleanup.sh
```diff
@@ -3,3 +3,4 @@ contracts_path=$(git rev-parse --show-toplevel)/contracts/wasm
 cd $contracts_path
 go install ../../tools/schema
 schema -go -rs -ts -clean
+rm $contracts_path/Cargo.lock
```
muXxer: Aren't you already in the `contracts_path` during that call here?

Is that the only `cargo.lock` in the repo?

howjmay: I have checked smart contract won't generate `cargo.lock` files

BugFreeSoftware: I'd rather see this be part of schema tool. Just add a cleanSourceFile() to RustGenerator.CleanUp()

jorgemmsilva (approved):

PR #2152, opened 2023-03-22
Update WasmLib core contracts interface and docs

BugFreeSoftware:
WasmLib core contracts interface regenerated with integrated docs
Also added missing core contract docs to wiki
Bumped WasmLib version to 1.0.3
Regenerated Wasm contracts with new WasmLib


packages/wasmvm/wasmlib/interfaces/coreaccounts.yaml
```diff
@@ -4,71 +4,143 @@ structs: {}
...
+
+  # Mints or destroys tokens for the given foundry, which must be owned by the caller.
   foundryModifySupply:
     params:
-      foundrySN=s: Uint32
-      supplyDeltaAbs=d: BigInt
-      destroyTokens=y: Bool?
+      foundrySN=s: Uint32 # serial number of the foundry
+      supplyDeltaAbs=d: BigInt # positive nonzero amount to mint or destroy
+      destroyTokens=y: Bool? # mint (default) or destroy tokens
```
howjmay, line 29: Should we mention what true and false mean?

muXxer: I would say the name is self explaining :D

muXxer (approved):

packages/wasmvm/wasmlib/interfaces/coreaccounts.yaml
howjmay: okk xdd

PR #2161, opened 2023-03-24
fix: gaslimit on evm view calls

jorgemmsilva:


packages/isc/request_evm.go
```diff
@@ -211,11 +210,7 @@ func (r *evmOffLedgerCallRequest) Assets() *Assets {
 }
 
 func (r *evmOffLedgerCallRequest) GasBudget() (gas uint64, isEVM bool) {
-	if r.callMsg.Gas > 0 {
-		return r.callMsg.Gas, true
-	}
-	// see VMContext::calculateAffordableGasBudget() when EstimateGasMode == true
-	return math.MaxUint64, false
+	return r.callMsg.Gas, true
```
muXxer, line 213: Before the change, the return value `isEVM` was false in case of `callMsg.Gas == 0`.
This behaviour changed now... Is it still correct?

packages/vm/vmcontext/runreq.go
```diff
@@ -256,7 +256,7 @@ func (vmctx *VMContext) callFromRequest() dict.Dict {
 
 func (vmctx *VMContext) getGasBudget() uint64 {
 	gasBudget, isEVM := vmctx.req.GasBudget()
-	if !isEVM {
+	if !isEVM || gasBudget == 0 {
```
muXxer, line 259: What would happen if someone passes a regular non-view call with `gasBudget == 0`?

tools/local-setup/docker-compose.yml
```diff
@@ -141,7 +141,7 @@ services:
       - "traefik.http.routers.wasp-dashboard.middlewares=rewrite-wasp-dashboard"
       - "traefik.http.middlewares.rewrite-wasp-dashboard.stripprefix.prefixes=/wasp/dashboard"
     environment:
-      - WASP_API_URL=http://localhost/wasp/api
+      - WASP_API_URL=http://localhost/wasp/api/v1
```
muXxer: thats wrong. the new dashboard container already has the correct routes. you only need to `docker compose pull`

packages/isc/request_evm.go
jorgemmsilva: yes, we previously relied on gas being math.MaxUInt64 to identify "estimate gas runs that want to emulate max-tokens", now that behavior is for 0, which is in line of what view-calls want to archive (run with maxGas, not bound by the amount of tokens the user has to pay for the gas)

packages/vm/vmcontext/runreq.go
jorgemmsilva: get gas budget 0 :stuck_out_tongue: (that will be automatically adjusted to `minGas` later)

muXxer (approved):

PR #2175, opened 2023-03-26
Push the active state to access nodes.

kape1395:
The active state on the access nodes is needed to make responses to the EVM calls faster. They are replied to the call as processed when the request is stored in the block log and is present at least in the active state. Till this change, the caller on the access node had to wait for the L1 confirmation before getting the response. That meant waiting for the milestones.


packages/chain/chainMgr/chainMgr.go
```diff
@@ -160,15 +159,17 @@ type cmtLogInst struct {
...
 	cmtLogs                 map[iotago.Ed25519Address]*cmtLogInst                            // All the committee log instances for this chain.
 	consensusStateRegistry  cmtLog.ConsensusStateRegistry                                    // Persistent store for log indexes.
 	latestActiveCmt         *iotago.Ed25519Address                                           // The latest active committee.
 	latestConfirmedAO       *isc.AliasOutputWithID                                           // The latest confirmed AO (follows Active AO).
-	activeAccessNodes       []*cryptolib.PublicKey                                           // All the nodes authorized for being access nodes (for the ActiveAO).
-	activeAccessNodesCB     func([]*cryptolib.PublicKey)                                     // Called, when a list of access nodes has changed.
+	activeNodesCB           func() ([]*cryptolib.PublicKey, []*cryptolib.PublicKey)          // All the nodes authorized for being access nodes (for the ActiveAO).
+	tractStateActCB         func(ao *isc.AliasOutputWithID)                                  // We will call this to set new AO for the active state.
```
jorgemmsilva, line 168: renaming this to something like `trackActiveStateCB` would be easier to grok.

packages/chain/chainMgr/var_access_state.go
```diff
@@ -0,0 +1,271 @@
...
+	"github.com/iotaledger/hive.go/logger"
+	iotago "github.com/iotaledger/iota.go/v3"
+	"github.com/iotaledger/wasp/packages/isc"
+)
+
+// Tracks the active state at the access nodes. If this node is part of the committee,
+// then the tip tracked by this node should be ignored and the state tracked by the
+// committee should be used. The algorithm itself is similar to the `varLocalView`
+// in the `cmtLog`.
+type VarAccessState interface {
```
jorgemmsilva, line 18: could be renamed to something more descriptive like: `AccessNodesState` or `VarAccessNodesState`

packages/chain/chainMgr/chainMgr.go
```diff
@@ -314,10 +326,26 @@ func (cmi *chainMgrImpl) handleInputChainTxPublishResult(input *inputChainTxPubl
...
 	cmi.log.Debugf("handleInputConsensusOutputDone: %+v", input)
+	msgs := gpa.NoMessages()
 	// >     IF ConsensusOutput.BaseAO == NeedConsensus THEN
 	// >         Add ConsensusOutput.TX to NeedPublishTX
 	if true { // TODO: Reconsider this condition. Several recent consensus instances should be published, if we run consensus instances in parallel.
 		txID := input.consensusResult.NextAliasOutput.TransactionID()
+		if !cmi.needPublishTX.Has(txID) && input.consensusResult.Block != nil {
+			// Inform the access nodes on new block produced.
+			block := input.consensusResult.Block
+			activeAccessNodes, activeCommitteeNodes := cmi.activeNodesCB()
```
jorgemmsilva, line 337: this is a bit odd. shouldn't `cmi.activeNodesCB()` be a"callback" function? It's not used like that here

packages/chain/chainMgr/chainMgr.go
```diff
@@ -371,6 +389,16 @@ func (cmi *chainMgrImpl) handleMsgCmtLog(msg *msgCmtLog) gpa.OutMessages {
 	})
 }
 
+func (cmi *chainMgrImpl) handleMsgBlockProduced(msg *msgBlockProduced) gpa.OutMessages {
+	cmi.log.Debugf("handleMsgBlockProduced: %+v", msg)
+	// TODO: Save the block <------ this...
```
jorgemmsilva, line 394: this is done already, no?

packages/chain/chainMgr/var_access_state.go
```diff
@@ -0,0 +1,271 @@
...
+		return vas.value, true
+	}
+	if vas.value.Equals(newTip) {
+		vas.log.Debugf("⊳ Tip remains %v.", vas.value)
+		return vas.value, false
+	}
+	vas.log.Debugf("⊳ New tip=%v, was %v", newTip, vas.value)
+	vas.value = newTip
+	return vas.value, true
+}
```
jorgemmsilva, line 139: it would be nice to write documentation for what these `bool` return values mean, so its easy to revisit later without going though the logic

packages/chain/chainMgr/var_access_state.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	consumed, published, err := vas.extractConsumedPublished(tx)
+	if err != nil {
+		vas.log.Debugf("BlockProduced(tx.ID=%v): Ignoring because of %v", txID, err)
+		return vas.value, false
+	}
+	//
+	vas.log.Debugf("BlockProduced: consumed.ID=%v, published=%v", consumed.ToHex(), published)
+	stateIndex := published.GetStateIndex()
+	// prevLatest := vas.findLatestPending()
```
jorgemmsilva, line 65:
```suggestion
```

packages/chain/chainMgr/var_access_state.go
```diff
@@ -0,0 +1,271 @@
...
+// in the `cmtLog`.
+type VarAccessState interface {
+	Value() *isc.AliasOutputWithID
+	BlockProduced(tx *iotago.Transaction) (*isc.AliasOutputWithID, bool)
+	BlockConfirmed(ao *isc.AliasOutputWithID) (*isc.AliasOutputWithID, bool)
+}
+
+type varAccessStateImpl struct {
+	chainID   isc.ChainID
+	value     *isc.AliasOutputWithID                                     // Will point to the latest known good state while the chain don't have the current good state.
```
jorgemmsilva, line 26: what do you think about calling this something like `LastKnownConfirmedAO` or something like that? 
`value` is too generic imo

jorgemmsilva (approved):

packages/chain/chainMgr/chainMgr.go
kape1395: Done.

packages/chain/chainMgr/var_access_state.go
kape1395: Renamed to `varAccessNodesState`.

packages/chain/chainMgr/chainMgr.go
kape1395: But that's actually a callback. This function is passed via the constructor and assigned to a field.

packages/chain/chainMgr/chainMgr.go
kape1395: Not yet. But that's not critical.

packages/chain/chainMgr/var_access_state.go
kape1395: Added docs, e.g.
```
	// Considers the produced (not yet committed) block / TX and returns new tip AO,
	// and boolean indicating if the tip has changed because of this call.
	BlockProduced(tx *iotago.Transaction) (*isc.AliasOutputWithID, bool)
```

packages/chain/chainMgr/var_access_state.go
kape1395: Thanks, removed it.

packages/chain/chainMgr/var_access_state.go
kape1395: Renamed it to `tipAO`.

PR #2195, opened 2023-03-28
fix: Fix AccountFoundries and add test

howjmay:


packages/vm/core/accounts/impl_views.go
```diff
@@ -82,7 +82,7 @@ func viewAccountFoundries(ctx isc.SandboxView) dict.Dict {
 	foundries := accountFoundriesMapR(ctx.StateR(), account)
 	ret := dict.New()
 	foundries.MustIterate(func(k []byte, v []byte) bool {
-		ret.Set(kv.Key(k), v)
+		ret.Set(kv.Key(k), []byte{0xff})
```
dessaya: Just curious, why this change?

contracts/wasm/corecontracts/test/core_accounts_test.go
```diff
@@ -257,6 +257,52 @@ func TestFoundryModifySupply(t *testing.T) {
...
+		require.Equal(t, []byte{0xff}, f.Results.Foundries().GetBytes(foundry.SN()).Value())
+	}
+}
+
+func TestAccountNFTAmount(t *testing.T) {}
+
+func TestAccountNFTAmountInCollection(t *testing.T) {}
+
+func TestAccountNFTsInCollection(t *testing.T) {}
+
```
dessaya: Are these left empty intentionally?

packages/vm/core/accounts/impl_views.go
howjmay: I am referring the wiki
https://wiki.iota.org/shimmer/smart-contracts/guide/core_concepts/core_contracts/accounts#accountfoundriesa-agentid

contracts/wasm/corecontracts/test/core_accounts_test.go
howjmay: Yes. I will finish them in another PR

packages/vm/core/accounts/impl_views.go
dessaya: But the `v` value already should contain `0xff`. If it does not, then we have a bug somewhere else.

contracts/wasm/corecontracts/test/core_accounts_test.go
dessaya: Can you remove them in this PR? Or add a comment explaining why there are empty? Otherwise it may be confusing.

packages/vm/core/accounts/impl_views.go
dessaya: Just to clarify, I think it is better to leave `v` here because in the future we might decide how a `true` value is encoded (it happened recently).

howjmay: It was `0x1` there. Let me check what cause it

contracts/wasm/corecontracts/test/core_accounts_test.go
howjmay: ok

packages/vm/core/accounts/impl_views.go
howjmay: done

PR #2208, opened 2023-03-28
feat(evm): support debug_traceTransaction (with callTracer)

### LINKED ISSUE #2104: Support `debug_traceTransactions` in JSON/RPC endpoint for Internal Transaction fetching in EVM (allows you to see error messages and internal transactions in the explorer)
According to https://docs.blockscout.com/for-developers/information-and-settings/node-tracing-json-rpc-requirements we need this functionality to allow internal transaction fetching. This is more relevant now given users facing a error message now have no option to see that actual error message in the explorer.

dessaya:
closes: #2104


packages/chainutil/evmcall.go
```diff
@@ -23,7 +23,7 @@ func Call(ch chain.ChainCore, aliasOutput *isc.AliasOutputWithID, call ethereum.
 	}
 
 	iscReq := isc.NewEVMOffLedgerCallRequest(ch.ID(), call)
-	res, err := executeISCVM(ch, aliasOutput, iscReq)
+	res, err := runISCReqeust(ch, aliasOutput, iscReq)
```
muXxer:
```suggestion
	res, err := runISCRequest(ch, aliasOutput, iscReq)
```

packages/chainutil/evmestimategas.go
```diff
@@ -74,7 +43,7 @@ func EstimateGas(ch chain.ChainCore, aliasOutput *isc.AliasOutputWithID, call et
 	executable := func(gas uint64) (failed bool, used uint64, err error) {
 		call.Gas = gas
 		iscReq := isc.NewEVMOffLedgerCallRequest(ch.ID(), call)
-		res, err := executeISCVM(ch, aliasOutput, iscReq)
+		res, err := runISCReqeust(ch, aliasOutput, iscReq)
```
muXxer:
```suggestion
		res, err := runISCRequest(ch, aliasOutput, iscReq)
```

packages/chainutil/runvm.go
```diff
@@ -0,0 +1,58 @@
...
+	evmTracer *isc.EVMTracer,
+) ([]*vm.RequestResult, error) {
+	vmRunner := runvm.NewVMRunner()
+	task := &vm.VMTask{
+		Processors:           ch.Processors(),
+		AnchorOutput:         aliasOutput.GetAliasOutput(),
+		AnchorOutputID:       aliasOutput.OutputID(),
+		Store:                ch.Store(),
+		Requests:             reqs,
+		TimeAssumption:       time.Now(),
```
muXxer: Before this code was part of `estimategas_evm.go`, where it is fine to run the task with the current time I guess.
But now this is part of the `chainutil` and I wouldn't have expected that the `runISCTask` executes with local current time....
Should we maybe pass the time to `runISCTask` or at least write a comment somewhere?

packages/chainutil/runvm.go
```diff
@@ -0,0 +1,58 @@
...
+		Log:                  ch.Log().Desugar().WithOptions(zap.AddCallerSkip(1)).Sugar(),
+	}
+	err := vmRunner.Run(task)
+	if err != nil {
+		return nil, err
+	}
+	return task.Results, nil
+}
+
+func runISCReqeust(
```
muXxer:
```suggestion
func runISCRequest(
```

packages/evm/jsonrpc/waspevmbackend.go
```diff
@@ -93,11 +94,15 @@ func (b *WaspEVMBackend) evictWhenExpired(txHash common.Hash) {
...
+	return chainutil.EVMCall(b.chain, aliasOutput, callMsg)
 }
 
 func (b *WaspEVMBackend) EVMEstimateGas(aliasOutput *isc.AliasOutputWithID, callMsg ethereum.CallMsg) (uint64, error) {
-	return chainutil.EstimateGas(b.chain, aliasOutput, callMsg)
+	return chainutil.EVMEstimateGas(b.chain, aliasOutput, callMsg)
+}
+
+func (b *WaspEVMBackend) EVMTraceTransaction(aliasOutput *isc.AliasOutputWithID, iscRequestsInBlock []isc.Request, txIndex uint64, tracer tracers.Tracer) error {
+	return chainutil.EVMTraceTransaction(b.chain, aliasOutput, iscRequestsInBlock, txIndex, tracer)
```
muXxer: Do we need to be able to pass the `TimeAssumption` to maybe track errors that are related to the time?

packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -83,6 +84,17 @@ func (bctx *blockContext) mintBlock() {
 	bctx.emu.MintBlock()
 }
 
+func getTracer(ctx isc.Sandbox, bctx *blockContext) tracers.Tracer {
+	tracer := ctx.EVMTracer()
+	if tracer == nil {
+		return nil
+	}
+	if int(tracer.TxIndex) != len(bctx.txs) {
+		return nil
```
muXxer, line 94: Should we maybe return an error in that case instead of no tracer?

packages/vm/core/evm/evmtest/evm_test.go
```diff
@@ -1774,3 +1776,29 @@ func TestEVMCallViewGas(t *testing.T) {
...
+		require.NoError(t, err)
+		return traceLatestTx()
+	}()
+	trace2 := func() string {
+		iscTest := env.deployISCTestContract(ethKey)
+		_, err := iscTest.triggerEvent("Hi from EVM!")
+		require.NoError(t, err)
+		return traceLatestTx()
+	}()
+	require.NotEqualValues(t, trace1, trace2)
```
muXxer: Can we exactly match the expected results of each trace, or is that non-deterministic?

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+		Gas:   uintToHex(gas),
+		Value: bigToHex(value),
+	}
+	if create {
+		t.callstack[0].Type = "CREATE"
+	}
+}
+
+// CaptureEnd is called after the call finishes to finalize the tracing.
+func (t *callTracer) CaptureEnd(output []byte, gasUsed uint64, _ time.Duration, err error) {
```
muXxer, line 79: Will the duration be added later at some point?

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+		if err.Error() == "execution reverted" && len(output) > 0 {
+			t.callstack[0].Output = bytesToHex(output)
+		}
+	} else {
+		t.callstack[0].Output = bytesToHex(output)
+	}
+}
+
+// CaptureState implements the EVMLogger interface to trace a single step of VM execution.
+func (t *callTracer) CaptureState(pc uint64, op vm.OpCode, gas, cost uint64, scope *vm.ScopeContext, rData []byte, depth int, err error) {
```
muXxer, line 92: Will this be added later?
Maybe add a comment if not.

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// CaptureFault implements the EVMLogger interface to trace an execution fault.
+func (t *callTracer) CaptureFault(pc uint64, op vm.OpCode, gas, cost uint64, _ *vm.ScopeContext, depth int, err error) {
```
muXxer, line 96: Will this be added later?
Maybe add a comment if not.

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// CaptureEnter is called when EVM enters a new scope (via call, create or selfdestruct).
+func (t *callTracer) CaptureEnter(typ vm.OpCode, from common.Address, to common.Address, input []byte, gas uint64, value *big.Int) {
+	if t.config.OnlyTopCall {
+		return
+	}
+	// Skip if tracing was interrupted
+	if atomic.LoadUint32(&t.interrupt) > 0 {
```
muXxer, line 105: Can we use the new atomic bool instead?

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// Stop terminates execution of the tracer at the first opportune moment.
+func (t *callTracer) Stop(err error) {
+	t.reason = err
+	atomic.StoreUint32(&t.interrupt, 1)
+}
+
+func bytesToHex(s []byte) string {
+	return "0x" + common.Bytes2Hex(s)
```
muXxer, line 172: We could also use `iotago.EncodeHex`.

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+	} else {
+		call.Error = err.Error()
+		if call.Type == "CREATE" || call.Type == "CREATE2" {
+			call.To = ""
+		}
+	}
+	t.callstack[size-1].Calls = append(t.callstack[size-1].Calls, call)
+}
+
+func (*callTracer) CaptureTxStart(gasLimit uint64) {}
```
muXxer, line 148: Will this be added later?
Maybe add a comment if not.

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (*callTracer) CaptureTxEnd(restGas uint64) {}
```
muXxer, line 150: Will this be added later?
Maybe add a comment if not.

packages/evm/jsonrpc/tracer_call.go
```diff
@@ -0,0 +1,186 @@
...
+
+// GetResult returns the json-encoded nested list of call traces, and any
+// error arising from the encoding or forceful termination (via `Stop`).
+func (t *callTracer) GetResult() (json.RawMessage, error) {
+	if len(t.callstack) != 1 {
```
muXxer, line 155: So this `GetResult` only works if the callstack is exactly size 1?

muXxer (changes requested):

packages/chainutil/runvm.go
dessaya: Nice catch! We can use the timestamp that is stored for each block

packages/evm/jsonrpc/tracer_call.go
dessaya: Idk. This file is copied verbatim from go-ethereum :sweat_smile:

packages/evm/jsonrpc/tracer_call.go
dessaya: This is probably implementing an interface, and for this specific implementation no action is needed.

packages/evm/jsonrpc/tracer_call.go
dessaya: I'd rather not change this file as it is 99% copied from go-ethereum.

packages/vm/core/evm/evmimpl/internal.go
dessaya: No, `nil` here means that some transaction in this block is being traced, but it is not the current one.

muXxer: Ah ok cool, can you add that as a comment, because I will have forgotten this tomorrow :D

PR #2218, opened 2023-03-29
fix: log mutations as hex

jorgemmsilva:


packages/kv/item.go
```diff
@@ -24,5 +26,5 @@ func (items Items) Swap(i, j int) {
 }
 
 func (it *Item) Format(format string) string {
-	return fmt.Sprintf(format, it.Key, it.Value)
+	return fmt.Sprintf(format, it.Key, iotago.EncodeHex(it.Value))
```
dessaya: The key may also be non-ascii.

Also, another option is to use `%q` in the format string.

BugFreeSoftware (approved):

PR #2226, opened 2023-03-29
evm: add limits for eth_getLogs

dessaya:


jorgemmsilva (approved):
Some tests for the limits would be nice, but it LGTM!

PR #2228, opened 2023-03-30
Document Test2Chains extensively

BugFreeSoftware:


dessaya (approved):

jorgemmsilva (approved):
IMO that 2chains_test should be unified with the one in sbtests. No point keeping this duplicated.

nice descriptions 👍

Issue #2076, opened 2023-03-08
Properly urldecode the agentIDs in the API


lmoe:
The agentID parameter is not properly handled when using agentIds with "00000@tst..." and need to be urldecoded. 

Diego suggested a middleware that does it automatically. Currently all of those agentids are causing a Bad Request, so it should be solved in a timely manner.

PR #2232, opened 2023-03-31
Fix: get balance receives URLencoded param from webapi. It does not r…

davinat0r:
When requesting L2 account balances (hname@addr) from the wasp-dashboard the decoding of the AgentID would fail because it receives the URLencoded parameter. For example:

wasp-dashboards requests balance for 000000@L1addr --> the wasp node tries to retrieve the balance of 000000%40L1addr.

This is what the server returns:
 ```
{
     "Message": "Invalid property: agentID",
     "Error": "NewAgentIDFromString: wrong format"
}
```

This has been tested on:
`wasp version 0.6.0-alpha.7`
`wasp-cli version 0.6.0-alpha.7`

### DIFF
--- a/packages/webapi/params/decoder.go
+++ b/packages/webapi/params/decoder.go
@@ -47,6 +47,10 @@ func DecodeHNameFromHNameHexString(e echo.Context, key string) (isc.Hname, error
 }
 
 func DecodeAgentID(e echo.Context) (isc.AgentID, error) {
+	var tmp_agent string = e.Param(ParamAgentID)
+	if strings.Contains(tmp_agent, "%40") == true {
+		tmp_agent = strings.Replace(tmp_agent, "%40", "@", 1)
+	}
 	agentID, err := isc.NewAgentIDFromString(e.Param(ParamAgentID))
 	if err != nil {
 		return nil, apierrors.InvalidPropertyError(ParamAgentID, err)
--- a/packages/webapi/params/decoder.go
+++ b/packages/webapi/params/decoder.go
@@ -2,6 +2,7 @@ package params
 
 import (
 	"strconv"
+	"strings"
 
 	"github.com/labstack/echo/v4"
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/webapi/params/decoder.go [hunk 1]
```diff
@@ -47,7 +48,11 @@ func DecodeHNameFromHNameHexString(e echo.Context, key string) (isc.Hname, error
 }
 
 func DecodeAgentID(e echo.Context) (isc.AgentID, error) {
-	agentID, err := isc.NewAgentIDFromString(e.Param(ParamAgentID))
+	var tmp_agent string = e.Param(ParamAgentID)
+	if strings.Contains(tmp_agent, "%40") == true {
+		tmp_agent = strings.Replace(tmp_agent, "%40", "@", 1)
+	}
+	agentID, err := isc.NewAgentIDFromString(tmp_agent)
```
dessaya, line 55: Thank you for the contribution @ScovottoDavide !

Unfortunately I don't think this is the correct solution, since it does not fix the root cause. The URL-encoded parameter should be decoded by the http engine, and for some reason that's not the case. We have to investigate that (#2076).

dessaya (changes requested):

packages/webapi/params/decoder.go
davinat0r: @dessaya Yes that is odd. 
Maybe a generic `url.PathUnescape(e.Param(ParamAgentID))` could temporary solve the problem. I'm locally running a wasp node like this. 
Anyway, thank you!

tools/local-setup/docker-compose.yml
```diff
@@ -16,61 +22,14 @@ services:
     volumes:
       - "/var/run/docker.sock:/var/run/docker.sock:ro"
 
-  ###############
```
jorgemmsilva: why is this PR removing hornet-nest from here?

jorgemmsilva (changes requested):

tools/local-setup/docker-compose.yml
davinat0r: @jorgemmsilva Sorry, my bad. I modified the compose because I already had a private tangle running. I replaced it with the original yaml file.

jorgemmsilva:
Closing this as it was already fixed by @lmoe .

Thank you @ScovottoDavide for bringing this issue up :raised_hands:

PR #2246, opened 2023-04-02
test: Add corecontract tests

howjmay:


packages/vm/core/accounts/impl_views.go
```diff
@@ -145,11 +147,23 @@ func viewAccountNFTsInCollection(ctx isc.SandboxView) dict.Dict {
...
+		}
+
+		if nft.Issuer.String() == collectionID.String() && bytes.Equal(nft.Owner.Bytes(), aid.Bytes()) {
+			cnt++
+		}
+		return true
+	})
 	return dict.Dict{
-		ParamNFTAmount: codec.EncodeUint32(nftsByCollectionMapR(ctx.StateR(), aid, kv.Key(collectionID[:])).Len()),
+		ParamNFTAmount: codec.EncodeUint32(cnt),
```
dessaya: Why this change?
The current implementation is O(1) and this will make it O(n).

packages/vm/core/accounts/impl_views.go
```diff
@@ -145,11 +147,23 @@ func viewAccountNFTsInCollection(ctx isc.SandboxView) dict.Dict {
 }
 
 func viewAccountNFTAmountInCollection(ctx isc.SandboxView) dict.Dict {
-	params := ctx.Params()
-	aid := params.MustGetAgentID(ParamAgentID)
-	collectionID := codec.MustDecodeNFTID(params.Get(ParamCollectionID))
+	aid := ctx.Params().MustGetAgentID(ParamAgentID)
+	collectionID := codec.MustDecodeNFTID(ctx.Params().Get(ParamCollectionID))
```
BugFreeSoftware: Try to cache values from the sandbox. ctx.Params() is a prime candicate for that. It is faster and it reduces gas fees.
Also, the Params has the kvdecoder built in, so no need for codec.MustDecodexxx. You can simply do params.MustDecodexxx like you do in the previous line.

packages/wasmvm/wasmlib/go/wasmlib/wasmtypes/scagentid.go
```diff
@@ -92,10 +92,11 @@ func AgentIDFromBytes(buf []byte) (a ScAgentID) {
 		a.address = ChainIDFromBytes(buf[:ScChainIDLength]).Address()
 		a.hname = HnameFromBytes(buf[ScChainIDLength:])
 	case ScAgentIDEthereum:
+		// buf doesn't contain Kind byte
 		if len(buf) != ScLengthEth {
 			panic("invalid AgentID length: eth agentID")
 		}
-		a.address = AddressFromBytes(buf)
+		a.address = AddressFromBytes(append([]byte{ScAddressEth}, buf...))
```
BugFreeSoftware, line 99: AddressFromBytes will already correctly make an ETH ScAddress with type byte.
You may want to pull the latest develop or wasmclient into your PR. 
The scagentid.go as is is correct and passes all tests.

packages/vm/core/accounts/impl_views.go
howjmay: But the current implementation didnt really set the NFT with collection successfully, so with the previous implementation, I retrieve `nil`. I don't know how to fix this, so I presented this workaround temporarily

dessaya: Can you clarify? What do you mean by "set the NFT with collection"? Where do you retrieve nil?

BugFreeSoftware (changes requested):

packages/vm/core/accounts/impl_views.go
howjmay: Thanks @dessaya for informing. The tests passed now. I reverted the change here

packages/vm/core/accounts/impl_views.go
howjmay: Such a func doesn't exist in `KVDecoder` interface. Let me add it in this PR

PR #2249, opened 2023-04-03
chore: bump statemetadata version

jorgemmsilva:


kape1395:
Should we check that nicely somewhere?

kape1395 (approved):

Issue #2088, opened 2023-03-10
Command to verify magic contracts with explorer


fijter:
The current magic contracts are not verified with the explorer (this is a manual/api action), it would be nice to do so if there is a explorer so you can directly see the functionality in the explorer and interact with them (see https://explorer.evm.testnet.shimmer.network/api-docs, look for verify). In order to do this we need to send some requests with the right info to the API of an explorer. This could be a wasp-cli task that takes an API URL and that will call them all in a row. The verification arguments need to be exactly matching in order for the verification to succeed including exact solidity compiler version etc. so we might want to keep track of that info somewhere when the chain is deployed.

fijter:
Explorer does not like magic contracts

PR #2251, opened 2023-04-03
WIP: Verify contracts with the CLI

luke-thorne:
# Description of change

Add functionality for the CLI to parse contract imports and upload a given contract and all imports to Blockscout for verification. Also adds a command to check on the status of the contract verification. 

#2088 

## Type of change

I deployed several ERC20 contracts that have external imports included (OpenZeppelin library) and was able to successfully verify the contract code on Blockscout via the CLI.


tools/wasp-cli/contract/cmd.go
```diff
@@ -0,0 +1,120 @@
...
+	var compilerVersion string
+	var optimization bool
+	var constructorArguments string
+	var autoDetectConstructorArguments bool
+	var evmVersion string
+	var optimizationRuns int
+	var importRemaps []string
+
+	cmd := &cobra.Command{
+		Use:   "verify <blockscoutAPIAddress> <addressHash> <name>  <contractSourceCodeFilePath> [--args]",
```
fijter, line 55: It's cool that it is so generic, however I do think it becomes a really annoying chore to manually input all the by default deployed magic contracts every time you spawn a chain, maybe a additional command could be added that loops over those default contracts so that you only have to enter the blockscout API address and the loop takes care of the addresses (those are hardcoded for the default magic contracts), names and sources. For this we also need to take the compiler version into consideration at time of starting the chain so that has to be stored somewhere.

luke-thorne: The compiled magic contracts are stored in the wasp binary and generated at build time, right? So the compiler version used at build time is the version to use in wasp? 

As for hardcoding the magic contracts information, I rarely use the wasp-cli from the root of the wasp repo so I'm not positive how to hardcode the path to the contract files. I can hardcode the addresses of the magic contracts but I tried to built it so that the user can specify an import mapping like `--import-remapping @iscmagic=https://raw.githubusercontent.com/iotaledger/wasp/v0.6.2` and let the CLI grab any relevant imports based on that mapping (which includes the version tag used so the contract code matches the compiled bytecode.

The best UX I can think of is to have a command like `wasp-cli contract verify-magic-contracts --import-remapping @iscmagiv=https://raw.githubusercontent.com/iotaledger/wasp/v0.6.2` that would just have a hardcoded list of contracts to iterate through and submit for verification using the generic method built for the `verify` command.

PR #2253, opened 2023-04-04
feat: Add subprojects into parent Cargo.toml under schema -init

howjmay:


BugFreeSoftware (changes requested):
Not quite how I meant it.
This will add the project at init whether the user will decide to have a Rust SC or not.
If not, then the workspace will be wrong, because the subfolders will not exist.
Can you change it so that it will add any missing subfolders whenever schema -rs is run?

howjmay:
Moved!

BugFreeSoftware (approved):

PR #2257, opened 2023-04-04
Save preliminary blocks in the WAL.

kape1395:
This will make the sync of the active state on the access nodes faster.
The nodes share the blocks they produce and send the corresponding blocks along.
If the block is unseen, the node will save it to the WAL to find it faster later, if needed.


packages/chain/statemanager/state_manager.go
```diff
@@ -300,6 +337,21 @@ func (smT *stateManager) handleNodePublicKeys(req *reqChainNodesUpdated) {
 	smT.nodeRandomiser.UpdateNodeIDs(peerNodeIDs)
 }
 
+func (smT *stateManager) handlePreliminaryBlock(msg *reqPreliminaryBlock) {
+	if !smT.wal.Contains(msg.block.Hash()) {
```
jorgemmsilva, line 341: if could be inverted to reduce nesting

jorgemmsilva (approved):

PR #2275, opened 2023-04-05
fix(evm): ws publish blocks in order

dessaya:


kape1395 (changes requested):
We should solve that in the chain itself.

PR #2279, opened 2023-04-06
evm: publish new blocks in a single separate goroutine

dessaya:


packages/evm/jsonrpc/evmchain.go
```diff
@@ -54,47 +56,61 @@ func NewEVMChain(backend ChainBackend, pub *publisher.Publisher, log *logger.Log
...
 
 	return e
 }
 
+func (e *EVMChain) publishNewBlock(blockIndex uint32) {
+	state, err := e.backend.ISCStateByBlockIndex(blockIndex)
+	if err != nil {
+		log.Errorf("NewEVMChain.Hook.ISCStateByBlockIndex(blockIndex=%v) returned error: %v", blockIndex, err)
+		return
+	}
```
kape1395, line 84: The publisher provides a block. It includes a trie root. Maybe we can get the state by trie root faster instead of looking it up by index again?

kape1395 (approved):

packages/evm/jsonrpc/evmchain.go
dessaya: The BlockInfo contains the *previous* trie root but not the current one. Does the publisher provide the trie root in some other way?

kape1395: Yes, the publisher receives `state.Block`, which contains `TrieRoot()`.

dessaya: Ok, I'll try to improve it in a future PR. Thx!

PR #2308, opened 2023-04-11
refactor: Use MustWaspAPIURL for most of places

howjmay:


tools/wasp-cli/cli/config/config.go
```diff
@@ -100,7 +104,7 @@ func WaspAPIURL(nodeName string) string {
 func NodeAPIURLs(nodeNames []string) []string {
 	hosts := make([]string, 0)
 	for _, nodeName := range nodeNames {
-		hosts = append(hosts, WaspAPIURL(nodeName))
+		hosts = append(hosts, MustWaspAPIURL(nodeName))
```
howjmay, line 107: I think this may push an extra space in the slice, which may not be a good result we want to see

tools/wasp-cli/cli/config/config.go
```diff
@@ -78,7 +78,11 @@ func L1FaucetAddress() string {
 }
 
 func GetToken(node string) string {
-	return viper.GetString(fmt.Sprintf("authentication.wasp.%s.token", node))
+	token := viper.GetString(fmt.Sprintf("authentication.wasp.%s.token", node))
+	if token == "" {
+		log.Fatalf("wasp node '%s' doesn't exist in config", node)
```
dessaya: This error message can be misleading. The wasp node may exist but not have a `token`.

```suggestion
		log.Fatalf("token for wasp node '%s' not defined in config", node)
```

howjmay: done

dessaya (approved):

PR #2331, opened 2023-04-14
Leaky mempool missing-requests fixed.

### LINKED ISSUE #2324: Mempool missing requests increase over time.

kape1395:
Closes https://github.com/iotaledger/wasp/issues/2324.

### DIFF
--- a/packages/chain/mempool/distSync/distSync.go
+++ b/packages/chain/mempool/distSync/distSync.go
@@ -4,9 +4,11 @@
 package distSync
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 
+	"github.com/samber/lo"
 	"golang.org/x/exp/slices"
 
 	"github.com/iotaledger/hive.go/ds/shrinkingmap"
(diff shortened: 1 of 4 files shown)


packages/chain/mempool/distSync/distSync.go
```diff
@@ -203,13 +216,20 @@ func (dsi *distSyncImpl) handleInputTimeTick() gpa.OutMessages {
 	msgs := gpa.NoMessages()
 	nodePerm := dsi.rnd.Perm(nodeCount)
 	counter := 0
-	dsi.needed.ForEach(func(_ isc.RequestRefKey, reqRef *isc.RequestRef) bool { // Access is randomized.
+	dsi.needed.ForEach(func(reqRefKey isc.RequestRefKey, reqNeeded *distSyncReqNeeded) bool { // Access is randomized.
+		stillNeeded := lo.ContainsBy(reqNeeded.waiters, func(ctx context.Context) bool { return ctx.Err() == nil })
+		if !stillNeeded {
+			dsi.needed.Delete(reqRefKey)
```
dessaya, line 222: Does deleting inside ForEach not break the loop?

dessaya (approved):

packages/chain/mempool/distSync/distSync.go
kape1395: It uses a map inside, and it is safe to delete map entries while iterating over it, afaik.
E.g. https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop

Juliusan (commented):
Looks good to me

PR #2350, opened 2023-04-18
Cleanup and small fixes to schema tool

BugFreeSoftware:


tools/schema/generator/rstemplates/cargotoml.go
```diff
@@ -8,7 +8,7 @@ var cargoToml = map[string]string{
 	"../Cargo.toml": `
 [package]
 $#emit package$cargoMain
-license = $license
+license = "$license"
```
howjmay, line 11: I was thinking should we support the string in yaml both with/without quotation marks?
This change will force the user can't use quotation marks

BugFreeSoftware: Quotation marks are stripped off by yaml anyway. So BlaBla without quotes ends up as the same string as "BlaBla" with quotes.

howjmay (approved):

tools/schema/generator/rstemplates/cargotoml.go
howjmay: got it!

PR #2362, opened 2023-04-19
dbinspector: add diff functionality

dessaya:
This computes the diff between two trie roots, in theory allowing to
identify the nodes that can be garbage collected.

Example output:

```
Diff between blocks #100 -> #200

amount of nodes: 1551 -> 4669 (+3118)
in common: 1192 (76.85% of #100) (25.53% of #200)
only on #100: 359 (23.15%)
only on #200: 3477 (74.47%)

Elapsed: 205.175732ms
Speed: 30315 nodes/s
```

### DIFF
--- a/tools/dbinspector/trie-diff.go
+++ b/tools/dbinspector/trie-diff.go
@@ -1,0 +1,134 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/iotaledger/hive.go/kvstore"
+	"github.com/iotaledger/wasp/packages/chaindb"
+	"github.com/iotaledger/wasp/packages/trie"
+)
+
+type trieDiffStats struct {
+	start     time.Time
+	lastShown time.Time
+	onlyOn1   int
+	onlyOn2   int
+	inCommon  int
+}
+
+func trieDiff(ctx context.Context, kvs kvstore.KVStore) {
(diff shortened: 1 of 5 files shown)


tools/dbinspector/trie-diff.go
```diff
@@ -0,0 +1,134 @@
...
+	var current2 *nodeData
+	var ok2 bool
+	current1, ok1 = <-ch1
+	current2, ok2 = <-ch2
+	for ok1 && ok2 {
+		if current1.Commitment == current2.Commitment {
+			diff.inCommon++
+			current1, ok1 = <-ch1
+			current2, ok2 = <-ch2
+		} else if bytes.Compare(current1.key, current2.key) < 0 {
```
jorgemmsilva, line 74: this diff thing relies on the order of `.IterateNodes` right?

as an aleternative for GC/pruning, could we leverage the commitment tree itself? it should be fast to transverse the tries from the root and find where they diverge

jorgemmsilva (approved):

tools/dbinspector/trie-diff.go
dessaya: Yes, `IterateNodes` iterates in order, and this algorithm depends on that.

I didn't understand the question? What do you mean by the commitment tree?
It even sounds like the same thing but with different words haha. Here we pick two trie roots, and traverse both tries in order (and in parallel).

PR #2379, opened 2023-04-21
Refactor `peer_message.go`

muXxer:
Cluster tests passed for me.

### DIFF
--- a/packages/peering/lpp/lppNetImpl.go
+++ b/packages/peering/lpp/lppNetImpl.go
@@ -426,8 +426,6 @@ func (n *netImpl) NumUsers() int {
 }
 
 // Await implements peering.PeerSender for the Self() node.
-//
-//nolint:revive
 func (n *netImpl) Await(timeout time.Duration) error {
 	return nil // This node is alive immediately.
 }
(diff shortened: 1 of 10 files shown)


packages/peering/lpp/lppPeer.go
```diff
@@ -201,6 +201,8 @@ func (p *peer) IsAlive() bool {
 }
 
 // Await implements peering.PeerSender interface for the remote peers.
+//
+//nolint:revive
```
Juliusan: Github notifies:
```directive `//nolint:revive` is unused for linter "revive" (nolintlint)```
Maybe it should be removed.

File: packages/peering/lpp/lppNetImpl.go [hunk 2]
```diff
@@ -426,6 +426,8 @@ func (n *netImpl) NumUsers() int {
 }
 
 // Await implements peering.PeerSender for the Self() node.
+//
+//nolint:revive
```
Juliusan: Github notifies:
directive `//nolint:revive` is unused for linter "revive" (nolintlint)
Maybe it should be removed.

Juliusan (approved):

packages/peering/lpp/lppPeer.go
muXxer: That will get a problem with the next golangci version, but there are tons of other errors we need to address in another PR anyway. I will remove it for now :+1:

packages/peering/lpp/lppNetImpl.go
muXxer: That will get a problem with the next golangci version, but there are tons of other errors we need to address in another PR anyway. I will remove it for now :+1:

PR #2382, opened 2023-04-23
feat: Add health endpoint

### LINKED ISSUE #2274: Improve Health endpoint for API to actually look at the health instead of just returning 200 without looking at the functioning of the chain

howjmay:
close #2274


howjmay:
I haven't set the correct route

packages/metrics/chain_messages.go
```diff
@@ -58,6 +58,11 @@ type TxInclusionStateMsg struct {
 	State string
 }
 
+type ChainConfirmedState struct {
```
kape1395: I think this, and related functions should be not needed.

packages/metrics/chain.go
```diff
@@ -713,3 +716,7 @@ func (m *ChainMetricsProvider) OutPullTxInclusionState() IMessageMetric[iotago.T
 func (m *ChainMetricsProvider) OutPullOutputByID() IMessageMetric[iotago.OutputID] {
 	return m.outPullOutputByIDMetrics
 }
+
+func (m *ChainMetricsProvider) MaxChainConfirmedStateLag() uint32 { // TODO: Call it from the health endpoint.
```
jorgemmsilva:
```suggestion
func (m *ChainMetricsProvider) MaxChainConfirmedStateLag() uint32 {
```
TODO can be removed?

packages/webapi/api.go
```diff
@@ -33,8 +33,21 @@ import (
...
+func AddHealthEndpoint(server echoswagger.ApiRoot, chainService interfaces.ChainService, metricsService interfaces.MetricsService) {
+	server.GET("/health", func(e echo.Context) error {
+		chainIDs, err := chainService.GetAllChainIDs()
+		if err != nil {
+			return e.String(http.StatusInternalServerError, fmt.Sprintf("failed to get all chain IDs: %s", err))
+		}
+
+		for _, chainID := range chainIDs {
+			lag := metricsService.GetMaxChainConfirmedStateLag(chainID)
+			if lag > 2 {
```
jorgemmsilva: this `2` should probably be a configurable parameter

packages/webapi/api.go
```diff
@@ -33,8 +33,21 @@ import (
...
-func addHealthEndpoint(server echoswagger.ApiRoot) {
-	server.GET("/health", func(c echo.Context) error { return c.NoContent(http.StatusOK) }).
+func AddHealthEndpoint(server echoswagger.ApiRoot, chainService interfaces.ChainService, metricsService interfaces.MetricsService) {
+	server.GET("/health", func(e echo.Context) error {
+		chainIDs, err := chainService.GetAllChainIDs()
+		if err != nil {
+			return e.String(http.StatusInternalServerError, fmt.Sprintf("failed to get all chain IDs: %s", err))
+		}
+
+		for _, chainID := range chainIDs {
```
jorgemmsilva: could we use the `MaxLag` here?

packages/webapi/api.go
```diff
@@ -33,8 +33,21 @@ import (
 
 const APIVersion = 1
 
-func addHealthEndpoint(server echoswagger.ApiRoot) {
-	server.GET("/health", func(c echo.Context) error { return c.NoContent(http.StatusOK) }).
+func AddHealthEndpoint(server echoswagger.ApiRoot, chainService interfaces.ChainService, metricsService interfaces.MetricsService) {
+	server.GET("/health", func(e echo.Context) error {
+		chainIDs, err := chainService.GetAllChainIDs()
+		if err != nil {
+			return e.String(http.StatusInternalServerError, fmt.Sprintf("failed to get all chain IDs: %s", err))
```
jorgemmsilva: it's safe to panic in case of an internal error, it should be cached and logged by the middleware

jorgemmsilva (approved):

packages/webapi/api.go
howjmay: Um. Actually this part I have a question. We need to return an error if the node is unhealthy, but what is the error that we are going to return?

jorgemmsilva: http status 500 seems ok

jorgemmsilva (approved):

PR #2407, opened 2023-04-26
evm: ERC20BaseTokens add a special case for approve(MAX_UINT256)

### LINKED ISSUE #2390: Approve MAX_UINT256 for 0x1074010000000000000000000000000000000000 reverts
Approving MAX_UINT256 on 0x1074010000000000000000000000000000000000 reverts with 'amount is too large'

This is potentially similar to #2112

dessaya:
Also: fix the semantic meaning of `approve` where the amount specified replaces any previous approval, instead of adding.

closes: #2390


packages/vm/core/evm/evmimpl/iscmagic_state.go
```diff
@@ -43,11 +49,42 @@ func getAllowance(ctx isc.SandboxBase, from, to common.Address) *isc.Assets {
...
+		}
+		allowance.NativeTokens = lo.Values(ntSet)
+	})
+}
+
 func addToAllowance(ctx isc.Sandbox, from, to common.Address, add *isc.Assets) {
+	withAllowance(ctx, from, to, func(allowance *isc.Assets) {
+		allowance.Add(add)
+	})
+}
```
jorgemmsilva, line 81: does this method still make sense, since we always "set" the allowance?

packages/vm/core/evm/evmimpl/iscmagic_state.go
```diff
@@ -43,11 +49,42 @@ func getAllowance(ctx isc.SandboxBase, from, to common.Address) *isc.Assets {
...
+	withAllowance(ctx, from, to, func(allowance *isc.Assets) {
+		if !numTokens.IsUint64() {
+			// Calling `approve(MAX_UINT256)` is semantically equivalent to an "infinite" allowance
+			if numTokens.Cmp(gethmath.MaxBig256) == 0 {
+				numTokens = big.NewInt(0).SetUint64(math.MaxUint64)
+			} else {
+				panic("base tokens amount must be an uint64")
+			}
+		}
+		allowance.BaseTokens = numTokens.Uint64()
```
jorgemmsilva, line 62: we could potentially cap to the user current max balance, like suggested here: https://github.com/iotaledger/wasp/issues/2390#issuecomment-1523591419

jorgemmsilva (approved):

packages/vm/core/evm/evmimpl/iscmagic_state.go
dessaya: I think it's simpler and more flexible to allow approving more than the balance. E.g. it allows you to approve, then deposit if for some reason you wanted to do it in that order. This closer to the ERC20 behavior anyway.

packages/vm/core/evm/evmimpl/iscmagic_state.go
dessaya: It is left mainly for backward compatibility. Also the ISCSandbox::allow habdler still uses it (I didn't want to break its behavior).

packages/vm/core/evm/evmimpl/iscmagic_state.go
MoonLabsDev: I would recommend to do

if numTokens.Cmp(math.MaxUint64) >= 0
{
    numTokens = big.NewInt(0).SetUint64(math.MaxUint64)
}

This ensures that any value above Uint64 doesnt revert (which for many will be unexpected behavior), but instead it will approve MaxUint64

normally approval doesnt fail/revert at all and it would force users to maybe build special code if a token they interact with is inherited by your ERC20NativeTokens/BaseTokens Contract.

Besides that the "potentially cap to the users current max balance" for ERC20NativeTokens is for safety measures. To elaborate:

- On many chains, malicous actors do phishing attacks
   - Either they "send ETH" to themselfes and users dont realize it (as most arent experts and dont review their transactions or would even understand what they review
   - OR the malicious actor lets them "approve" a token, so that they can steal it
   
In case 1 they will get their ETH stolen once. In case 2 they will get their token stolen forever, until they revoke the approval, BUT only this 1 approved token can be stolen.

as "mSMR" as it was called by the commnuity is equal to SMR, a malicious actor can make a user approve mSMR (and thereby SMR) to himself. sure, as in case 2, only the "affected token" can be stolen, BUT as mSMR = SMR, it means i steal your gas token, so the user has no gas to revoke the approval. as he doesnt have any gas he also CANT access any other token in their wallet, which is equal to loosing the assets. Even IF he sends more SMR to have gas to undo the approval.... most malicious actors will have bots running that frequently check the RPCs to see if any hacked wallet has SMR again. they will steal it before the user has ANY chance to undo the approval. IF you thereby only let them allow approval up to their current balance, it would mitigate the issue as worst case "current balance can be stolen", but afterwards the user can safely send more SMR to either undo the remaining approval or to access his other tokens again. Please consider it.

PR #2434, opened 2023-05-05
fix: prevent overflow on minBurn

jorgemmsilva:


dessaya (approved):

packages/vm/gas/table.go
```diff
@@ -101,6 +101,10 @@ func linear(a uint64) BurnFunction {
 
 func minBurn(minGasBurn uint64) BurnFunction {
 	return func(currentBurnedGas uint64) uint64 {
-		return minGasBurn - currentBurnedGas
+		ret := minGasBurn - currentBurnedGas
+		if ret > minGasBurn {
+			return 0 // prevent overflow
+		}
+		return ret
```
muXxer:
```suggestion
		if minGasBurn < currentBurnedGas {
			// prevent overflow
			return 0
		}
		
		return minGasBurn - currentBurnedGas
```

muXxer: maybe a bit easier to read

PR #2440, opened 2023-05-08
Discard a pipe on consensus completion.

kape1395:
This could be used as a basis for closing pipe goroutines in all the components using them.
For now, only the consensus used this discard function.

### DIFF
--- a/packages/chain/cons/cons_gr/gr.go
+++ b/packages/chain/cons/cons_gr/gr.go
@@ -168,13 +168,12 @@ func New(
 	constInstRaw := cons.New(chainID, chainStore, me, myNodeIdentity.GetPrivateKey(), dkShare, procCache, netPeeringID[:], gpa.NodeIDFromPublicKey, log).AsGPA()
 	cgr.consInst = gpa.NewAckHandler(me, constInstRaw, redeliveryPeriod)
 
-	netRecvPipeInCh := cgr.netRecvPipe.In()
 	unhook := net.Attach(&netPeeringID, peering.PeerMessageReceiverChainCons, func(recv *peering.PeerMessageIn) {
 		if recv.MsgType != msgTypeCons {
 			cgr.log.Warnf("Unexpected message, type=%v", recv.MsgType)
 			return
 		}
-		netRecvPipeInCh <- recv
+		cgr.netRecvPipe.TryAdd(recv)
 	})
 	cgr.netDisconnect = unhook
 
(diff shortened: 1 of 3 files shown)


packages/chain/cons/cons_gr/gr.go
```diff
@@ -168,13 +168,12 @@ func New(
...
 	cgr.consInst = gpa.NewAckHandler(me, constInstRaw, redeliveryPeriod)
 
-	netRecvPipeInCh := cgr.netRecvPipe.In()
 	unhook := net.Attach(&netPeeringID, peering.PeerMessageReceiverChainCons, func(recv *peering.PeerMessageIn) {
 		if recv.MsgType != msgTypeCons {
 			cgr.log.Warnf("Unexpected message, type=%v", recv.MsgType)
 			return
 		}
-		netRecvPipeInCh <- recv
+		cgr.netRecvPipe.TryAdd(recv)
```
jorgemmsilva, line 176: I kinda liked the simple chan syntax from before.
It also avoided mutexes.
Shouldn't this component know when the chan is closed? Since it is the one who closes it?

jorgemmsilva (approved):

packages/chain/cons/cons_gr/gr.go
kape1395: The callback is executed in a different thread (net). That's all the problem, knowing when the channel should not be used for input anymore.

packages/util/pipe/pipe.go
```diff
@@ -69,7 +77,25 @@ func (ch *InfinitePipe[E]) Len() int {
 }
 
 func (ch *InfinitePipe[E]) Close() {
+	ch.closeLock.Lock()
```
muXxer, line 80: Can we use an atomic.Bool instead of the bool and the lock?

muXxer (approved):

packages/util/pipe/pipe.go
```diff
@@ -89,6 +115,11 @@ func (ch *InfinitePipe[E]) infiniteBuffer() {
 		case output <- next:
 			ch.buffer.Remove()
 		case ch.length <- ch.buffer.Length():
+		case <-ch.discardCh:
+			// Close the pipe without waiting for the values to be consumed.
+			close(ch.output)
+			close(ch.length)
+			return
```
Juliusan, line 122: `break` whould have sufficed, as the channels are closed in the end of the function. And shouldn't `discardCh` be also closed? As it is not closed after regular `Close`.

PR #2442, opened 2023-05-08
Feat: unprocessable

jorgemmsilva:
adds logic to handle "unprocessable requests", meaning: requests where the sender doesn't provide enough funds to split the request into the required accounting outputs.

adds an entrypoint to the `blocklog` contract to allow those requests to be retried

related issue: https://github.com/iotaledger/wasp/issues/2067


packages/isc/request.go
```diff
@@ -69,6 +69,10 @@ type OnLedgerRequest interface {
 	IsInternalUTXO(ChainID) bool
 	OutputID() iotago.OutputID
 	Features() Features
+	SetRetryOutputID(iotago.OutputID)
+	RetryOutputID() iotago.OutputID
+	IsRetry() bool
+	RetryOutput(iotago.AliasID) iotago.Output
```
kape1395: Why do we need such an invasive change here? Can't we have a function in a core contract to execute one of the saved requests?

jorgemmsilva: yup, we have that, there is the `retry` entrypoint on the `blocklog` contract.

This is to mark a request to be retried, basically saying:
"this is the original request, but it must be consumed with another output/outputID, because its assets were saved on a temporary UTXO"

not sure how to do this without adding some methods to the "onledger request type" itself

packages/vm/core/blocklog/unprocessable.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// save request reference / address of the sender
+func SaveUnprocessable(state kv.KVStore, req isc.OnLedgerRequest, blockIndex uint32, outputIndex uint16) {
+	rec := unprocessableRequestRecord{
+		blockIndex:  blockIndex,
+		outputIndex: outputIndex,
+		req:         req,
+	}
+	unprocessableMap(state).SetAt(req.ID().Bytes(), rec.bytes())
```
dessaya, line 63: Is this exploitable in any way?
e.g. Does this consume gas for storage?

packages/vm/runvm/runtask.go
```diff
@@ -51,10 +48,33 @@ func runTask(task *vm.VMTask) {
...
+func runTask(task *vm.VMTask) {
+	vmctx := vmcontext.CreateVMContext(task)
+
+	vmctx.OpenBlockContexts()
+
+	// run the batch of requests
+	results, numSuccess, numOffLedger := runRequests(vmctx, task.Requests, 0, task.Log)
+	{
+		// run any scheduled retry of "unprocessable" requests
+		results2, numSuccess2, numOffLedger2 := runRequests(vmctx, task.UnprocessableToRetry, uint16(len(results)), task.Log)
```
dessaya, line 67: So the unprocessable requests are always executed _after_ the regular requests. Is this exploitable?

packages/isc/request.go
dessaya: Maybe apply "interface segregation"? Not much of an improvement, just an idea...

```go
type OnLedgerRequest interface {
  ...
  SetRetry(iotago.OutputID) RetryOnLedgerRequest
  AsRetry() RetryOnLedgerRequest // returns nil if it is not a retry request
}

type RetryOnLedgerRequest {
  OnLedgerRequest

  RetryOutputID() iotago.OutputID
  RetryOutput(iotago.AliasID) iotago.Output
}

func (r *onLedgerRequestData) SetRetry(oid iotago.OutputID) RetryOnLedgerRequest {
	r.retryOutputID = oid
    return this
}

func (r *onLedgerRequestData) AsRetry(oid iotago.OutputID) RetryOnLedgerRequest {
	if r.retryOutputID == nil { return nil }
    return this
}
```

jorgemmsilva: Not sure if this would help, I want the VM to interpret this as an "onledger request", adding a new type would only add more edge cases :/

packages/vm/core/blocklog/unprocessable.go
jorgemmsilva: it does not, but users need to lock a substantial amount of funds just to store a few bytes, so I wouldn't say its a feasible attack

packages/vm/runvm/runtask.go
jorgemmsilva: good point. They are definitely "frontrunnable", not sure if there is any attack possible by running "in the end of the batch".
So I see that these "retries" could be exploited if they try to do something fancy in the same request as the assets are deposited, but nobody can do that at least for now that we only support EVM requests. Even if in the future with multiple VMs these requests could be exploited somehow, I don't think it's an issue since on-ledger requests should be mainly used as "a deposit mechanism".

packages/isc/request.go
dessaya: Ok, another idea:

```go
type retryOnLedgerRequest struct {
  OnLedgerRequest
  retryOutputID iotago.OutputID
}
```

This new type would implement `OnLedgerRequest`, so it could be interpreted as one.

I'm insisting because I agree with Karolis that this change seems intrusive here. I.e., from the point of view of, say, the mempool, the Request interface should stay the same, no? The concept of a "retry request" is something that makes sense only in the context of the VM.

jorgemmsilva: yup, this and a simple type assertion in the txbuilder is all that is needed. Thanks for the feedback!

dessaya (approved):

kape1395 (approved):

PR #2443, opened 2023-05-08
Fix reboot cluster tests

muXxer:


tools/cluster/templates/waspconfig.go
```diff
@@ -95,7 +95,7 @@ var WaspConfig = `
     "directory": "blocks"
   },
   "profiling": {
-    "enabled": false,
```
jorgemmsilva: keep this disabled pls, those tests take long enough lol

jorgemmsilva (approved):

tools/cluster/templates/waspconfig.go
muXxer: does this even change anything :D

PR #2448, opened 2023-05-09
allow calling ISC contracts from EVM contracts with proper caller

dessaya:
Includes foundry creation examples by Larry Ivange.


packages/vm/core/evm/evmimpl/iscmagic_handler.go
```diff
@@ -85,3 +86,15 @@ func reflectCall(handler any, method *abi.Method, args []any) []byte {
...
+func (h *magicContractHandler) call(target, ep isc.Hname, params dict.Dict, allowance *isc.Assets) dict.Dict {
+	return h.ctx.Privileged().CallOnBehalfOf(
+		isc.NewEthereumAddressAgentID(h.caller.Address()),
+		target, ep, params, allowance,
+	)
+}
+
+func (h *magicContractViewHandler) callView(target, ep isc.Hname, params dict.Dict) dict.Dict {
+	// TODO: currently there is no way to change the caller of a view in the VM
+	return h.ctx.CallView(target, ep, params)
```
jorgemmsilva: https://github.com/iotaledger/wasp/blob/develop/packages/vm/sandbox/sandboxbase.go#L142-L148

I think just calling ctx.Call here should be okay

jorgemmsilva (approved):

packages/vm/core/evm/evmimpl/iscmagic_handler.go
dessaya: Not really.
I implemented this in c3e3eaf but there is a problem here: https://github.com/iotaledger/wasp/pull/2448/files#diff-0a3f029823508bc069f9308ffce07683099970396982e2630e7242a4ee4150eeR204

The ISC `ViewContext` does not support the concept of a "caller" that is not an ISC contract.

For now this works, because I *think* that `ViewContext` is never used from the EVM jsonrpc. But I'm not 100% sure.

PR #2472, opened 2023-05-15
feat: Standardize the binary format of contracts events

### LINKED ISSUE #2086: Stardardize the binary format of all events triggered by core contracts

howjmay:
close #2086

### DIFF
--- a/contracts/native/inccounter/events.go
+++ b/contracts/native/inccounter/events.go
@@ -3,7 +3,6 @@ package inccounter
 import (
 	"bytes"
 	"fmt"
-	"time"
 
 	"github.com/iotaledger/wasp/packages/isc"
 	"github.com/iotaledger/wasp/packages/util"
@@ -26,7 +25,7 @@ func (e *InitializeEvent) Topic() []byte {
 
 func (e *InitializeEvent) Payload() []byte {
 	w := bytes.Buffer{}
-	if err := util.WriteUint64(&w, uint64(time.Now().Unix())); err != nil {
+	if err := util.WriteUint64(&w, e.Timestamp); err != nil {
 		panic(fmt.Errorf("failed to write event.Timestamp: %w", err))
 	}
 	if err := util.WriteInt64(&w, e.Counter); err != nil {
@@ -52,7 +51,7 @@ func (e *InitializeEvent) DecodePayload(payload []byte) {
 	}
 }
 
-var _ isc.Event = &InitializeEvent{}
+var _ isc.Event = &IncCounterEvent{}
 
 type IncCounterEvent struct {
 	Timestamp uint64
@@ -69,7 +68,7 @@ func (e *IncCounterEvent) Topic() []byte {
 
 func (e *IncCounterEvent) Payload() []byte {
 	w := bytes.Buffer{}
-	if err := util.WriteUint64(&w, uint64(time.Now().Unix())); err != nil {
+	if err := util.WriteUint64(&w, e.Timestamp); err != nil {
 		panic(fmt.Errorf("failed to write event.Timestamp: %w", err))
 	}
 	if err := util.WriteInt64(&w, e.Counter); err != nil {
@@ -110,7 +109,7 @@ func (e *IncCounterAndRepeatOnceEvent) Topic() []byte {
 
 func (e *IncCounterAndRepeatOnceEvent) Payload() []byte {
 	w := bytes.Buffer{}
-	if err := util.WriteUint64(&w, uint64(time.Now().Unix())); err != nil {
+	if err := util.WriteUint64(&w, e.Timestamp); err != nil {
 		panic(fmt.Errorf("failed to write event.Timestamp: %w", err))
 	}
 	if err := util.WriteInt64(&w, e.Counter); err != nil {
--- a/contracts/native/inccounter/events.go
+++ b/contracts/native/inccounter/events.go
@@ -30,8 +30,18 @@ func (e *InitializeEvent) Payload() []byte {
 	return w.Bytes()
 }
 
-func (e *InitializeEvent) Encode() []byte {
-	return append(e.Topic(), e.Payload()...)
+func (e *InitializeEvent) DecodePayload(payload []byte) {
+	r := bytes.NewReader(payload)
+	topic, err := util.ReadString16(r)
+	if err != nil {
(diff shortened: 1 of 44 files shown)


tools/schema/generator/gotemplates/events.go
```diff
@@ -23,9 +23,11 @@ $#each eventComment _eventComment
 func (e $TypeName) $EvtName($endFunc
 $#each event eventParam
 $#if event eventEndFunc2
-	evt := wasmlib.NewEventEncoder("$package.$evtName")
-$#each event eventEmit
-	evt.Emit()
+	enc := wasmtypes.NewWasmEncoder()
```
howjmay, line 42: I may need to implement the same `isc.Event` interface here.

So the following function will become encode the event object first, then call `Emit()` in which it will call `wasmlib.ScFuncContext{}.Event(buf)`

tools/schema/generator/gotemplates/events.go
```diff
@@ -23,9 +23,11 @@ $#each eventComment _eventComment
...
+	wasmtypes.StringEncode(enc, "$package.$evtName")
```
howjmay: I may need to change separate the `$package` part. Using Hname of the contract instead

BugFreeSoftware: We need to think this though first. After documentation I will implement inheritance of contract interfaces. A SC interface includes its events. So another SC that inherits an interface needs to emit events with the same name to make sure that you can plug in any SC with that interface and the client will be able to respond to the events.

packages/isc/sandbox_interface.go
```diff
@@ -127,6 +127,10 @@ type Sandbox interface {
 	Privileged() Privileged
 }
 
+type Event interface {
+	Encode() []byte
+}
```
dessaya: I suggest:

```suggestion
type Event interface {
    Topic() Hname
	Payload() []byte
}

func EncodeEvent(Event e) []byte {
    return append(e.Topic().Bytes(), e.Topic.Payload()...)
}
```

This way me make it explicit that all events have a "topic" (that could be anything, we just have to make sure not to use the same topic for 2 different events), and that the first 4 bytes are the topic.

So for example:

```go
type FoundryCreateNewEvent struct {
	SerialNumber uint32
}

func (e *FoundryCreateNewEvent) Topic() isc.Hname {
	return isc.Hn("FoundryCreateNewEvent")
}

func (e *FoundryCreateNewEvent) Payload() []byte {
	return codec.MustEncodeUint32(e.SerialNumber)
}
```

howjmay: I make the interface as a single one

```
type Event interface {
	Topic() []byte
	Payload() []byte
	Encode() []byte
}
```

dessaya: I don't think the `Encode()` function should be part of the interface.
The idea is to standardize the format, and for that we want all events to share a common format. If we allow each event to `Encode()` itself however it wants, we defeat the purpose.

My proposal has only `Topic()` and `Payload()`: the topic is the common part of all events that enforces a 4-byte topic, and the payload is the custom part for each event. So we standardize that the encoding is `<topic><payload>` and we centralize that in a single `EncodeEvent(Event)` function.

howjmay: I will separate the Encode() from the interface, but the length of Topic() I am still discussing with @BugFreeSoftware for a comprehensive solution

howjmay: But for the decoding part. If we want a single `EncodeEvent()` to decode all the events, it seems to be hard. It may be can just served as a wrapper for the `Decode(e DecodeEventObj)` method for each Event Object. But in this way, I need an interface to define the type to be used as input argument `DecodeEventObj`.

Do you think this is ok? If it is ok, then I think I may need a proper naming

dessaya: Can you clarify?

I didn't think too much about the decoding because the events are something that we "fire and forget". If a client wants to decode it, it would first inspect the topic, and then based on that decode the payload using the proper Event subtype.

So maybe the Event interface should have a `DecodePayload(payload []byte)`.

howjmay: Sorry I am still confused about why the following interface defeat the purpose of standardize the format.

```
type Event interface {
	Topic() []byte
	Payload() []byte
	Encode() []byte
}
```


The encoded byte slice will still follow the similar format. 

Putting them all together maybe easy to unify them? So have a `DecodePayload(payload []byte)` in the Event interface is good for me

dessaya: It's about not repeating the same logic N times if you have N different events. If `Encode() []byte` is part of the interface then you have to implement tha same thing N times.

howjmay: Oh true. Them I think I want to make the interface as following

```
type Event interface {
	Topic() []byte
	Payload() []byte
        DecodePayload(payload []byte)
}
```

And have two public function 
```
func Encode(e Event) []byte
func Decode(e Event, payload []byte) Event
```

howjmay: BTW how can we get iotago.Address from string?

dessaya: `iotago.ParseBech32()`

howjmay:
I need to inspect the performance of the event encode/decode. 
Now `TestChained` with `--short` can passed on my machine. It took `2.545s` on my machine, but it is way slower on Github Action.

contracts/native/inccounter/events.go
```diff
@@ -0,0 +1,137 @@
...
+	w := bytes.Buffer{}
+	if err := util.WriteBytes8(&w, []byte("InitializeEvent")); err != nil {
+		panic(fmt.Errorf("failed to write InitializeEvent: %w", err))
+	}
+	return w.Bytes()
+}
+
+func (e *InitializeEvent) Payload() []byte {
+	w := bytes.Buffer{}
+	if err := util.WriteUint64(&w, uint64(time.Now().Unix())); err != nil {
```
kape1395: Why do you take the current time?

howjmay: So if I want to take a timestamp around the current time, how can I get it?

howjmay: Should I call blocklog contract?

howjmay: Got it!

PR #2482, opened 2023-05-17
Unregister pipe metrics properly.

kape1395:


packages/metrics/chain.go
```diff
@@ -667,21 +681,24 @@ func (m *ChainMetricsProvider) PrometheusCollectorsWebAPI() []prometheus.Collect
 }
 
 func (m *ChainMetricsProvider) RegisterChain(chainID isc.ChainID) {
-	m.chainsRegistered = append(m.chainsRegistered, chainID)
+	m.GetChainMetrics(chainID)
```
jorgemmsilva, line 684: Kinda odd that "register" function calls "getSomething" to create an object

packages/metrics/chain_pipe.go
```diff
@@ -50,6 +73,9 @@ func (m *chainPipeMetrics) makeLabels(pipeName string) prometheus.Labels {
 }
 
 func (m *chainPipeMetrics) TrackPipeLen(name string, lenFunc func() int) {
+	m.regLock.Lock()
```
jorgemmsilva, line 76: Could be Rlock here?

jorgemmsilva (approved):

packages/metrics/chain.go
kape1395: That's because the API creates these objects as well. And it is possible that the chain metrics object will be present already on this call. Otherwise, we should reject API calls while the chain is not registered yet.

packages/metrics/chain_pipe.go
kape1395: No, because we modify the structures here:
```
m.lenMetrics[name] = collector
```

PR #2487, opened 2023-05-17
feat(trie): implement pruning with reference counting

dessaya:


packages/trie/prune.go
```diff
@@ -0,0 +1,47 @@
...
+		deleteNode, deleteValue := refcounts.Dec(n)
+		if deleteNode {
+			deletedNodes = append(deletedNodes, n.Commitment)
+		}
+		if deleteValue {
+			deletedValues = append(deletedValues, n.Terminal.Bytes())
+		}
+		if deleteNode {
+			return IterateContinue
+		}
```
jorgemmsilva: could be simplified like this?
```suggestion
		if deleteValue {
			deletedValues = append(deletedValues, n.Terminal.Bytes())
		}
		if deleteNode {
			deletedNodes = append(deletedNodes, n.Commitment)
			return IterateContinue
		}
```

jorgemmsilva (approved):
looks great :sunglasses: , just a minor suggestion

packages/trie/refcounts.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (d *Refcounts) incNode(commitment Hash) uint32 {
+	return incRefcount(d.nodes, commitment[:])
+}
+
+func (d *Refcounts) Dec(node *NodeData) (deleteNode, deleteValue bool) {
+	n := decRefcount(d.nodes, node.Commitment[:])
+	deleteNode = n == 0
+	if n == 0 && node.Terminal != nil && !node.Terminal.IsValue {
```
jorgemmsilva, line 45: question:
should this check be removed here and on the `Inc` function?
```
!node.Terminal.IsValue
```
don't we always want to change the refcount, regardless of whether the commitment is an hash or the value itself?

dessaya: No, the refcount that is being changed here is the refcount of the values that are stored outside the trie. (`!node.Terminal.IsValue` means "the value is stored outside the trie".)

The value is stored as `hash(value) => value` in the value store. Multiple nodes may reference the same value, so we need to keep reference counts for this independently from the reference counts for each node.

Issue #2385, opened 2023-04-24
Create a public info endpoint which can always be queried (no authentication), we can make this endpoint part of the Alias ID metadata to request more information from the chain.


fijter:
Right now, to interact with a chain, you need the alias ID for on-ledger requests, and the json/rpc endpoint or API for off-ledger requests. It would be good to have all that data available with just one variable: The Alias ID. Given the Alias ID you can see the Alias metadata, this could contain a publicly exposed URL containing settable metadata. 

This issue has a couple of deliverables:

 - [ ] A publicly accessible endpoint (regardless of authentication settings) serving a metadata json
 - [ ] A way to set metadata values that are exposed in the new API endpoint - this could be on a per chain or per node way. Ideally both but can get away with just one of them (should be prefixed with `chain_` and `node_` to make clear distinction. Implementation of this can be interpreted in multiple ways, where a simple config file is the MVP but a wasp-cli command and dashboard interface preferred/nice to have.
 - [ ] A set of standard names which are well documented as a standard that should always be available and exposed in the json (can even have a null value if not provided). Should be things like `name`, `evm_rpc_url`, `evm_rpc_ws`, etc.)
 - [ ] Some basic configuration values like the gas parameters could also be exposed for additional info

WIth this you can get all the info you'd ever need from a chain with just the Alias ID, and even more if you have/provide full API access on that node.

PR #2498, opened 2023-05-22
Add metadata feature

lmoe:
# Description of change

Adds a metadata feature for Firefly.

* makes `chains/:chainID` public
* stores the publicly reachable URL on the Tangle 
* adds a new command to the wasp-cli to configure the metadata urls

fixes issue #2385

## Type of change


packages/vm/core/governance/governanceimpl/metadata.go
```diff
@@ -0,0 +1,52 @@
+package governanceimpl
+
+import (
+	"github.com/iotaledger/wasp/packages/isc"
+	"github.com/iotaledger/wasp/packages/kv/codec"
+	"github.com/iotaledger/wasp/packages/kv/dict"
+	"github.com/iotaledger/wasp/packages/vm/core/governance"
+)
+
+const MaxURLLength = 1000
```
jorgemmsilva: while this can work, I would rather use a real value like we had before:
```
const MaxCustomMetadataLength = iotago.MaxMetadataLength - serializer.OneByte - serializer.UInt32ByteSize - state.L1CommitmentSize - gas.GasPolicyByteSize - serializer.UInt16ByteSize
```

at least for the bytes that will be added to the L1 output [here](https://github.com/iotaledger/wasp/blob/develop/packages/vm/vmcontext/txbuilder.go#L29)

also keep in mind that the length of a string in bytes is different of its length as "string"

jorgemmsilva (approved):

packages/vm/core/governance/governanceimpl/metadata.go
lmoe: During development MaxCustomMetadataLength was around 8k bytes. This means, a very long URL. 

I think this should rather be applied to the whole StateMetadata validation before sending it away, instead of just checking the Metadata URL with it.

As long as the chars stay between a-Z + :/. the byte size should be 1 byte per char afaik.

jorgemmsilva: Sure, but if the other parts of the metadata change for some reason, it's easier to know the real hard limit, even if we enforce a lower limit for now

PR #2509, opened 2023-05-23
feat: extend chain metadata

lmoe:
# Description of change

Adds the option to give each chain:

* a name
* a description
* ~a contact email~
* a project website 

Furthermore, all metadata related fields are stored in a separate structure to keep it clean.
API was regenerated and `wasp-cli` was extended to set the metadata values. 

Tests were written and each field gets tested.

`--skip-version-check` for the `wasp-cli` was moved to the rootCmd, to also include `login` and other commands. 

## Type of change


packages/isc/chaininfo.go
```diff
@@ -7,14 +7,23 @@ import (
...
 )
 
+type ChainMetadata struct {
+	EVMJsonRPCURL   string
+	EVMWebSocketURL string
+
+	ChainName        string
+	ChainDescription string
+	ChainOwnerEmail  string
+	ChainWebsite     string
```
dessaya: why not

```suggestion
	Name        string
	Description string
	OwnerEmail  string
	Website     string
```

packages/vm/core/governance/governanceimpl/chaininfo.go
```diff
@@ -19,16 +19,32 @@ func getChainInfo(ctx isc.SandboxView) dict.Dict {
...
+		ret.Set(governance.VarMetadataChainDescription, codec.EncodeString(info.Metadata.ChainDescription))
+	}
+
+	if len(info.Metadata.ChainOwnerEmail) > 0 {
+		ret.Set(governance.VarMetadataChainOwnerEmail, codec.EncodeString(info.Metadata.ChainOwnerEmail))
+	}
+
+	if len(info.Metadata.ChainWebsite) > 0 {
+		ret.Set(governance.VarMetadataChainWebsite, codec.EncodeString(info.Metadata.ChainWebsite))
 	}
```
dessaya: Why not have a single `VarMetadata` in the state and just

```
ret.Set(governance.VarMetadata, info.Metadata.Bytes())
```

packages/vm/core/governance/governanceimpl/metadata.go
```diff
@@ -12,31 +12,71 @@ import (
...
+	MaxURLLength              = 1000
+	MaxChainNameLength        = 255
+	MaxChainDescriptionLength = 2048
+	MaxChainOwnerEmailLength  = 255
 )
 
 func setMetadata(ctx isc.Sandbox) dict.Dict {
 	ctx.RequireCallerIsChainOwner()
 
 	publicURLBytes := ctx.Params().Get(governance.ParamPublicURL)
```
dessaya: Same here, why not have a single ParamMetadata:

```
meta := MetadataFromBytes(ctx.Params().Get(governance.ParamMetadata))
err := meta.Validate()
ctx.RequireNoError(err)
```

I'll stop here, the suggestion applies to the rest of the PR: centralize all metadata-related logic in a single object.

packages/isc/chaininfo.go
```diff
@@ -7,14 +7,23 @@ import (
 	"github.com/iotaledger/wasp/packages/vm/gas"
 )
 
+type ChainMetadata struct {
+	EVMJsonRPCURL   string
+	EVMWebSocketURL string
+
+	ChainName        string
+	ChainDescription string
+	ChainOwnerEmail  string
```
jorgemmsilva: do we really want this? seems like "hey send phishing emails here" kinda thing :joy: 

reminds of the old days of "webmaster@website.com", now "chainmaster@website.com" :trollface:

packages/vm/core/governance/governanceimpl/chaininfo.go
jorgemmsilva: I agree this is cleaner. 

However it means implementing custom de/serialization for yet another object :smile: 

true hand-made software  :trollface:

lmoe: Meh, I'm changing it to one parameter. 

Thought it would be easier to use and access as you can simply get each value by key instead of requiring de/encode of the whole metadata struct. Maybe less clean on the backend side, but easier to query on the frontend side.

packages/isc/chaininfo.go
lmoe: Good point :D 
I have it removed

packages/isc/chain_metadata.go
```diff
@@ -0,0 +1,76 @@
+package isc
+
+import (
+	"github.com/iotaledger/hive.go/serializer/v2/marshalutil"
+)
+
+type ChainMetadata struct {
```
jorgemmsilva: could we rename this to "publicMetadata" or "integrationMetadata", something like that?

I'm afraid "chainMetadata" will be easily confused with "stateMetadata" :sweat_smile:

packages/isc/chain_metadata.go
```diff
@@ -0,0 +1,76 @@
...
+	return string(ret), nil
+}
+
+func writeMetadataString(mu *marshalutil.MarshalUtil, str string) {
+	mu.
+		WriteUint16(uint16(len(str))).
+		WriteBytes([]byte(str))
+}
+
+func MetadataFromMarshalUtil(mu *marshalutil.MarshalUtil) (*ChainMetadata, error) {
```
jorgemmsilva: please write a unit test for serialization/deserialization

packages/vm/core/governance/governanceimpl/metadata.go
```diff
@@ -13,47 +13,40 @@ import (
...
-	ctx.Requiref(len(publicURL) <= MaxCustomMetadataLength, "public url size too big (%d>%d)", len(publicURL), MaxCustomMetadataLength)
+	var publicURLBytes []byte
+	var metadataBytes []byte
 
-	evmJSONRPCUrlBytes := ctx.Params().Get(governance.ParamMetadataEVMJsonRPCURL)
-	evmJSONRPCUrl, err := codec.DecodeString(evmJSONRPCUrlBytes, "")
-	ctx.RequireNoError(err)
-	ctx.Requiref(len(evmJSONRPCUrl) <= MaxURLLength, "evm json rpc url size too big (%d>%d)", len(evmJSONRPCUrl), MaxURLLength)
+	publicURLBytes = ctx.Params().Get(governance.ParamPublicURL)
+	metadataBytes = ctx.Params().Get(governance.ParamMetadata)
```
jorgemmsilva, line 25: why isn't the publicURL part of the metadata itself? :thinking:

packages/vm/core/governance/interface.go
```diff
@@ -47,7 +47,7 @@ var (
 	FuncStopMaintenance      = coreutil.Func("stopMaintenance")
 	ViewGetMaintenanceStatus = coreutil.ViewFunc("getMaintenanceStatus")
 
-	// L1 metadata
+	// L2 metadata
```
jorgemmsilva:
```suggestion
	// public metadata
```
or something like this

packages/vm/core/testcore/sbtests/2chains_test.go
```diff
@@ -26,6 +26,9 @@ import (
 func Test2Chains(t *testing.T) { run2(t, test2Chains) }
 
 func test2Chains(t *testing.T, w bool) {
+	// if !w {
+	// 	t.Skip("TODO remove")
+	// }
```
jorgemmsilva:
```suggestion
```

packages/wasmvm/wasmlib/as/wasmlib/coregovernance/consts.ts
```diff
@@ -9,32 +9,29 @@ export const ScName        = 'governance';
...
+export const ParamActions     = 'n';
+export const ParamAddress     = 'S';
+export const ParamCertificate = 'ic';
+export const ParamChainOwner  = 'o';
+export const ParamFeePolicy   = 'g';
+export const ParamGasLimits   = 'l';
+export const ParamGasRatio    = 'e';
+export const ParamMetadata    = 'md';
+export const ParamPubKey      = 'ip';
+export const ParamPublicURL   = 'x';
```
jorgemmsilva, line 23: gee what is this? do we really need all these updates to wasm stuff? :/

jorgemmsilva (approved):

packages/vm/core/governance/governanceimpl/metadata.go
lmoe: The publicURL is stored in L1 and the metadata is exposed on the webapi. So publicUrl leads to the metadata and I thought it makes sense to separate those two, as the publicUrl does not belong to the metadata in my opinion.

packages/isc/chain_metadata.go
lmoe: How about `PublicChainMetadata`? Just PublicMetadata or IntegrationMetadata does not really explain what it does.

packages/vm/core/governance/governanceimpl/metadata.go
jorgemmsilva: makes sense :+1:

PR #2510, opened 2023-05-23
feat: blocklog pruning

dessaya:


packages/evm/jsonrpc/evmchain.go
```diff
@@ -573,13 +576,15 @@ func blockchainDB(chainState state.State) *emulator.BlockchainDB {
...
 	gasFeePolicy := governance.MustGetGasFeePolicy(govPartition)
+	blockKeepAmount := governance.GetBlockKeepAmount(govPartition)
 	bdbPartition := subrealm.NewReadOnly(
 		chainState,
 		kv.Key(evm.Contract.Hname().Bytes())+evm.KeyEVMState+emulator.KeyBlockchainDB,
 	)
 	return emulator.NewBlockchainDB(
 		buffered.NewBufferedKVStore(bdbPartition),
 		gas.EVMBlockGasLimit(gasLimits, &gasFeePolicy.EVMGasRatio),
+		int(blockKeepAmount),
```
jorgemmsilva: why the mismatch types for `blockKeepAmount`? `int32` in gov contract and `int` in blockchaindb?

documentation/docs/guide/evm/using.md
```diff
@@ -22,8 +22,7 @@ keywords:
 ## 1. Deploy an IOTA Smart Contracts Chain
 
 When [deploying an IOTA Smart Contracts chain](../chains_and_nodes/setting-up-a-chain.md), EVM support is automatically
-added with the default configuration. The `wasp-cli chain deploy` command accepts some EVM-specific options listed
-below:
+added with the default configuration. The `wasp-cli chain deploy` command accepts the following EVM-specific option:
 
 * `--evm-chainid <n>`: EVM chain ID (default: 1074).
```
jorgemmsilva, line 27: missing the new `block-keep-amount`

packages/vm/core/blocklog/external.go
```diff
@@ -115,10 +112,32 @@ func GetEventsByBlockIndex(partition kv.KVStoreReader, blockIndex uint32, totalR
 	return ret
 }
 
-func GetBlockInfo(partition kv.KVStoreReader, blockIndex uint32) (*BlockInfo, error) {
-	ret, err := BlockInfoFromBytes(getBlockInfoBytes(partition, blockIndex))
+func PruneEventsByBlockIndex(partition kv.KVStore, blockIndex uint32, totalRequests uint16) []string {
```
jorgemmsilva: should this be an "external" func? it's not used outside `blocklog`, right? probably could  be "not exported" and moved to `internal.go`

packages/vm/core/blocklog/internal.go
```diff
@@ -20,6 +20,31 @@ func SaveNextBlockInfo(partition kv.KVStore, blockInfo *BlockInfo) {
 	registry.Push(blockInfo.Bytes())
 }
 
+func Prune(partition kv.KVStore, latestBlockIndex uint32, blockKeepAmount int32) {
```
jorgemmsilva: this one should probably be in `external.go`

packages/vm/core/blocklog/internal.go
```diff
@@ -20,6 +20,31 @@ func SaveNextBlockInfo(partition kv.KVStore, blockInfo *BlockInfo) {
...
+	if blockKeepAmount <= 0 {
+		// keep all blocks
+		return
+	}
+	if latestBlockIndex < uint32(blockKeepAmount) {
+		return
+	}
+	toDelete := latestBlockIndex - uint32(blockKeepAmount)
+	// assume that all blocks prior to `toDelete` have been already deleted, so
+	// we only need to delete this one.
```
jorgemmsilva: we should maybe enfoce (prune those blocks) this when the "block to keep" setting is changed, so to not leave "zombie blocks"

jorgemmsilva (approved):

packages/vm/core/blocklog/internal.go
dessaya: But how would that work? What would happen if blockKeepAmount was 10000 and it's changed to 10? Deleting 9990 blocks in a loop is a no-go IMO.

jorgemmsilva: >Deleting 9990 blocks in a loop is a no-go IMO.

why? :sweat_smile: too slow?

dessaya: Because of the size of the block. Isn't there a limit for the list of mutations in the block?

jorgemmsilva: there isn't afaik, but you're right. That would probably result in gigantic blocks, not a good thing.

PR #2519, opened 2023-05-25
serialize VMError params with binary format

dessaya:


packages/isc/vmerror.go
```diff
@@ -338,3 +281,134 @@ func VMErrorMustBe(err error, expected VMErrorBase) {
...
+			}
+			ret[i] = n
+		case reflect.Int64:
+			n, err := mu.ReadInt64()
+			if err != nil {
+				return nil, err
+			}
+			ret[i] = n
+		default:
+			panic(fmt.Sprintf("no handler for param of kind %s", reflect.Kind(kind)))
```
jorgemmsilva, line 433: won't this be easy to crash? just pass an Array /Bool/Map/struct/whatever else

jorgemmsilva (approved):

packages/isc/vmerror.go
dessaya: Yes... crashing is OK IMO because these errors are created under the context of the VM.  But maybe it should crash when the error is created instead. I'll do that.

PR #2524, opened 2023-05-27
Refactor events

BugFreeSoftware:


packages/vm/vmcontext/internal.go
```diff
@@ -199,14 +201,24 @@ func (vmctx *VMContext) storeUnprocessable(lastInternalAssetUTXOIndex uint16) {
...
+	timestamp := uint64(vmctx.Timestamp().UnixNano())
+
+	// event data is hContract / topic / timestamp / payload
+	eventData := make([]byte, 0, 4+2+len(topic)+8+len(payload))
+	eventData = append(eventData, util.Uint32To4Bytes(uint32(hContract))...)
+	eventData = append(eventData, util.Uint16To2Bytes(uint16(len(topic)))...)
+	eventData = append(eventData, []byte(topic)...)
+	eventData = append(eventData, util.Uint64To8Bytes(timestamp)...)
+	eventData = append(eventData, payload...)
+	eventKey := vmctx.eventLookupKey().Bytes()
```
jorgemmsilva: this should be a struct with proper `marshal/unmarshal` methods, so its easier to use. That function can also be used in the publisher/webapi to publish a humanly-readable JSON.

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
```diff
@@ -22,7 +22,8 @@ func (h *magicContractHandler) GetEntropy() hashing.HashValue {
 
 // handler for ISCSandbox::triggerEvent
 func (h *magicContractHandler) TriggerEvent(s string) {
-	h.ctx.Event(s)
+	// TODO adjust triggerevent and all .sol code
+	h.ctx.Event("evm.event", []byte(s))
 }
```
jorgemmsilva, line 27: IMO the signature should be changed to `TriggerEvent(topic string, s string)`
but maybe it doesn't even make sense that EVM can issue ISC events and we could deprecate this method altogether. Up for discussion

packages/publisher/events.go
```diff
@@ -119,17 +120,30 @@ func PublishBlockEvents(blockApplied *blockApplied, events *Events, log *logger.
...
 		// must be refactored so its possible to filter by "events from a contract"
+		// EDIT (Eric): But a block can contain multiple requests to different
+		// contracts, so this is not where this info is encoded. Instead it's
+		// in the event data for each event.
 		Issuer: &isc.NilAgentID{},
 		// TODO should be possible to filter by request ID (not possible with current events impl)
 		// RequestID: event.RequestID,
-		Payload: blockEvents,
+		Payload: payload,
 		ChainID: chainID,
```
jorgemmsilva: there should be a "core event" type that is used in `MustSaveEvent` , that type can be deserialized here and used to provide a nice struct that can be here to be filtered

packages/webapi/models/core_blocklog.go
```diff
@@ -54,5 +54,5 @@ type RequestProcessedResponse struct {
 }
 
 type EventsResponse struct {
-	Events []string `json:"events" swagger:"required"`
+	Events [][]byte `json:"events" swagger:"required"`
```
jorgemmsilva: should return a proper structure with "issuer,timestamp,etc"

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
BugFreeSoftware: Yeah already talked to @dessaya about this and he has offered to take this up as it involves changing a lot of .sol files as well.

packages/isc/event.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type Event struct {
+	ContractID Hname  `json:"contractID"`
+	Payload    []byte `json:"payload"`
+	Topic      string `json:"topic"`
+	Timestamp  uint64 `json:"timestamp"`
+}
+
+func NewEvent(event []byte) (*Event, error) {
```
jorgemmsilva, line 17: could use marshalutil like most of the code, for consistency

packages/vm/core/blocklog/internal.go
```diff
@@ -148,62 +151,67 @@ func getRequestEventsInternal(partition kv.KVStoreReader, reqID isc.RequestID) (
...
-func pruneEventsByBlockIndex(partition kv.KVStore, blockIndex uint32, totalRequests uint16) []string {
-	ret := make([]string, 0)
+func pruneEventsByBlockIndex(partition kv.KVStore, blockIndex uint32, totalRequests uint16) {
+	// TODO what about these contract LUTs?
+	// scLut := collections.NewMap(partition, prefixSmartContractEventsLookup)
+	// we need to walk over all possible hContract values
+	// for each do a get of the lut
+	// for each lut scan and skip each block with a value < latestBlockIndex
+	// save the remaining part slice of the lut
+
```
jorgemmsilva, line 201: https://github.com/iotaledger/wasp/issues/2529
created an issue so this is not forgotten

jorgemmsilva (approved):

PR #2525, opened 2023-05-29
fix: write failed evm receipts after execution, rather than at the end of block

jorgemmsilva:


packages/vm/core/evm/emulator/emulator.go
```diff
@@ -20,9 +20,11 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 
 	"github.com/iotaledger/wasp/packages/evm/evmutil"
+	"github.com/iotaledger/wasp/packages/isc"
```
dessaya: Can this dependency be avoided? The EVMEmulator is designed to be agnostic from ISC.

packages/vm/core/evm/emulator/emulator.go
```diff
@@ -279,6 +285,26 @@ func (e *EVMEmulator) SendTransaction(
 	gasUsed := uint64(0)
 	if result != nil {
 		gasUsed = result.UsedGas
+		if ctx != nil {
+			// convert burnt EVM gas to ISC gas
```
dessaya: Same thing: all this logic is ISC-related and should not be handled by EVMEmulator.

It's just to keep a well-defined boundary between layers of abstraction.

jorgemmsilva:
@dessaya pushed an update where the ISC dependency is removed from `evm/emulator`.

While I understand the wish to keep both layers separated, they are already connected in the sense that we pass functions like `gasBurnEnable` which is  ISC-only functionality. I guess not importing the package directly does reduce the coupling in a way, but the logic is intertwined, extracting into generic fuctions just adds a bit more navigational effort IMO.

dessaya:
Yes, I agree. But still, I think it's beneficial to keep this separation, as it makes the emulator easier to test in isolation.

packages/vm/core/evm/emulator/emulator.go
```diff
@@ -287,7 +288,12 @@ func (e *EVMEmulator) SendTransaction(
 
 	gasUsed := uint64(0)
 	if result != nil {
-		gasUsed = result.UsedGas
+		// chargeISCGas will keep gasBurnEnabled = false and mutate `result` when charging ISC gas fails
+		if chargeISCGas != nil {
+			gasUsed, iscGasErr = chargeISCGas(result)
+		} else {
+			gasUsed = result.UsedGas
+		}
```
dessaya, line 296: https://github.com/iotaledger/wasp/blob/a302676ddcfa1b02a9b52a0506f88f7a7dce9dfa/packages/vm/core/evm/evmimpl/iscmagic.go#L110

Maybe it's possible to obtain the same result by modifying the logic there so that it deducts the gas for each ISC call automatically.

But we can do that in a separate PR.

dessaya (approved):

PR #2528, opened 2023-05-30
Request ordering should respect the nonces.

kape1395:


packages/chain/cons/bp/aggregated_batch_proposals.go
```diff
@@ -132,6 +133,28 @@ func (abp *AggregatedBatchProposals) OrderedRequests(requests []isc.Request, ran
...
+			oj, ok := sortBuf[j].req.(isc.OffLedgerRequest)
+			if !ok {
+				continue
+			}
+			if oi.SenderAccount().Equals(oj.SenderAccount()) && oi.Nonce() > oj.Nonce() {
+				// Swap entries, if they are from the same account and have out-of-order nonces.
+				sortBuf[i], sortBuf[j] = sortBuf[j], sortBuf[i]
+			}
+		}
+	}
```
jorgemmsilva, line 156: if possible we should extract to a function and re-use in solo

jorgemmsilva (approved):

packages/chain/cons/bp/aggregated_batch_proposals.go
jorgemmsilva: and add unit tests :stuck_out_tongue:

kape1395: To have something to skip later? :)

kape1395: @jorgemmsilva, the best way to reuse this function is probably to use this object directly. I.e prepare a batch proposal and aggregate it with N=1.

dessaya: Sorry for coming late to the party...

Are we sure the sorting algorithm is correct? It seems to me that it swaps everything twice so that in the end it ends up in the same place.

dessaya: nvm, I see it was fixed afterwards :)

PR #2545, opened 2023-05-31
Clean up and sync Array16 and Array32

BugFreeSoftware:


contracts/native/micropay/impl.go
```diff
@@ -255,13 +255,13 @@ package micropay
 // }
 
 // func decodePayments(state kv.KVStoreReader, a assert.Assert) []*Payment {
```
jorgemmsilva: should we remove this "micropay" contract? this thing seems to be commented out for 1year+

packages/util/rwutil.go
```diff
@@ -398,6 +398,113 @@ func WriteTime(w io.Writer, ts time.Time) error {
 	return WriteUint64(w, uint64(ts.UnixNano()))
 }
 
+//////////////////// Size16, Size32 \\\\\\\\\\\\\\\\\\\\
+
+// use ULEB16 decoding so that WasmLib can use it as well
+func BytesToSize16(buf []byte) uint16 {
```
jorgemmsilva, line 404: to be honest I have no clue about the cleaver 0x stuff going on on this file, I'll just trust it works :shrug:

jorgemmsilva (approved):

contracts/native/micropay/impl.go
BugFreeSoftware: Yeah, will remove it

PR #2547, opened 2023-05-31
A benchmark to investigate the rocksdb write aplification.

kape1395:


jorgemmsilva (approved):

jorgemmsilva (commented):
wonder if we should move this into some other directory, doesn't really make sense to run this with the regular unit tests, right?

kape1395:
> wonder if we should move this into some other directory, doesn't really make sense to run this with the regular unit tests, right?

Yes, indeed. That's why I have added `t.Skip()` at the beginning. Maybe you know a better place/approach to keep such a test / benchmark?

jorgemmsilva:
we could create some `benchmark` dir, we have some other tests like this. but I guess its okay to just skip

PR #2552, opened 2023-05-31
impr: (vm) allow only strictly squential nonces, (mempool) don't propose nonce gaps

jorgemmsilva:


packages/chain/cons/cons_test.go
```diff
@@ -338,7 +333,7 @@ func testChained(t *testing.T, n, f, b int) {
 		allRequests[0] = append(tcl.MakeTxAccountsDeposit(scClient), tcl.MakeTxDeployIncCounterContract()...)
 	}
 	incTotal := 0
-	for i := 1; i < b; i++ {
+	for i := 0; i < b-1; i++ {
```
kape1395, line 336: Why have you moved it to 0 based?
The block number 0 is already produced 3 lines above. Hence I start here from 1.

packages/chain/mempool/mempool.go
```diff
@@ -542,10 +548,56 @@ func (mpi *mempoolImpl) handleConsensusProposalForChainHead(recv *reqConsensusPr
 			return true // Keep them for now
 		})
 	}
+
+	expectedAccountNonces := map[string]uint64{} // string is isc.AgentID.String()
+	requestsNonces := map[string][]reqRefNonce{} // string is isc.AgentID.String()
+	accountsContractStatePartition := subrealm.NewReadOnly(mpi.chainHeadState, kv.Key(accounts.Contract.Hname().Bytes()))
```
kape1395: I prefer to wrap such logic into something like `accounts.StateAccess` by analogy with `governance.StateAccess`.

kape1395 (approved):

packages/chain/cons/cons_test.go
jorgemmsilva: yeah, I've changed below so that it is assigned to `allRequests[i+1]`, so the initial request is kept.

Changed the loop to start at 0 so that the nonce calculation is correct: if `i` starts at 1, `i*reqPerBlock+ii` would yield 3  as the first nonce, instead of 0

packages/chain/mempool/mempool.go
jorgemmsilva: Fair point, I like that pattern :+1:  updated

packages/chain/cons/cons_test.go
```diff
@@ -283,26 +278,26 @@ func TestChained(t *testing.T) {
...
+	// if testing.Short() {
+	tests = []test{
+		{n: 1, f: 0, b: 10}, // Low N
+		{n: 2, f: 0, b: 10}, // Low N
+		{n: 3, f: 0, b: 10}, // Low N
+		{n: 4, f: 1, b: 10}, // Smallest possible resilient config.
+		{n: 10, f: 3, b: 5}, // Maybe a typical config.
+		{n: 12, f: 3, b: 3}, // Check a non-optimal N/F combinations.
+	}
+	// } else {
```
dessaya: is this intentionally left commented out?

dessaya (approved):

packages/chain/cons/cons_test.go
jorgemmsilva: ! Nice catch, thanks 👍

PR #2556, opened 2023-06-01
Variable length encoding for sizes in rwutil readers and writers

BugFreeSoftware:
Also cleaned up unused functions, and improved consistency in their usage


BugFreeSoftware:
Abandoned effort, switched to better solution using io.Reader and io.Writer

PR #2558, opened 2023-06-03
Alternate: remove marshalutil

BugFreeSoftware:


howjmay (approved):

contracts/native/inccounter/events.go
```diff
@@ -1,12 +1,14 @@
...
 	"github.com/iotaledger/wasp/packages/isc"
 	"github.com/iotaledger/wasp/packages/util"
 )
 
 func eventCounter(ctx isc.Sandbox, val int64) {
-	var buf []byte
-	buf = append(buf, util.Int64To8Bytes(val)...)
-	ctx.Event("inccounter.counter", buf)
+	w := new(bytes.Buffer)
+	_ = util.WriteInt64(w, val)
```
howjmay: maybe just not return here?

packages/chain/chainmanager/msg_cmt_log.go
```diff
@@ -40,22 +40,22 @@ func (msg *msgCmtLog) SetSender(sender gpa.NodeID) {
 }
 
 func (msg *msgCmtLog) MarshalBinary() ([]byte, error) {
-	w := bytes.NewBuffer([]byte{})
+	w := new(bytes.Buffer)
```
howjmay: Should we use `bytes.NewBuffer()` here?
https://pkg.go.dev/bytes#NewBuffer

BugFreeSoftware: bytes.NewBuffer is for a reader. Not for a writer.

dessaya: I think you can do `var w bytes.Buffer` to avoid an allocation.

BugFreeSoftware: We're not avoiding allocation. `new(bytes.Buffer)` allocates.
It's the new-style Go equivalent of `w := &bytes.Buffer{}`
Your version allocates a `Buffer` instead of a `*Buffer`, so all reader functions
would need to use `&w` instead of simply `w` like they do now.
There are a lot of subtle reasons why we should have gone with `io.Reader` and `io.Writer` all the way instead of the `marshalutil` functionality that awkwardly duplicates existing standard Go functionality, but misses the point w.r.t. streaming. Currently there is a hodgepodge of usages of both and a lot of usages and/or naming are inconsistent.
We should simply have the following standard for each structure:
```
func (x *X) Read(r io.Reader) error {}
func (x *X) Write(w io.Writer) error {}
func (x *X) Bytes() []byte {
    return util.BytesFromWriter(x.Write)
}
func XFromBytes(data []byte) (*X, error) {
   return util.BytesFromReader(data, new(X))
}
```
The latter 2 are one-liner functions that re-use the Read/Write functions.
This way we always have marshaling to/from `[]byte`, but also the stream functionality where necessary.
I guess we should have followed some basic standardization from the start, but most people were too busy learning Go on the fly.

BugFreeSoftware:
Superseded by other PRs

PR #2561, opened 2023-06-05
refactor: Add key length limit for trie kvstore

### LINKED ISSUE #2492: Set a hard cap to kvstore key length
in theory it's possible to create a VERY DEEP trie that will cause "stack overflow" when iterating the trie (example: `bufferedNode.commitNode`). 

By setting a hard cap to the length of keys, this attack vector can be eliminated.

howjmay:
closes #2492


jorgemmsilva (approved):
code looks good to me, just a question regarding the `256` value, how did we chose that number?

Issue #2529, opened 2023-05-30
event LUTs in blocklog are not cleaned on pruning


jorgemmsilva:
maybe those LUTs can be removed entirely if we use a cleaver key that allows to query by contract (say blockindex+hname+...)

lmoe:
We will follow up with the idea to get rid of the LUTs completely.
Functions depending on them should work without the LUTs, be removed, or replaced by something better.

This is a breaking change.

PR #2564, opened 2023-06-06
feat: add LUT pruning 

lmoe:
# Description of change
* Adds pruning for the event/request lookup table
* Adds test for the pruning methods
* Removes depguard linter rule
* Applies various linter fixes to support golangci-lint 1.53.2
  
fixes issue #2529 

## Type of change

`go test ./packages/vm/core/blocklog/blocklog_test.go`

### DIFF
--- a/packages/vm/core/blocklog/internal.go
+++ b/packages/vm/core/blocklog/internal.go
@@ -191,23 +191,47 @@ func getSmartContractEventsInternal(partition kv.KVStoreReader, contract isc.Hna
 	}
 }
 
-func pruneEventsByBlockIndex(partition kv.KVStore, blockIndex uint32, totalRequests uint16) {
-	// TODO what about these contract LUTs?
-	// scLut := collections.NewMap(partition, prefixSmartContractEventsLookup)
-	// we need to walk over all possible hContract values
-	// for each do a get of the lut
-	// for each lut scan and skip each block with a value < latestBlockIndex
-	// save the remaining part slice of the lut
+func pruneEventLookupByBlockIndex(partition kv.KVStore, blockIndex uint32) {
+	scLut := collections.NewMap(partition, prefixSmartContractEventsLookup)
 
+	scLut.Iterate(func(lutKey []byte, lutValue []byte) bool {
+		eventKeys := bytes.NewBuffer(lutValue)
+		newEventKeys := make([]byte, 0)
+
+		for {
+			eventKeyLookup, err := EventLookupKeyFromBytes(eventKeys)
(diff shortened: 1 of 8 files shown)


packages/vm/core/blocklog/internal.go
```diff
@@ -191,23 +191,47 @@ func getSmartContractEventsInternal(partition kv.KVStoreReader, contract isc.Hna
...
-	// TODO what about these contract LUTs?
-	// scLut := collections.NewMap(partition, prefixSmartContractEventsLookup)
-	// we need to walk over all possible hContract values
-	// for each do a get of the lut
-	// for each lut scan and skip each block with a value < latestBlockIndex
-	// save the remaining part slice of the lut
+func pruneEventLookupByBlockIndex(partition kv.KVStore, blockIndex uint32) {
+	scLut := collections.NewMap(partition, prefixSmartContractEventsLookup)
+
+	scLut.Iterate(func(lutKey []byte, lutValue []byte) bool {
```
dessaya, line 197: So this iterates the whole LUT? If I understand correctly it's one key for each ISC contract.

packages/vm/core/blocklog/internal.go
```diff
@@ -191,23 +191,47 @@ func getSmartContractEventsInternal(partition kv.KVStoreReader, contract isc.Hna
...
+				panic(err)
+			}
 
+			if eventKeyLookup.BlockIndex() != blockIndex {
+				eventKeyBytes := eventKeyLookup.Bytes()
+				newEventKeys = append(newEventKeys, eventKeyBytes...)
+			}
+		}
+
+		scLut.SetAt(lutKey, newEventKeys)
```
dessaya, line 217: And each value contains a blob with all the events generated by that SC? That could be thousands of items :open_mouth: 

Is this LUT really useful? Can we remove it, or at least refactor it so that is more efficient? Big loops inside SC code are not good.

packages/vm/core/blocklog/internal.go
lmoe: That is how I understood `Iterate` and the tests prove it. 
Each key is the contractHName and each value contains n items of event keys.

The tests show that both contract keys get correctly filtered.

kape1395:
If the LUT is part of the state, shouldn't we manage it from within a smart contract for all nodes to have the same state? Maybe I misunderstood something?

jorgemmsilva:
closing so this doesn't get merged by mistake.
We decided to go with another solution so this PR shouldn't apply anymore

Issue #2485, opened 2023-05-17
In-memory Trie-cache is too slow


fijter:
Replace the current cache we are using golang-lru with another solution and test.

fijter:
luke benchmarking several libraries against our testnet database.

luke-thorne:
@dessaya I put together some graphs of the data I was able to gather. I used a Zipfian-based generator to emulate a few regularly active addresses and mostly inactive addresses. 

# TLDR;
golang-lru and fastcache were remarkably similar. I'll replacing it with fastcache but other alternatives had significantly lower hitrates or higher memory usage. 

# ns/op

fastcache had the most consistent ns/op across the board regardless of the cache size.

![Image](https://github.com/iotaledger/wasp/assets/88324445/fbccc75f-4458-4d3e-95cd-4de0408efdfb)

# Hitrate

There was some variance when the cache was small but as the caches got bigger with bigger data sets the main options converged on 43%.

![Image](https://github.com/iotaledger/wasp/assets/88324445/ba5a2d61-7e62-40ab-ad65-b3fd8115fe76)

# Allocs/op

fastcache was the only option that didn't have a variance in allocs/op.

![Image](https://github.com/iotaledger/wasp/assets/88324445/caf50051-6c4f-477f-acb3-2c8e6a24e9ee)

# B/op

theine used the least B/op but it's a negligible difference. 

![Image](https://github.com/iotaledger/wasp/assets/88324445/d46f0572-99e5-4b76-bbd8-7051aa9b0dc3)

dessaya:
That's interesting! But does the benchmark generate concurrent accesses? The main bottleneck with golang-lru was caused by this lock: https://github.com/hashicorp/golang-lru/blob/master/lru.go#L96

https://github.com/hashicorp/golang-lru/issues/9

luke-thorne:
It does not but [fastcache](https://github.com/VictoriaMetrics/fastcache) is specifically designed to scale with the CPUs available and remain thread-safe. It also emphasizes reducing the GC overhead. They have some benchmarks with a few caches on their Github as well.

Yiling-J:
@luke-thorne I'm the author of Theine and I think there are some misconfigurations in your hit ratio benchmark.

https://github.com/iotaledger/wasp/blob/9851d05df83b033654409f5a11ee9186bbc9b5a2/packages/trie/test/trie_test.go#L723
Here Theine use size instead of (size * itemCost), but Set method 
[...]

PR #2565, opened 2023-06-06
Replace trie cache

luke-thorne:
# Description of change

Change trie cache from golang-lru to fastcache.
See #2485 for benchmarking details with some alternatives.

#2485 

## Type of change

- Ran benchmarks against several caches, the interface used with the cache does not change. 
- ran `make test`


dessaya (approved):
Awesome!

We are using golang-lru in 3 other places, should we change those as well? So that we use the same thing everywhere.

luke-thorne:
@dessaya 
I'm not sure how to convert `*state` to `[]byte` for `state.state`.

Can you help me with this function? [packages/state/store.go:50
](https://github.com/iotaledger/wasp/blob/76327bc136557710cbf8d8b3f22935d9864d1770/packages/state/store.go#L50)
```go
func (s *store) stateByTrieRoot(root trie.Hash) (*state, error) {
	if r := s.stateCache.GetBig(nil, root.Bytes()); r != nil {
		return &state{}, nil
	}
	r, err := newState(s.db, root)
	if err != nil {
		return nil, err
	}
	s.stateCache.SetBig(root.Bytes(), r)
	return r, nil
}
```

dessaya:
hmmm... it seems that in this case fastcache is not very useful, because it only allows to store `[]byte` values. We want to keep a cache of `*State` instances, not `[]byte`.

luke-thorne:
Thought about forking fastcache to allow setting types with generics but they base their whole cache on a ring buffer with encoded (k, v) pairs so the length of bytes in both the key and value is essential. Do we try fastcache in the other 2 places and leave lru here or just leave it all as lru?

I could fork clockpro and make it a typed cache like lru is but it is slower according to my benchmarks.
@dessaya

dessaya:
Let's use fastcache wherever it makes sense and keep golang-lru elsewhere. Thanks!

PR #2567, opened 2023-06-06
feat: Add method SetPayoutAddress to governance contract

### LINKED ISSUE #2559: Add a method on the governance contract that allows setting a payout account address (can be a EVM address too) for the fees going to the chain owner.
This will allow us to configure a fee payout handled by a on-chain dao or multisig wallet controlled by a ecosystem committee without giving away chain management permissions. The payout address should only be for the chain owner part and defaults to what it does right now, sending it to chain owner account if not set.

howjmay:
closes #2559


documentation/docs/guide/core_concepts/accounts/the-common-account.mdx
```diff
@@ -20,56 +20,3 @@ The common account is controlled by the chain owner defined in the chain [root c
 
 This account is used to store funds collected by fees, or sent to the chain's L1 address, or to invalid contracts (this
 way the funds are not permanently lost).
```
jorgemmsilva: this is not true anymore, since the common account is now only used to accumulate funds for minSD changes.

it would be good to add some explanation in the gov contract about what happens with the fees.

packages/vm/core/governance/governanceimpl/chainowner.go
```diff
@@ -52,6 +52,36 @@ func delegateChainOwnership(ctx isc.Sandbox) dict.Dict {
...
+
+func getMinSD(ctx isc.SandboxView) dict.Dict {
+	ret := dict.New()
+	minSD := ctx.StateR().Get(governance.StateVarMinSD)
+	if minSD == nil {
+		minSD = codec.EncodeUint64(governance.MinimumBaseTokensOnCommonAccount)
+	}
+	ret.Set(governance.ParamSetMinSD, minSD)
+	return ret
+}
```
jorgemmsilva: calling this `minSD` is not very descriptive.
What about something like `minCommonAccountBalance` or something like that?

packages/vm/core/governance/interface.go
```diff
@@ -85,6 +92,9 @@ const (
 
 // params
 const (
+	// MinimumBaseTokensOnCommonAccount can't harvest the minimum
+	MinimumBaseTokensOnCommonAccount = uint64(3000)
```
jorgemmsilva: this shouldn't be under `//params`, the name also doesn't make sense anymore since this is meant to be just a default value.
what about: `defaultMinCommonAccountBalance` ?

packages/vm/core/governance/internal.go
```diff
@@ -65,6 +65,21 @@ func MustGetChainInfo(state kv.KVStoreReader, chainID isc.ChainID) *isc.ChainInf
 	return info
 }
 
+func MustGetMinSD(state kv.KVStoreReader) uint64 {
+	d := kvdecoder.New(state)
+	minSD, err := d.GetUint64(StateVarMinSD)
+	if err != nil {
+		// FIXME we should use MinimumBaseTokensOnCommonAccount instead.
```
jorgemmsilva: `//FIXME`  here

but this could be done differently: set the default value in `governance.SetInitialState`, and error here if no value is available.

packages/vm/core/testcore/accounts_test.go
```diff
@@ -472,10 +450,8 @@ func TestFoundries(t *testing.T) {
 		err = ch.MintTokens(sn, allOtherTokens, senderKeyPair)
 		require.NoError(t, err)
 
-		// assert that no extra base tokens were used for the storage deposit
-		receipt := ch.LastReceipt()
 		commonAccountBalanceAfterLastMint := ch.L2CommonAccountBaseTokens()
-		require.Equal(t, commonAccountBalanceAfterLastMint, commonAccountBalanceBeforeLastMint+receipt.GasFeeCharged)
+		require.Equal(t, commonAccountBalanceAfterLastMint, commonAccountBalanceBeforeLastMint)
```
jorgemmsilva, line 453: could add an assertion that the fee was sent to the correct address.

packages/vm/vmcontext/runreq.go
```diff
@@ -363,13 +363,44 @@ func (vmctx *VMContext) chargeGasFee() {
...
-	vmctx.mustMoveBetweenAccounts(sender, accounts.CommonAccount(), transferToOwner)
+
+	// ensure common account has at least minSD, and transfer the rest of gas fee to payout address
+	// if the payout address is not set in governance contract, then chain owner will be used
+	var minSD uint64
+	vmctx.callCore(governance.Contract, func(s kv.KVStore) {
+		minSD = governance.MustGetMinSD(s)
+	})
+	commonAccountBal := vmctx.GetBaseTokensBalance(accounts.CommonAccount())
+	sendToPayout := sendToOwner
```
jorgemmsilva: this re-assignment here is weird, why not rename `sendToOwner` to `sendToPayout` instead?

packages/vm/vmcontext/runreq.go
```diff
@@ -363,13 +363,44 @@ func (vmctx *VMContext) chargeGasFee() {
...
+		vmctx.mustMoveBetweenAccounts(sender, accounts.CommonAccount(), transfer)
+	}
+	if sendToPayout > 0 {
+		var payoutAddr isc.AgentID
+		var err error
+		vmctx.callCore(governance.Contract, func(s kv.KVStore) {
+			payoutAddr, err = governance.GetPayoutAddress(s)
+			if err != nil {
+				payoutAddr = vmctx.chainOwnerID
+			}
```
jorgemmsilva: this is a weird one, error handling is not correct.

IMO:
 - `governance.GetPayoutAddress` shouldn't return an error, rather it should panic if something is wrong
 - if no `payoutAddr` is set, `nil` is returned and we fallback to `vmctx.chainOwnerID` in this case

packages/vm/core/governance/internal.go
howjmay: I can move the FIXME to `governance.SetInitialState`, but I don't think I can fix it right now. It relates to too much calculation it the tests, so it should be time-consuming to fix tests. I want to fix this in another PR. Do you think that is OK?

howjmay:
@jorgemmsilva BTW using `Address` in the `SetPayoutAddress` func, but the PayoutAddress is actually in AgentID type. Do you think this is OK? Or I should replace Address with AgentID?

documentation/docs/guide/core_concepts/accounts/the-common-account.mdx
howjmay: added

jorgemmsilva:
Good point, AgentID is a better choice

packages/vm/core/governance/internal.go
jorgemmsilva: I feel this should be part of this PR, what tests are you talking about in particular?

documentation/docs/guide/core_concepts/core_contracts/governance.md
```diff
@@ -321,3 +321,24 @@ A ratio between two values `x` and `y`, expressed as two `int32` numbers `a:b`,
...
+
+`GetPayoutAgentID` gets the payout AgentID.
+
+Returns the payout AgentID of the chain.
+
+- `s` (`AgentID`): the payout AgentID.
+
+### `SetMinCommonAccountBalance`
+
+`SetMinCommonAccountBalance` sets the minimum storage deposit.
```
jorgemmsilva: this is not exactly correct.
should be something like:
>sets the minimum balanced to be held in the common account

packages/vm/core/governance/internal.go
```diff
@@ -65,6 +67,25 @@ func MustGetChainInfo(state kv.KVStoreReader, chainID isc.ChainID) *isc.ChainInf
 	return info
 }
 
+func MustGetMinCommonAccountBalance(state kv.KVStoreReader) uint64 {
+	d := kvdecoder.New(state)
+	minCommonAccountBalance, err := d.GetUint64(StateVarMinCommonAccountBalance)
+	if err != nil {
+		// FIXME we should use DefaultMinCommonAccountBalance instead.
```
jorgemmsilva: fix me here

packages/vm/core/governance/governanceimpl/impl.go
```diff
@@ -56,4 +60,6 @@ func SetInitialState(state kv.KVStore, chainOwner isc.AgentID, blockKeepAmount i
 	state.Set(governance.VarGasLimitsBytes, gas.LimitsDefault.Bytes())
 	state.Set(governance.VarMaintenanceStatus, codec.Encode(false))
 	state.Set(governance.VarBlockKeepAmount, codec.EncodeInt32(blockKeepAmount))
+	// FIXME we should use MinimumBaseTokensOnCommonAccount instead.
+	// state.Set(governance.StateVarMinCommonAccountBalance, codec.EncodeUint64(governance.DefaultMinCommonAccountBalance))
```
jorgemmsilva: fixme here

packages/vm/core/governance/governanceimpl/chainowner.go
```diff
@@ -52,6 +52,36 @@ func delegateChainOwnership(ctx isc.Sandbox) dict.Dict {
...
+
+func getMinCommonAccountBalance(ctx isc.SandboxView) dict.Dict {
+	ret := dict.New()
+	minCommonAccountBalance := ctx.StateR().Get(governance.StateVarMinCommonAccountBalance)
+	if minCommonAccountBalance == nil {
+		minCommonAccountBalance = codec.EncodeUint64(governance.DefaultMinCommonAccountBalance)
+	}
+	ret.Set(governance.ParamSetMinCommonAccountBalance, minCommonAccountBalance)
+	return ret
+}
```
jorgemmsilva: if we set the default value at "initial state" this logic is not needed, we could just do:
```
return dict.Dict {
  governance.ParamSetMinCommonAccountBalance: ctx.StateR().Get(governance.StateVarMinCommonAccountBalance)
}
```

packages/vm/core/governance/internal.go
```diff
@@ -65,6 +67,25 @@ func MustGetChainInfo(state kv.KVStoreReader, chainID isc.ChainID) *isc.ChainInf
...
+	return minCommonAccountBalance
+}
+
+func MustGetPayoutAgentID(state kv.KVStoreReader) isc.AgentID {
+	d := kvdecoder.New(state)
+	agentID, err := d.GetAgentID(StateVarPayoutAgentID)
+	if err != nil && !errors.Is(err, codec.ErrNilAgentID) {
+		panic(err)
+	}
+	return agentID
```
jorgemmsilva: code can be reused on `getPayoutAgentID`, also we could use `MustGetAgentID` instead of the panic logic

packages/vm/core/testcore/accounts_test.go
```diff
@@ -830,46 +810,33 @@ func TestTransferAndHarvestNativeTokens(t *testing.T) {
...
-	v.req = solo.NewCallParams("accounts", "harvest").
-		WithGasBudget(100_000)
-	_, err = v.ch.PostRequestSync(v.req, v.chainOwner)
-	require.NoError(t, err)
-
-	rec := v.ch.LastReceipt()
-	t.Logf("receipt from the 'harvest' tx: %s", rec)
-
+	// FIXME we need to trigger the transfer part in runreq.go
+	t.Skip()
```
jorgemmsilva: fixme here

packages/vm/vmcontext/runreq.go
```diff
@@ -362,14 +363,44 @@ func (vmctx *VMContext) chargeGasFee() {
...
 	sender := vmctx.req.SenderAccount()
+	if sendToValidator != 0 {
+		transferToValidator := &isc.Assets{}
+		transferToValidator.BaseTokens = sendToValidator
+		vmctx.mustMoveBetweenAccounts(sender, vmctx.task.ValidatorFeeTarget, transferToValidator)
+	}
+
+	// ensure common account has at least minSD, and transfer the rest of gas fee to payout AgentID
+	// if the payout AgentID is not set in governance contract, then chain owner will be used
+	var minSD uint64
```
jorgemmsilva: minSD here is confusing, please rename to `minCommonAccountBalance`

jorgemmsilva (approved):
a few minor commends, and some "FIXME" still remaining

packages/vm/core/governance/internal.go
howjmay: That will be many Test2Chain tests will fail. 
Some listed in the CI here
https://github.com/iotaledger/wasp/actions/runs/5348139723/jobs/9697609590?pr=2567
https://github.com/iotaledger/wasp/actions/runs/5348139723/jobs/9697609932?pr=2567
So if this PR is no rush, then I will include this part in this PR

jorgemmsilva (approved):

PR #2575, opened 2023-06-08
simplify Read(io.Reader) and Write(io.Writer) marshaling

BugFreeSoftware:


packages/vm/core/governance/accessnodes.go
```diff
@@ -53,24 +53,15 @@ type AccessNodeInfo struct {
...
-	return &a, nil
+func NewAccessNodeInfoFromBytes(pubKey, data []byte) (*AccessNodeInfo, error) {
+	rr := rwutil.NewBytesReader(data)
+	return &AccessNodeInfo{
+		NodePubKey:    pubKey,
+		ValidatorAddr: rr.ReadBytes(),
+		Certificate:   rr.ReadBytes(),
+		ForCommittee:  rr.ReadBool(),
+		AccessAPI:     rr.ReadString(),
+	}, rr.Err
```
dessaya, line 64: Is the evaluation order guaranteed here? i.e. `rr.Err` is guaranteed to be evaluated after the rest?

dessaya (approved):

packages/vm/core/governance/accessnodes.go
BugFreeSoftware: Yes. in both cases. Within the struct the members are set in the order specified, even if they were defined in another order. And with return values each of the comma-separated values is evaluated left-to-right.

BugFreeSoftware: Same thing goes for function parameters.

PR #2577, opened 2023-06-09
Marshaling refactor for Message types

BugFreeSoftware:
Implements the new simplified marshaling into all gpa.Message types.
Also made a number of consistency improvements in naming and usage of the Message types.
And fixed potential bugs


packages/isc/agentid.go
```diff
@@ -5,29 +5,30 @@ package isc
...
 	iotago "github.com/iotaledger/iota.go/v3"
 	"github.com/iotaledger/wasp/packages/parameters"
+	"github.com/iotaledger/wasp/packages/util/rwutil"
 )
 
-type AgentIDKind uint8
-
 const (
-	AgentIDKindNil AgentIDKind = iota
+	AgentIDKindNil rwutil.Kind = iota
```
dessaya, line 18: Isn't there a way to keep the type `AgentIDKind`? It feels like losing some benefits of the static type checker if every different object must use the same `rwutil.Kind` type. (e.g. you could inadvertently do something wrong like `agentID.Kind() == MsgTypeBlockMessage` and the compiler will not tell you).

dessaya (approved):

packages/isc/agentid.go
BugFreeSoftware: I agree. I originally added the Kind type because those message types were all over the place and all were byte type. I wanted to make it easy to spot if they were used improperly. Went a bit overboard here. Will correct. Will also introduce a MessageKind in gpa.interface.

PR #2597, opened 2023-06-14
refactor: lookup tables

lmoe:
# Description of change

Merges extend/remove LUT branches.

- Event LUT is removed completely 
   - Lookup 'query' was moved into a getter function.
- Request LUT needs to remain and is now getting pruned with the other k/v pairs

## Type of change

The request LUT pruning and event fetching is tested in `packages/vm/core/blocklog/blocklog_test.go`

The removal of the event LUT was tested with `make test`.

### DIFF
--- a/packages/vm/core/blocklog/internal.go
+++ b/packages/vm/core/blocklog/internal.go
@@ -1,10 +1,8 @@
 package blocklog
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
-	"io"
 	"math"
 
 	iotago "github.com/iotaledger/iota.go/v3"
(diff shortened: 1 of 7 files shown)


packages/vm/core/blocklog/internal.go
```diff
@@ -165,30 +156,32 @@ func getRequestEventsInternal(partition kv.KVStoreReader, reqID isc.RequestID) (
...
-	var ret [][]byte
-	for {
-		key, err := EventLookupKeyFromBytes(keysBuf)
-		if err != nil {
-			if errors.Is(err, io.EOF) {
-				return ret, nil
-			}
-			return nil, fmt.Errorf("getSmartContractEventsIntern unable to parse key. %w", err)
+	filteredEvents := make([][]byte, 0)
+	events.Iterate(func(elemKey []byte, value []byte) bool {
```
jorgemmsilva: the events key is prefixed with block index:
https://github.com/iotaledger/wasp/blob/develop/packages/vm/core/blocklog/events.go#L11

this could be simplified if we leverage that, instead of iterating *everything*

dessaya: Agree. Not just for simplification, it is important not to do big loops in SC code.

packages/vm/core/blocklog/internal.go
```diff
@@ -230,6 +223,29 @@ func getRequestLogRecordsForBlockBin(partition kv.KVStoreReader, blockIndex uint
 	return ret, true
 }
 
+func pruneRequestLookupByBlockIndex(partition kv.KVStore, blockIndex uint32) {
+	lut := collections.NewMap(partition, prefixRequestLookupIndex)
+
+	lut.Iterate(func(lutKey []byte, lutValue []byte) bool {
```
dessaya: Same here. We should avoid iterating the whole LUT.

dessaya (approved):

PR #2598, opened 2023-06-14
State pruning in store

Juliusan:
# Description of change

Used pruning in state manager to prune old states from the store. Pruning is run every time state manager commits a new state to the store. It then checks, if more states, than needed, are stored in the DB and deletes the unneeded ones starting from the oldest states in order not to make pruned state gaps in case of failure. Each pruning attempt has an upper limit on number of states they are going to prune. Both the number of historic states to keep in store and maximum number of states to prune in one run can be configured (default is 10k states in store and max 1000 states to prune in one run). Moreover, `BlockKeepAmount` of governance contract is taken into account while determining the number of historic states to actually keep. The result is larger of the two values: the one from governance contract and the one from configuration. Metrics are added to measure pruning.

In addition to that:
* State manager parameters (block cache size, get block resend period etc..) can be configured.
* Comparing blocks is unified throughout state manager package. It should probably be part of Block functionality though.
* Maximum block index committed to the store metrics added to state manager.

## Type of change

All unit tests pass. Cluster tests fail more or less in the same way as in `develop` branch.


packages/chains/chains.go
```diff
@@ -51,8 +52,17 @@ type Chains struct {
...
+	walEnabled                          bool
+	walFolderPath                       string
+	smBlockCacheMaxSize                 int
+	smBlockCacheBlocksInCacheDuration   time.Duration
+	smBlockCacheBlockCleaningPeriod     time.Duration
+	smStateManagerGetBlockRetry         time.Duration
+	smStateManagerRequestCleaningPeriod time.Duration
+	smStateManagerTimerTickPeriod       time.Duration
+	smPruningMinStatesToKeep            int
+	smPruningMaxStatesToDelete          int
```
dessaya, line 65: Maybe group these into structs like `WalParams` and `SmParams`?

dessaya (approved):

packages/chains/chains.go
Juliusan: State manager parameters are grouped later on. There are just two wal parameters. I would not be against grouping them though.

PR #2641, opened 2023-06-28
fix:jsonrpc history

jorgemmsilva:
fix evm jsonrpc querying historical data on blocks that have already been pruned from the active state.

for now this relies on an in-memory index for txs and blocks by number/hash. This can be persisted to disk later (on a separate non-state-trie DB)

### DIFF
--- a/packages/evm/jsonrpc/jsonrpccache/cache.go
+++ b/packages/evm/jsonrpc/jsonrpccache/cache.go
@@ -1,0 +1,172 @@
+package jsonrpccache
+
+import (
+	"fmt"
+	"math/big"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+
+	"github.com/iotaledger/wasp/packages/state"
+	"github.com/iotaledger/wasp/packages/trie"
+	"github.com/iotaledger/wasp/packages/vm/core/blocklog"
+	"github.com/iotaledger/wasp/packages/vm/core/evm/emulator"
+	"github.com/iotaledger/wasp/packages/vm/core/governance"
+)
+
+type Cache struct {
+	lastBlockIndexCached *uint32
+	blockchainDB         func(chainState state.State) *emulator.BlockchainDB
+	blockTrieHashByIndex map[uint32]*trie.Hash
+	blockByHash          map[common.Hash]*types.Block
+	blockByNum           map[uint64]*types.Block
+	txBlockByHash        map[common.Hash]*types.Block
+	txsByBlockHash       map[common.Hash]types.Transactions
+	mutex                sync.RWMutex
+}
+
+func New(blockchainDB func(chainState state.State) *emulator.BlockchainDB) *Cache {
+	return &Cache{
+		lastBlockIndexCached: nil,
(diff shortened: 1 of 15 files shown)


packages/evm/jsonrpc/jsonrpccache/cache.go
```diff
@@ -0,0 +1,172 @@
...
+		c.blockTrieHashByIndex[i] = &blockTrieHash
+
+		evmBlock := db.GetCurrentBlock()
+		c.blockByHash[evmBlock.Hash()] = evmBlock
+
+		c.blockByNum[uint64(i)] = evmBlock
+		blockTransactions := evmBlock.Transactions()
+		c.txsByBlockHash[evmBlock.Hash()] = blockTransactions
+		for _, tx := range blockTransactions {
+			tx := tx // FIX FOR ITERATION VALUE SHARING
```
dessaya: Is this really needed here? `tx` isn't stored in a closure...

packages/vm/core/blocklog/stateaccess.go
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package blocklog
+
+import (
+	"github.com/iotaledger/wasp/packages/kv"
+	"github.com/iotaledger/wasp/packages/kv/subrealm"
+)
+
+type StateAccess struct {
```
dessaya, line 11: `StatePartition`, `StateSubrealm`?

btw, I like this pattern of having an object to encapsulate the contract state. Let's do that with all core contracts.

packages/evm/jsonrpc/jsonrpccache/cache.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (c *Cache) CacheBlock(trieRoot trie.Hash, stateByTrieRoot func(trieRoot trie.Hash) (state.State, error)) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	state, err := stateByTrieRoot(trieRoot)
+	if err != nil {
+		panic(err)
+	}
+	blockKeepAmount := governance.NewStateAccess(state).GetBlockKeepAmount()
```
dessaya, line 49: How does this work? `NewStateAccess()` returns an instance of `*StateAccess`, and I don't see that `StateAccess` has a `GetBlockKeepAmount` method :thinking:

dessaya (approved):

packages/evm/jsonrpc/jsonrpccache/cache.go
jorgemmsilva: that's right, thanks :+1:

packages/vm/core/blocklog/stateaccess.go
jorgemmsilva: Yeah, trend started by Karolis, I like it as well!
right now there are some implemented for core contracts and all are called `StateAccess`, so I'll keep the naming consistent for now

packages/evm/jsonrpc/jsonrpccache/cache.go
jorgemmsilva: this one is from `governance` :stuck_out_tongue:

PR #2642, opened 2023-06-28
Refactor VMTask & VMContext

dessaya:
This is the first step of a longer refactor I have in mind.

Here I just split `VMTask` and `VMContext` into smaller objects. After this I plan to change the logic into a more functional style.


packages/vm/vmtask.go
```diff
@@ -14,51 +14,48 @@ import (
...
 	EstimateGasMode bool
-	// If EVMtrace is set, all requests will be executed normally up until the EVM
+	// If EVMTracer is set, all requests will be executed normally up until the EVM
 	// tx with the given index, which will then be executed with the given tracer.
 	EVMTracer            *isc.EVMTracer
 	EnableGasBurnLogging bool // for testing and Solo only
-
-	// INPUTS_OUTPUTS:
+	// If maintenance mode is enabled, only requests to the governance contract will be executed
+	MaintenanceModeEnabled bool
```
jorgemmsilva, line 40: I'm thinking whether it even makes sense that this is passed as a task parameter, the vm can get it from the state :thinking:

jorgemmsilva (approved):

packages/vm/vmtask.go
dessaya: Good point! I'll take a look and fix in the next PR.

PR #2655, opened 2023-07-02
feat: add ledger, stronghold and keychain into cli

lmoe:
# Description of change

Adds a more secure layer into the wasp-cli.

It is based on the [wasp-wallet-sdk library](https://github.com/iotaledger/wasp-wallet-sdk/tree/main)
Which is based on the [iota-sdk native library](https://github.com/iotaledger/iota-sdk-native-bindings)

This feature introduces Ledger and Stronghold wallet providers.
 
Alternatively, the user can choose to remain using the common seed - but it will now be stored in the OS Keychain instead of the `wasp-cli.json`. The seed will be loaded into a more secure encrypted in-memory vault at runtime. 

Starting from this version, each user is required to have the `iota_sdk` library next to the `wasp-cli`. Prebuilt releases are available [here](https://github.com/iotaledger/iota-sdk-native-bindings/releases) and are shipped with `wasp-cli` releases as well.

## Type of change

The most delicate parts are in the two referenced dependencies. 

The Go library has tests, which are to be found [here](https://github.com/iotaledger/wasp-wallet-sdk/tree/main/test)
The Rust native lib does not have tests yet.


packages/cryptolib/keypair.go
```diff
@@ -52,6 +52,23 @@ func (k *KeyPair) GetPublicKey() *PublicKey {
 	return k.publicKey
 }
 
+func (k *KeyPair) SignBytes(data []byte) []byte {
+	return k.GetPrivateKey().Sign(data)
+}
+
+func (k *KeyPair) Sign(addr iotago.Address, payload []byte) (iotago.Signature, error) {
+	// TODO: Validate this
```
dessaya: So... who is going to validate this? :rofl:

tools/wasp-cli/cli/wallet/providers/inmemory.go
```diff
@@ -0,0 +1,43 @@
...
+	return newInMemoryWallet(keyPair, addressIndex)
+}
+
+func CreateNewInMemory() {
+	keyChain := config.NewKeyChain()
+	seed := cryptolib.NewSeed()
+	err := keyChain.SetSeed(seed)
+	log.Check(err)
+
+	log.Printf("In memory seed saved in the keychain.\n")
```
dessaya:
```suggestion
	log.Printf("In-memory seed saved in the keychain.\n")
```

dessaya (approved):

Dr-Electron:
After some weeks of working with this branch I didn't really have any major problems.

Some notes tho:
- If I run `init -c wasp-cli.json` no config is created. It uses my keychain as default. But maybe a wasp-cli.json with that info should at least get generated. This allows me to specify a different config path with the -c option on `init` so that I can use that file for all future commands
- Even if I change the scheme to stronghold the cli still seems to look for the keychain and store stuff there. Everything is working with stronghold just needs som cleanup I guess to not check for keychain stuff 🤔

packages/cryptolib/keypair.go
lmoe: I did! Seems to work fine. :D

PR #2657, opened 2023-07-03
fix: prevent proposing batches with 1 random request

jorgemmsilva:
there was an edge case where nodes could propose batches with a single request that could not be executed (always skipped). This could deadlock the consensus logic.

also contains the logic from https://github.com/iotaledger/wasp/pull/2656, that should be merged first


packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -77,6 +77,12 @@ func createEmulator(ctx isc.Sandbox, l2Balance *l2Balance) *emulator.EVMEmulator
 
 // IMPORTANT: Must only be called from the ISC VM (when the request is done executing)
 func AddFailedTx(ctx isc.Sandbox, tx *types.Transaction, receipt *types.Receipt) {
+	if tx == nil {
+		panic("nil tx")
+	}
+	if receipt == nil {
+		panic("nil receipt")
+	}
```
dessaya, line 85: This left intentionally? Seems like something added just for diagnosing a specific issue.

dessaya (approved):
Code looks ok, no idea about the logic though :stuck_out_tongue:

packages/vm/core/evm/evmimpl/internal.go
jorgemmsilva: it was left on purpose, I spend a bunch of time trying to figure out where the "segfault" came from. our "panic handling magic" makes it very hard to follow the stack traces.
basically, this can happen when a  request with a bad nonce enters the VM, EVM panics, but no receipt is produced.

packages/chain/mempool/mempool.go
```diff
@@ -590,20 +590,26 @@ func (mpi *mempoolImpl) handleConsensusProposalForChainHead(recv *reqConsensusPr
...
+		recv.Respond(refs)
 		return
 	}
+
 	//
 	// Wait for any request.
-	mpi.waitReq.WaitAny(recv.ctx, func(req isc.Request) {
-		recv.Respond([]*isc.RequestRef{isc.RequestRefFromRequest(req)})
+	mpi.waitReq.WaitAny(recv.ctx, func(_ isc.Request) {
+		mpi.handleConsensusProposalForChainHead(recv)
```
kape1395, line 612: Why is this loop needed?

PR #2659, opened 2023-07-03
Snapshots

Juliusan:
# Description of change

Quite a large update, although it is supposed not to break the current chains. Haven't tested it enough, therefore just a draft. I hope to test it more when cluster tests are fixed in `develop`.

Current functionality:
* Allows creating snapshots. If `snapshots.period` config parameter is larger than 0, the node creates a snapshot in location named in `snapshots.localPath` every `snapshots.period` state.
* Uses snapshots from `snapshots.localPath` or from network locations mentioned in `snapshots.networkPaths` to avoid downloading all the blocks since origin.

Utilities, that may be useful in other places of Wasp:
* `readOnlyStore` - store, which panics if anything tries to edit its contents. Useful in tests, if you want to use the store for reading and you want to control who can edit the store.
* Methods in `packages/chain/statemanager/sm_gpa/sm_gpa_utils/test_utils.go`. Sometimes you must be sure that block/state, that you are getting in tests is exactly what was expected. IMO, `Block` and `State` should have `Equals` method, but for now I use these utilities.
* ~~`blockingReader` - see comment in the code. If there is no error, it always reads all the requested bytes possibly blocking to perform that. It can be used to decorate any reader and is much more convenient than checking if all the bytes are read on every read and figuring what to do, if it is not the case.~~ Removed after suggestions from Diego. Replaced by `io.CopyN`.
* `SliceStruct` - see comment in the code. Simply to have a single object, which can be appended to. Golang's `append` function might create another object, which is not always convenient.
* `SnapshotInfo` (renaming it, if used in other places) - just a way to have one object for block/state index and L1 commitment.

Known drawbacks:
* Does not use Eric's refactored marshaling/unmarshaling, because the core was developed prior to that.

## Type of change

- Enhancement (a non-breaking (hopefully) change which adds functionality)

## Documentation

Documented the changes in this PR https://github.com/iotaledger/iota-wiki/pull/1033. I hope, it will soon get merged to the main branch.


packages/chain/statemanager/sm_gpa/sm_gpa_utils/read_only_store.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/iotaledger/hive.go/kvstore"
+	"github.com/iotaledger/wasp/packages/state"
+	"github.com/iotaledger/wasp/packages/trie"
+)
+
+type readOnlyStore struct {
+	store state.Store
+}
+
+var _ state.Store = &readOnlyStore{}
```
dessaya, line 16: Wouldn't it be better to divide the Store interface into smaller interfaces? Like:

```
type ReadOnlyStore interface {
  HasTrieRoot(trieRoot trie.Hash) bool
  BlockByTrieRoot(trieRoot trie.Hash) (state.Block, error)
  ...
}

type Store interface {
  ReadOnlyStore
  SetLatest(trie.Hash) error
  NewOriginStateDraft() state.StateDraft
  ...
}
```

This way the default implementation of `Store` already satisfies `ReadonlyStore`. In your code, if you know that you will never need a full `Store` you can just receive a `ReadOnlyStore`.

packages/chain/statemanager/sm_gpa/state_manager_output.go
```diff
@@ -0,0 +1,67 @@
+//
+//
+//
+//
+//
+//
```
dessaya: ?

packages/chain/statemanager/sm_snapshots/blocking_reader.go
```diff
@@ -0,0 +1,46 @@
+package sm_snapshots
+
+import (
+	"errors"
+	"io"
+)
+
+type blockingReader struct {
```
dessaya: Maybe this can be replaced or reimplemented with https://pkg.go.dev/io#CopyN ?

packages/chain/statemanager/sm_snapshots/snapshotter.go
```diff
@@ -0,0 +1,190 @@
+//
+//
+//
+//
+//
+//
```
dessaya: ?

dessaya (approved):

packages/chain/statemanager/sm_gpa/sm_gpa_utils/read_only_store.go
Juliusan: Well, not really. In these cases I need an object, which looks as a regular `Store`, just behaves differently. E.g., In test I create a `SnapshotManager` which in general should be able to both read and write in to the store, but in this particular test I want to make sure that the code does not mess with the provided store as writing to it is not needed in this case.

Like here: https://github.com/iotaledger/wasp/pull/2659/files#diff-e424bfd427fb0fb6c529ea32f560d0972ef353cd7c0d176b457ec932417ac791R84-R85
`BlockFactory` is a special object, which uses its store to generate correct block sequences. `snapshotManagerOrig` in this test is needed just to read the store and produce snapshot files, which will be read by other snapshot manager. So although creating a new snapshot manager requires a full featured `Store` object, `BlockFactory` would like to make sure, that if anything uses its store, it's just for reading. Another approach would be to create a new store and somehow copy the contents. This would avoid `ReadOnlyStore`, but would be too much of a hassle IMO.

packages/chain/statemanager/sm_snapshots/blocking_reader.go
Juliusan: Ok, that works. Thanks for the reference :-)

packages/chain/statemanager/sm_snapshots/slice_struct.go
```diff
@@ -0,0 +1,69 @@
+package sm_snapshots
+
+import (
+	"github.com/samber/lo"
+)
+
+type sliceStructImpl[E any] struct {
```
jorgemmsilva, line 7: generic, reusable stuff like this should be in a package of it's own, rather than inside `sm_snapshots`.

the interface should also be part of this file, but I'd argue there is no need to create an interface for this at all, since there is only 1 implementation.

---


now, why is this needed at all?

to avoid stuff like this?
```
map[key] = append(someslice, element)
```

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
```diff
@@ -0,0 +1,439 @@
...
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return cancelFun, nil, fmt.Errorf("http request to %s got status code %v", url, response.StatusCode)
+	}
+
+	progressReporter := NewProgressReporter(log, fmt.Sprintf("downloading file %s", url), uint64(response.ContentLength))
+	reader := io.TeeReader(response.Body, progressReporter)
+	return cancelFun, reader, nil
+}
```
jorgemmsilva, line 418: I don't understand why the responsibility of calling `cancelFund` is assigned to the caller.

also why do we "tee" the response body stream

IMO this function could just  read the entire response.body into memory, close the response.body stream, close the "downloadCtx", and return the body bytes, is this not possible?

packages/chain/statemanager/sm_snapshots/interface.go
```diff
@@ -0,0 +1,57 @@
...
+	createSnapshotsNeeded() bool
+	handleUpdate()
+	handleBlockCommitted(SnapshotInfo)
+	handleLoadSnapshot(SnapshotInfo, chan<- error)
+}
+
+type snapshotter interface {
+	storeSnapshot(SnapshotInfo, io.Writer) error
+	loadSnapshot(SnapshotInfo, io.Reader) error
+}
```
jorgemmsilva, line 43: only a single implementation exists, I'd say we can remove this interface

packages/chain/statemanager/sm_snapshots/interface.go
```diff
@@ -0,0 +1,57 @@
...
+	SnapshotExists(uint32, *state.L1Commitment) bool
+	LoadSnapshotAsync(SnapshotInfo) <-chan error
+}
+
+type SnapshotManagerTest interface {
+	SnapshotManager
+	SnapshotReady(SnapshotInfo)
+	IsSnapshotReady(SnapshotInfo) bool
+	SetAfterSnapshotCreated(func(SnapshotInfo))
+}
```
jorgemmsilva, line 22: feels like this belongs in a test file/package?

packages/chain/statemanager/sm_snapshots/snapshot_info.go
```diff
@@ -0,0 +1,52 @@
...
+	return si.commitment
+}
+
+func (si *snapshotInfoImpl) GetTrieRoot() trie.Hash {
+	return si.GetCommitment().TrieRoot()
+}
+
+func (si *snapshotInfoImpl) GetBlockHash() state.BlockHash {
+	return si.GetCommitment().BlockHash()
+}
```
jorgemmsilva, line 38: I would drop the `Get` prefixes, seems superfluous

https://go.dev/doc/effective_go

> Go doesn't provide automatic support for getters and setters. There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner.

packages/chain/statemanager/sm_snapshots/snapshot_manager_mocked.go
```diff
@@ -0,0 +1,175 @@
+package sm_snapshots
```
jorgemmsilva, line 1: would be nice to have these test-specific functions/structs outside of the code. we can keep them in the same dir, just declare the package as `sm_snapshots_test`

https://pkg.go.dev/testing

packages/chain/statemanager/sm_snapshots/snapshot_manager_runner.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/iotaledger/wasp/packages/shutdown"
+	"github.com/iotaledger/wasp/packages/util/pipe"
+)
+
+type snapshotInfoCallback struct {
+	SnapshotInfo
+	callback chan<- error
+}
+
+type snapshotManagerRunner struct {
```
jorgemmsilva, line 16: Is there any high-level description of why we chose to go with so many different components?
 - manager
 - manager runner
 - snapshotter

what is the idea for this design?

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
```diff
@@ -0,0 +1,439 @@
...
+	lastIndexSnapshotted      uint32
+	lastIndexSnapshottedMutex sync.Mutex
+	createPeriod              uint32
+	snapshotter               snapshotter
+
+	availableSnapshots      *shrinkingmap.ShrinkingMap[uint32, SliceStruct[*commitmentSources]]
+	availableSnapshotsMutex sync.RWMutex
+
+	localPath    string
+	networkPaths []string
```
jorgemmsilva, line 46: was the networkPaths a requirement? What is the motivation for having this feature?

we could rely on mounting network drives (like smb) and let the OS/filesystem do its thing.

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
```diff
@@ -0,0 +1,439 @@
...
+			}
+			smiT.lastIndexSnapshottedMutex.Unlock()
+			smiT.log.Infof("Creating snapshot %v %s: snapshot created in %s", stateIndex, commitment, finalFilePath)
+		}()
+	}
+}
+
+func (smiT *snapshotManagerImpl) handleLoadSnapshot(snapshotInfo SnapshotInfo, callback chan<- error) {
+	smiT.log.Debugf("Loading snapshot %s", snapshotInfo)
+	// smiT.availableSnapshotsMutex.RLock() // Probably locking is not needed as it happens on the same thread as editing available snapshots
```
jorgemmsilva, line 206: enable mutex lock/unlock?

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
```diff
@@ -0,0 +1,439 @@
...
+	}
+	callback <- err
+}
+
+// -------------------------------------
+// Internal functions
+// -------------------------------------
+
+func (smiT *snapshotManagerImpl) cleanTempFiles() {
+	tempFileRegExp := tempSnapshotFileNameString("*", "*")
```
jorgemmsilva, line 277: do the tmp files also need to be protected with a mutex?

jorgemmsilva (approved):

jorgemmsilva:
merging to kick off the release, lets discuss the comments later

packages/chain/statemanager/sm_snapshots/slice_struct.go
Juliusan: Yes. That's the main reason. I wasn't sure if it is useful in other places of wasp, so left them there. Will move to more general location.

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
Juliusan: `TeeReader` here is used to log download status. I copied the idea from Hornet: https://github.com/iotaledger/hornet/blob/develop/pkg/snapshot/download.go#L278.

When I was designing the algorithm, I had in mind that snapshot can be quite large (e.g., several Gb). This might not be the case now when pruning is implemented. However, as was originally intended snapshot is read straight into `KVStore` instead of using intermediary byte slice. In such case closing should be done once everything is read (or read error occurs). This approach can possibly be optimised more by having `RestoreSnapshot` variant in `Store`, which takes some kind of key-value reader object as parameter.

packages/chain/statemanager/sm_snapshots/slice_struct.go
jorgemmsilva: I'd say this is an overkill abstraction, I don't see any problem with doing the assignment above

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
jorgemmsilva: okay, if we're downloading entire snapshots via http, we should probably use `multipart`
https://pkg.go.dev/mime/multipart#Part.Read

packages/chain/statemanager/sm_snapshots/interface.go
Juliusan: Two reasons:
* I want to somehow make difference between public and private members of `snapshotter` object, which itself is private. Golang sometimes reacts funnily to public functions of private structs and in fact it doesn't have private members - everything, that is called "private" is in fact package private. Having a separate interface I have more control (over myself :-D): I know, that anybody having `snapshotter` interface object is not going to call inner methods even in the same package. This way I can separate public code from private (not package private) code.
* I hate dealing with "reference/not a reference" mess, so I enjoy using interfaces more :-)

If these reasons (especially the first one) is not enough, I can remove the interface.

packages/chain/statemanager/sm_snapshots/interface.go
Juliusan: Will separate it to another file. I'll see, what I can do with separate package.

packages/chain/statemanager/sm_snapshots/snapshot_info.go
Juliusan: I don't agree that this is more clear, but if this is Go way of doing things, I'll change that.

packages/chain/statemanager/sm_snapshots/snapshot_manager_mocked.go
Juliusan: I never understood what is the purpose of separate test package. You cannot use private structs/methods of `sm_snapshots` in `sm_snapshots_tests`. So as far as I see it, it makes testing more difficult with no real benefit. I'll see, if I can have a separate package for tests in this occasion.

jorgemmsilva: You can test your package as a blackbox

packages/chain/statemanager/sm_snapshots/snapshot_manager_runner.go
Juliusan: It evolved during development.

`snapshotter` is a component, which is responsible only for taking required snapshot from the store and writing it to some `Writer` and reading the snapshot from some `Reader` and putting it to the store. It is an important part of snapshot functionality that can be developed and tested separately.

`SnapshotManager` is responsible for providing `snapshotter` service to state manager. It deals with synchronising/parallelising the requests, searches for available snapshots, deals with where to store snapshots etc. 

After developing `SnapshotManager`, I needed to mock it in State manager tests. I understood, that some parts of the code (synchronising/parallelising requests) will have to be duplicated. To avoid that, I extracted these parts into `snapshotManagerRunner` and used this component in `snapshotManagerImpl` and `mockedSnapshotManager`.

I'll document these things in the code. If it is not clear now, it will definitely be less clear later.

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
Juliusan: No. It wasn't. I tried to do this part similar to Hornet. Of course, we could use mounted network drives. It will simplify the Wasp code. And avoid additional development, which is needed in HTTP(S) functionality such as implementing continuation of broken download. We can discuss this in the next meeting.

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
Juliusan: 1. It's a read lock and
2. It happens in the same thread as code, locking the object for writing.
So I think there is no point in wasting time to obtain/return the lock. The only thing, that it can clash with, is another read operations. Or am I wrong here?

Juliusan: I left the comment in order not to think about it as an error, when someone (probably me) is going to find this case later.

packages/chain/statemanager/sm_snapshots/snapshot_manager.go
Juliusan: Cleaning temp files happens only at the start of the node. It shouldn't take long - maybe deleting one or two files. It is ensured, that this function is going to be completed, before snapshot manager is even started, so no temp files can be created during running this function.

PR #2681, opened 2023-07-06
refactor: Remove duplicated WithAllowance

howjmay:
The behavior of `WithAllowance` and `AddAllowance` are almost the same. Maybe just keep one

### DIFF
--- a/packages/solo/req.go
+++ b/packages/solo/req.go
@@ -73,11 +73,6 @@ func CallParamsFromDictByHname(hContract, hFunction isc.Hname, par dict.Dict) *C
 	return ret
 }
 
-func (r *CallParams) WithAllowance(allowance *isc.Assets) *CallParams {
-	r.allowance = allowance.Clone()
-	return r
-}
-
 func (r *CallParams) AddAllowance(allowance *isc.Assets) *CallParams {
 	if r.allowance == nil {
 		r.allowance = allowance.Clone()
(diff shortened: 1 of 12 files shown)


packages/solo/req.go
```diff
@@ -398,9 +393,10 @@ func (ch *Chain) EstimateGasOnLedger(req *CallParams, keyPair *cryptolib.KeyPair
 
 // EstimateGasOffLedger executes the given on-ledger request without committing
 // any changes in the ledger. It returns the amount of gas consumed.
-// if useFakeBalance is `true` the request will be executed as if the sender had enough base tokens to cover the maximum gas allowed
+// if useMaxBalance is `true` the request will be executed as if the sender had enough base tokens to cover the maximum gas allowed
 // WARNING: Gas estimation is just an "estimate", there is no guarantees that the real call will bear the same cost, due to the turing-completeness of smart contracts
-func (ch *Chain) EstimateGasOffLedger(req *CallParams, keyPair *cryptolib.KeyPair, useMaxBalance ...bool) (gas, gasFee uint64, err error) {
+func (ch *Chain) EstimateGasOffLedger(_req *CallParams, keyPair *cryptolib.KeyPair, useMaxBalance ...bool) (gas, gasFee uint64, err error) {
```
howjmay: I think the argument `_req` shouldn't be change after executing the function.

jorgemmsilva (changes requested):
I don't understand the motivation for this PR

`AddAllowance` and `WithAllowance` do different things, right? (`add` adds and `with` sets/overrides the current allowance)

jorgemmsilva:
closing for now, we can re-open if needed

packages/solo/req.go
howjmay: @jorgemmsilva Do you think the change for this function is reasonable?

PR #2714, opened 2023-07-12
refactor: more VM cleanup

dessaya:
Some highlights:

* Inverted the relationship between `vmContext` and `requestContext`. Now `requestContext` is the one that provides the sandbox interface and contains the call stack. This ensures that it is not possible to create a sandbox outside of a request context.
* `vmctx.callCore()` outside of a request context executed an unnecessary push/pop in the stack, looking for the core contract registry, etc. This is replaced with `withContractState()` which just creates a temporary state subrealm.
* Removed the "block context subscription" mechanism that was created in a generic way but used only by EVM.
* Replaced the `chainStateWrapper` object with a much simpler `kvStoreWithGasBurn`. Renamed some other functions to make it more explicit when kv access produces gas burn.
* `vmContext::currentStateUpdate` is moved to `requestContext::uncommittedState`. The `withStateUpdate()` function now passes the state as a parameter to its callback.


packages/evm/jsonrpc/evmchain.go
```diff
@@ -701,29 +701,28 @@ func blockchainDB(chainState state.State) *emulator.BlockchainDB {
...
-	)
 	accountsPartition := subrealm.NewReadOnly(chainState, kv.Key(accounts.Contract.Hname().Bytes()))
 	return emulator.NewStateDB(
-		buffered.NewBufferedKVStore(sdbPartition),
+		buffered.NewBufferedKVStore(evmStateSubrealmR(chainState)),
 		newL2Balance(accountsPartition),
 	)
 }
 
+func evmStateSubrealmR(chainState state.State) kv.KVStoreReader {
```
jorgemmsilva: there is already this one: https://github.com/iotaledger/wasp/blob/develop/packages/vm/core/evm/evmimpl/state.go#L16
that gets the "sub state from the EVM contract state", its used on `evmimpl.StateAccess`

this has the same name and is slightly different, that's a bit confusing

packages/testutil/testdbhash/TestInitLoad.hex
```diff
@@ -1 +1 @@
-0xaa44833634548eedaa3711ea40b0e2eca9e65392d70a0716f2418061d33ce7d5
```
jorgemmsilva, line 1: I assume this is because of the origin state and this PR does not a migration, right?

packages/vm/execution/interface.go
```diff
@@ -36,7 +43,7 @@ type WaspContext interface {
 	ChainInfo() *isc.ChainInfo
 	CurrentContractHname() isc.Hname
 	Params() *isc.Params
-	StateReader() kv.KVStoreReader
+	ContractStateReaderWithGasBurn() kv.KVStoreReader
```
jorgemmsilva, line 46: Why "withGasBurn" ? isn't that implicit?

packages/vm/vmimpl/gas.go
```diff
@@ -6,44 +6,45 @@ import (
...
-	vmctx.reqctx.gas.burnLog.Record(burnCode, g)
-	vmctx.reqctx.gas.burned += g
+	reqctx.gas.burnLog.Record(burnCode, g)
+	reqctx.gas.burned += g
 
-	if vmctx.reqctx.gas.burned > vmctx.reqctx.gas.budgetAdjusted {
-		vmctx.reqctx.gas.burned = vmctx.reqctx.gas.budgetAdjusted // do not charge more than the limit set by the request
+	if reqctx.gas.burned > reqctx.gas.budgetAdjusted {
+		reqctx.gas.burned = reqctx.gas.budgetAdjusted // do not charge more than the limit set by the request
+		// debug.PrintStack()
```
jorgemmsilva:
```suggestion
```

packages/vm/vmimpl/vmcontext.go
```diff
@@ -31,25 +31,22 @@ import (
...
 	feeCharged uint64
-	// is gas burn enabled
-	// TODO: should be in requestGas?
-	burnEnabled bool
 }
 
 type requestContext struct {
+	vm *vmContext
+
+	uncommittedState  *buffered.BufferedKVStore
```
jorgemmsilva, line 48: I wonder if it makes sense to have the uncommitted state here, feels like this should be a layer above and `requestContext` would only have access to Get/Set? just a thought

packages/vm/vmimpl/skipreq.go
```diff
@@ -23,33 +23,33 @@ const (
...
 }
 
 // checkReasonRequestProcessed checks if request ID is already in the blocklog
-func (vmctx *vmContext) checkReasonRequestProcessed() error {
-	reqid := vmctx.reqctx.req.ID()
+func (reqctx *requestContext) checkReasonRequestProcessed() error {
+	reqid := reqctx.req.ID()
 	var isProcessed bool
-	vmctx.callCore(blocklog.Contract, func(s kv.KVStore) {
+	withContractState(reqctx.uncommittedState, blocklog.Contract, func(s kv.KVStore) {
```
jorgemmsilva, line 52: there are some of these `withContractState(reqctx.uncommittedState,.....` shouldn't all of these use the "withGas" version of the store? we already control gas burn with the "gasBurnEnabled" flag

jorgemmsilva (commented):
great changes! :clap: 

just please make sure the cluster tests pass before merging :pray:

jorgemmsilva (approved):

packages/testutil/testdbhash/TestInitLoad.hex
dessaya: Yes, it's because I removed the block context subscriptions. It shouldn't need a migration because at most it will leave some small garbage in the state.

packages/vm/execution/interface.go
dessaya: Well... "explicit is better than implicit" :stuck_out_tongue: 

While refactoring I was confused about when gas burning was in effect... so I decided to make it explicit. Note that this renaming only affects the `WaspContext` but not the `Sandbox`.

packages/vm/vmimpl/vmcontext.go
dessaya: Maybe! The truth is that I'm still far from happy with the results... I mean, it's an improvement but it still convoluted and in need of a lot of work. I'll think about this and other changes I have in mind, but at some point I'll have to accept that it will never be perfect :grin:

packages/vm/vmimpl/skipreq.go
dessaya: I don't know, maybe? This is related to what I mentioned above. I find this gasBurnEnabled flag very confusing. Ideally in my mind all functions that need gas would receive the gas context by parameter, but this being in the state far away makes it very hard to reason about IMO. That's why I decided to avoid using "withGas" if I *knew* that I don't need gas.

dessaya: In other words: "explicit is better than implicit" ... using `uncommittedState` is explicitly saying "I don't use gas here".

jorgemmsilva: yeah I agree, that "stateful" flag is not good.
my feeling is that we should always assume that gas burn is always enabled, maybe provide a function that allows some code to run without charging gas? `doWithoutChargingGas(f func()...)` just an idea

PR #2723, opened 2023-07-16
refactor: Return status code 405 when auth is disabled

### LINKED ISSUE #2722: login route gives 404 when auth is disabled
should give something like: 405 Method Not Allowed

howjmay:
closes https://github.com/iotaledger/wasp/issues/2722


jorgemmsilva (approved):

lmoe:
The old behavior was simply an oversight.

Auth Status returns a JSON object containing the auth strategy. The strategy name should be 'none' if there is no auth enabled, rather than returning a 405 imo.

lmoe (approved):

PR #2731, opened 2023-07-19
refactor(vm): EVM gas + error handling

dessaya:
* Instead of storing the EVM tx and receipt in the `requestContext`, now the EVM contract registers a callback, which results in simpler and more generic code:

```
	ctx.Privileged().OnWriteReceipt(func(evmPartition kv.KVStore, chainInfo *isc.ChainInfo) {
		saveExecutedTx(evmPartition, chainInfo, tx, receipt)
	})
```

* Simplified EVM gas burning: the emulator is no longer responsible for burning ISC gas; it is done by the evm core contract.

* Simplified EVM error handling: the emulator catches all ISC panics in a single place and always produces an execution result with the correct gas burned amount and error message (instead of a generic "execution reverted").

### DIFF
--- a/packages/isc/sandbox_interface.go
+++ b/packages/isc/sandbox_interface.go
@@ -7,7 +7,6 @@ import (
 	"math/big"
 	"time"
 
-	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/eth/tracers"
 
 	iotago "github.com/iotaledger/iota.go/v3"
@@ -139,11 +138,11 @@ type Privileged interface {
 	DebitFromAccount(AgentID, *Assets)
 	CreditToAccount(AgentID, *Assets)
 	RetryUnprocessable(req Request, blockIndex uint32, outputIndex uint16)
-
-	// EVM
+	OnWriteReceipt(CoreCallbackFunc)
 	CallOnBehalfOf(caller AgentID, target, entryPoint Hname, params dict.Dict, allowance *Assets) dict.Dict
-	SetEVMFailed(*types.Transaction, *types.Receipt)
 }
+
+type CoreCallbackFunc func(contractPartition kv.KVStore)
 
 // RequestParameters represents parameters of the on-ledger request. The output is build from these parameters
 type RequestParameters struct {
--- a/packages/vm/core/evm/emulator/emulator.go
+++ b/packages/vm/core/evm/emulator/emulator.go
@@ -4,10 +4,12 @@
 package emulator
 
 import (
+	"bytes"
 	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/ethash"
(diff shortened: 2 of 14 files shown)


packages/isc/sandbox_interface.go
```diff
@@ -139,12 +138,12 @@ type Privileged interface {
...
 	CreditToAccount(AgentID, *Assets)
 	RetryUnprocessable(req Request, blockIndex uint32, outputIndex uint16)
-
-	// EVM
+	OnWriteReceipt(CoreCallbackFunc)
 	CallOnBehalfOf(caller AgentID, target, entryPoint Hname, params dict.Dict, allowance *Assets) dict.Dict
-	SetEVMFailed(*types.Transaction, *types.Receipt)
 }
 
+type CoreCallbackFunc func(partition kv.KVStore, chainInfo *ChainInfo)
```
jorgemmsilva: why do we pass `chainInfo` here?

packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -55,22 +55,17 @@ func createBlockchainDB(evmPartition kv.KVStore, chainInfo *isc.ChainInfo) *emul
...
 	chainInfo *isc.ChainInfo,
 	tx *types.Transaction,
 	receipt *types.Receipt,
 ) {
-	if tx == nil {
-		panic("nil tx")
-	}
-	if receipt == nil {
-		panic("nil receipt")
-	}
```
jorgemmsilva, line 70: can we keep these checks? they don't hurt, and they are a pain to track down because of the way our "panic catching" swallows stack traces :/

packages/vm/core/evm/emulator/emulator.go
```diff
@@ -330,9 +345,12 @@ func (e *EVMEmulator) SendTransaction(
 	}
 
 	buf.Commit()
```
jorgemmsilva, line 347: in the existing version `buf.Commit()` won't be called if `chargeISCGas` is called. I vaguely recall this being an important detail

jorgemmsilva (approved):
this logic is so tricky :confused: 

I like the "callback" logic, and the fact that we don't need to pass `chargeISCGas` into `SendTransaction` anymore,  but I'm afraid we could be introducing some sneaky bug with this "not strictly necessary" refactor.

packages/isc/sandbox_interface.go
dessaya: It's needed by `createBlockchainDB`, but I think I can get away without it. I'll remove it :+1:

packages/vm/core/evm/evmimpl/internal.go
dessaya: I'd rather fix that stack trace swallowing instead... is this an issue that affects solo or cluster tests?

packages/vm/core/evm/emulator/emulator.go
dessaya: Are you sure about that? In the existing version I don't see any `return` betweern `chargeISCGas()` and `buf.Commit()` :thinking:

packages/vm/core/evm/evmimpl/internal.go
jorgemmsilva: anything that runs the VM

packages/vm/core/evm/emulator/emulator.go
jorgemmsilva: `chargeISCGas` can panic if not enough gas, right?

dessaya:
Thanks for the review @jorgemmsilva !

> this logic is so tricky

Yeah... everything about the VM is still convoluted. My intention is to make it more maintainable and understandable. If this change does not improve that, then let's discard it :stuck_out_tongue: 

> I like the "callback" logic, and the fact that we don't need to pass chargeISCGas into SendTransaction anymore, but I'm afraid we could be introducing some sneaky bug with this "not strictly necessary" refactor.

I understand... but IMO it would also be very bad if we reached a point where we can't make changes in fear of breaking something not covered by tests. If that is the case, then let's test more.

packages/vm/core/evm/emulator/emulator.go
dessaya: I don't think it can panic because it's covered by `CatchPanic`.

packages/vm/core/evm/evmimpl/internal.go
dessaya: In solo: the "stack swallowing" happens by default but is easily disabled by adding `Debug: true, PrintStackTrace: true` in the solo initialization.

In cluster tests: I added a `reqctx.vm.task.Log.Debug(string(debug.Stack()))`, so now the stack trace is logged :)

packages/vm/core/evm/emulator/emulator.go
jorgemmsilva: Ah true, that's right 👍

PR #2734, opened 2023-07-20
Feat/single cache v2

shufps:
# Description of change

Adds a single cache instance with statistics ticker

### DIFF
--- a/packages/cache/cache.go
+++ b/packages/cache/cache.go
@@ -7,12 +7,16 @@ import (
 	"github.com/VictoriaMetrics/fastcache"
 )
 
-const partitionSize = 5
+const (
+	partitionSize = 5
+)
 
 type partition [partitionSize]byte
 
 type Stats struct {
 	*fastcache.Stats
+
+	// Number of handles
 	NumHandles uint64
 }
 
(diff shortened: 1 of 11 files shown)


jorgemmsilva (approved):

packages/cache/cache.go
```diff
@@ -0,0 +1,135 @@
...
+		return &CacheNoop{}, nil
+	}
+
+	// increment the handle counter
+	handleCounter++
+
+	// store counter into byte array
+	var partitionBytes partition
+	for i := 0; i < partitionSize; i++ {
+		partitionBytes[i] = byte(handleCounter >> (i * 8))
```
jorgemmsilva, line 115: I'm not good with these bitwise tricks :sweat_smile: .

a comment laying out the logic for this would be nice

jorgemmsilva (approved):

packages/cache/cache.go
Maenjul: In summary: The code fills the partitionBytes array with bytes by splitting the value of handleCounter into bytes and storing each byte at the corresponding position in the array. The right shift is used to isolate the bytes of handleCounter and convert them into individual byte values that are stored in the partitionBytes array.

PR #2735, opened 2023-07-21
Fix capitalized JSON fields in event payload

BugFreeSoftware:


packages/isc/event.go
```diff
@@ -50,18 +50,18 @@ func (e *Event) Write(w io.Writer) error {
 	return ww.Err
 }
 
+type EventJSON struct {
+	ContractID Hname  `json:"contractID" swagger:"desc(ID of the Contract that issued the event),required,min(1)"`
+	Topic      string `json:"topic" swagger:"desc(topic),required"`
+	Timestamp  uint64 `json:"timestamp" swagger:"desc(timestamp),required"`
```
muXxer, line 56: IMO you shouldn't use uint64 in json.
Maybe better convert it to string instead (thats what we do with uint64 values in the core API).

PR #2757, opened 2023-07-27
Fix utxos

dessaya:


packages/kv/collections/map.go
```diff
@@ -101,10 +101,14 @@ func (m *ImmutableMap) Len() uint32 {
 
 // Erase the map.
 func (m *Map) Erase() {
+	var keys [][]byte
 	m.IterateKeys(func(elemKey []byte) bool {
-		m.DelAt(elemKey)
+		keys = append(keys, elemKey)
 		return true
 	})
```
jorgemmsilva, line 108: could be extracted to a reusable `Keys()` func

jorgemmsilva (approved):
LGTM, but we should do this exact same change for the "unprocessable" requests and deprecate `(vmctx *vmContext) getOutputID` entirely.

https://github.com/iotaledger/wasp/blob/develop/packages/vm/core/blocklog/unprocessable.go#L18-L19

PR #2760, opened 2023-07-28
feat: Change gas for emit event as linear

### LINKED ISSUE #2742: Event publishing gas should be linear, not fixed
https://github.com/iotaledger/wasp/blob/develop/packages/vm/gas/table.go#L50

right now the cost for emiting events from the ISC sandbox is `constValue(10)`, this should instead scale with the size of the payload

howjmay:
closes #2742


packages/vm/gas/table.go
```diff
@@ -66,7 +66,7 @@ var burnTable = BurnTable{
 	BurnCodeUtilsBLSValidSignature:     {"bls valid", constValue(2000)},
 	BurnCodeUtilsBLSAddrFromPubKey:     {"bls addr", constValue(50)},
 	BurnCodeUtilsBLSAggregateBLS1P:     {"bls aggregate", linear(CoefBLSAggregate)},
-	BurnCodeMinimumGasPerRequest1P:     {"minimum gas per request", minBurn(10000)}, // TODO maybe make it configurable (gov contract?)
+	BurnCodeMinimumGasPerRequest1P:     {"minimum gas per request", minBurn(10000)}, // TODO maybe make it configurable (gov contract?) // FIXME not equal to wasmlib.MinGasFee
```
howjmay: This value is not relative to wasmlib.MinGasFee. Should it relate to wasmlib.MinGasFee?

jorgemmsilva: I guess wasmlib needs to be updated to whatever value we set here

howjmay: It was not the same too. But as you said, I think it is better to synchronize them

packages/chain/cons/cons_test.go
```diff
@@ -346,7 +347,8 @@ func testChained(t *testing.T, n, f, b int) {
 				inccounter.FuncIncCounter.Hname(),
 				dict.New(),
 				uint64(i*reqPerBlock+ii),
-				20000,
+				// FIXME may user other MinGasFee instead of wasmlib.MinGasFee
+				wasmlib.MinGasFee,
```
dessaya: At the very least, wasp core packages like this should not depend on wasmlib.

howjmay: done

packages/vm/gas/table.go
```diff
@@ -47,13 +47,13 @@ var burnTable = BurnTable{
...
 	BurnCodeCallContract:               {"call", constValue(100)},
-	BurnCodeEmitEventFixed:             {"event", constValue(10)},
+	BurnCodeEmitEvent1P:                {"event", linear(1)}, // 1 gas per byte
 	BurnCodeGetAllowance:               {"allowance", constValue(10)},
 	BurnCodeTransferAllowance:          {"transfer", constValue(10)},
 	BurnCodeEstimateStorageDepositCost: {"storage deposit estimate", constValue(5)},
 	BurnCodeSendL1Request:              {"send", linear(Coef1Send)},
 	BurnCodeDeployContract:             {"deploy", constValue(10)},
-	BurnCodeStorage1P:                  {"storage", linear(1)}, // 1 gas per byte
+	BurnCodeStorage1P:                  {"storage", linear(55)}, // 55 gas per byte
```
dessaya, line 56: Just curious, how did you arrive at this number?

dessaya (approved):

packages/vm/gas/table.go
howjmay: It is about the ratio of gas per byte of ETH storage and event

PR #2766, opened 2023-08-01
fix: unprocessable requests with blocklog pruning

dessaya:


packages/vm/core/blocklog/interface.go
```diff
@@ -9,11 +9,16 @@ import (
...
-	prefixRequestReceipts
-	prefixRequestEvents
-	prefixUnprocessableRequests
+	PrefixBlockRegistry      = "a"
+	prefixRequestLookupIndex = "b"
+	prefixRequestReceipts    = "c"
+	prefixRequestEvents      = "d"
+
+	// map of == request ID => unprocessableRequestRecord
+	prefixUnprocessableRequests = "u"
```
dessaya, line 18: Changed this intentionally to avoid doing a migration. The map is likely empty in the testnet anyway.

jorgemmsilva (approved):

Issue #2401, opened 2023-04-26
Wal: split blocks in subfolders by first two digits of block's hash


Juliusan:

Juliusan:
I've analized a WAL of 231187 blocks. Splitting it to subfolders by first two digits of the hash would result in full amount of subfolders: 256. Splitting by first three digits would also result in full amount of subfolders: 4096. And only splitting by first four digits would result in slightly lower subfolder count (63603), then the amount of four digit hexadecimal numbers (65536). Maybe splitting by first four digits should be implemented.

Issue #2400, opened 2023-04-26
Wal: write block to temporary file and move the file only, when it is done.


Juliusan:

shufps:
I like this ... moving a file is an atomic operation and there is no way to have corrupted/incomplete wal files in case  of a crash (except the tmp-file of course, but it's not used then)

PR #2771, opened 2023-08-03
WAL improvements

Juliusan:
Two WAL improvements:
* Block is written to temporary file and then moved to permanent log location (https://github.com/iotaledger/wasp/issues/2400). So far no deletion of unneded temporary files. It shouldn't be such a huge problem.
* Blocks in WAL folder are split into subfolders according to first two digits of block hash (https://github.com/iotaledger/wasp/issues/2401). For backwards compatibility reasons, blocks are read from WAL root folder too. It also makes it easier to copy blocks between nodes as it is not necessary to create subfolders for them. Finally, if block loading from WAL on node start is enabled, WAL directory is read recursively: blocks in root folder, it its subfolders, in their subfolders, etc... are loaded.


packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
```diff
@@ -45,37 +49,77 @@ func NewBlockWAL(log *logger.Logger, baseDir string, chainID isc.ChainID, metric
...
+	subfolderName := blockWALSubFolderName(commitment.BlockHash())
+	folderPath := filepath.Join(bwT.dir, subfolderName)
+	if err := ioutils.CreateDirectory(folderPath, 0o777); err != nil {
+		return fmt.Errorf("failed create folder %v for writing block index %v: %w", folderPath, blockIndex, err)
 	}
-	defer f.Close()
-	blockBytes := block.Bytes()
-	n, err := f.Write(blockBytes)
+	tmpFileName := blockWALTmpFileName(commitment.BlockHash())
+	tmpFilePath := filepath.Join(folderPath, tmpFileName)
```
jorgemmsilva, line 58: should this use `os.TempDir()` instead?

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
```diff
@@ -45,37 +49,77 @@ func NewBlockWAL(log *logger.Logger, baseDir string, chainID isc.ChainID, metric
...
+			tmpFilePath, finalFilePath, err)
 	}
+
 	bwT.metrics.BlockWritten(block.StateIndex())
-	bwT.LogDebugf("Block index %v %s written to wal; file name - %s", blockIndex, commitment, fileName)
+	bwT.LogDebugf("Block index %v %s written to wal; file name - %s", blockIndex, commitment, finalFileName)
 	return nil
 }
 
+func (bwT *blockWAL) containsWithPath(blockHash state.BlockHash) (bool, string) {
```
jorgemmsilva: could this be renamed to `has` or `blockFilepath` or something like that? `containsWithPath` is kinda of a weird name

jorgemmsilva: also the order of the returns should be switched to be "more standard"
example:
```
x , exists := something.Get(key)
```

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
```diff
@@ -45,37 +49,77 @@ func NewBlockWAL(log *logger.Logger, baseDir string, chainID isc.ChainID, metric
...
-	_, err := os.Stat(filepath.Join(bwT.dir, blockWALFileName(blockHash)))
-	return err == nil
+	result, _ := bwT.containsWithPath(blockHash)
+	return result
 }
 
 func (bwT *blockWAL) Read(blockHash state.BlockHash) (state.Block, error) {
-	fileName := blockWALFileName(blockHash)
-	filePath := filepath.Join(bwT.dir, fileName)
+	conains, filePath := bwT.containsWithPath(blockHash)
```
jorgemmsilva: typo

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
```diff
@@ -116,6 +152,28 @@ func (bwT *blockWAL) ReadAllByStateIndex(cb func(stateIndex uint32, block state.
 		}
 		blocksByStateIndex[stateIndex] = stateIndexPaths
 	}
+
+	var checkDir func(dirPath string, dirEntries []os.DirEntry)
+	checkDir = func(dirPath string, dirEntries []os.DirEntry) {
```
jorgemmsilva, line 157:
```suggestion
	checkDir := func(dirPath string, dirEntries []os.DirEntry) {
```

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
```diff
@@ -116,6 +152,28 @@ func (bwT *blockWAL) ReadAllByStateIndex(cb func(stateIndex uint32, block state.
...
+				checkFile(entryPath)
+			}
+		}
+	}
+
+	dirEntries, err := os.ReadDir(bwT.dir)
+	if err != nil {
+		return err
+	}
+	checkDir(bwT.dir, dirEntries)
```
jorgemmsilva, line 175: what is the point of this? to log in case some file is unreadable?

jorgemmsilva (approved):
looks ok overall, a few comments.

btw, I thought we wanted to include the block index in the file name?

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
Juliusan: Cannot be done due to recursive call in function definition.

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
Juliusan: What is the point of what? If WAL directory cannot be listed, error should be thrown. Otherwise, WAL files can be directly in WAL folder (this was the old behaviour and it might be convenient for administrator to add file to a parent folder) or in subfolders according to first two digits of hash. It was easier to do complete recursion that stop at level 2 of folder scanning. I saw no harm in that. However if subfolders cannot be listed, they should simply be ignored and not generate errors.

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
Juliusan: IDK. It seemed simpler this way. I am not sure if there are not going to be name collisions otherwise. If, e.g., two nodes are running on the same OS. This is real at least in test scenarios.

packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_wal.go
jorgemmsilva: I see, the purpose is to fill `blocksByStateIndex`. I would probably write it as:
```
blocksByStateIndex := checkDir(bwT.dir, dirEntries)
```
but whatever works :shrug:

PR #2775, opened 2023-08-04
feat: mempool nonce override

jorgemmsilva:
allows cancellation of requests by sending a new request with the same nonce

refactors the way mempool does nonce selection, now we keep an ordered list of requests per account, which facilitates the selection process.

adds new logs to allow easier debugging per account

### DIFF
--- a/packages/chain/mempool/mempool.go
+++ b/packages/chain/mempool/mempool.go
@@ -558,6 +558,11 @@ func (mpi *mempoolImpl) refsToPropose() []*isc.RequestRef {
 				mpi.log.Debugf("refsToPropose, account: %s, removing old nonce from pool: %d", account, e.req.Nonce())
 				mpi.offLedgerPool.Remove(e.req)
 			}
+			if e.old {
+				// this request was marked as "old", do not propose it
+				mpi.log.Debugf("refsToPropose, account: %s, skipping old request: %s", account, e.req.ID().String())
+				continue
+			}
 			if reqNonce == accountNonce {
 				// expected nonce, add it to the list to propose
 				mpi.log.Debugf("refsToPropose, account: %s, proposing reqID %s with nonce %d: d", account, e.req.ID().String(), e.req.Nonce())
--- a/packages/chain/mempool/mempool.go
+++ b/packages/chain/mempool/mempool.go
@@ -565,7 +565,7 @@ func (mpi *mempoolImpl) refsToPropose() []*isc.RequestRef {
 				accountNonce++ // increment the account nonce to match the next valid request
 			}
 			if reqNonce > accountNonce {
-				mpi.log.Debugf("refsToPropose, account: %s, req %s has a nouce %d which is too high, won't be proposed", account, e.req.ID().String(), e.req.Nonce())
+				mpi.log.Debugf("refsToPropose, account: %s, req %s has a nonce %d which is too high, won't be proposed", account, e.req.ID().String(), e.req.Nonce())
 				return // no more valid nonces for this account, continue to the next account
 			}
 		}
--- a/packages/chain/mempool/mempool.go
+++ b/packages/chain/mempool/mempool.go
@@ -48,7 +48,6 @@ import (
 	"time"
 
 	"github.com/samber/lo"
-	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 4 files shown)


packages/chain/mempool/mempool.go
```diff
@@ -551,53 +545,31 @@ func (mpi *mempoolImpl) refsToPropose() []*isc.RequestRef {
...
+				reqRefs = append(reqRefs, isc.RequestRefFromRequest(e.req))
+				accountNonce++ // increment the account nonce to match the next valid request
 			}
-			// check for gaps within the request list
-			for i := 1; i < len(refNonces); i++ {
-				if refNonces[i].nonce != refNonces[i-1].nonce+1 {
-					doNotPropose = append(doNotPropose, refNonces[i].ref)
-				}
+			if reqNonce > accountNonce {
+				mpi.log.Debugf("refsToPropose, account: %s, req %s has a nouce %d which is too high, won't be proposed", account, e.req.ID().String(), e.req.Nonce())
```
BugFreeSoftware: Nouce instead of nonce

packages/chain/mempool/typed_pool_by_nonce.go
```diff
@@ -0,0 +1,173 @@
...
+	)
+	if exists {
+		// same nonce, delete old request with overlapping nonce, replace with the new one
+		p.Remove(reqsForAcount[index].req)
+		// refresh `reqsForAcount` after removing the old req
+		reqsForAcount, exists = p.reqsByAcountOrdered.Get(account)
+		if !exists {
+			reqsForAcount = make([]*OrderedPoolEntry[V], 0)
+		}
+		// TODO could	this create some race condition if there is ongoing voting on a batch propostal containing the old nonce?
```
BugFreeSoftware: Are all nodes eventually supposed to end up with the same requests in the same order?
Or is it possible for two requests with the same nonce to be sent out quickly and end up in different order at different nodes? This should definitely be deterministic.
Of course, this means that that particular account is shooting itself in the foot in that case, and it can always send one with the same nonce again to all nodes to get unstuck.

BugFreeSoftware (approved):

packages/chain/mempool/typed_pool_by_nonce.go
jorgemmsilva: Batch selection will always be deterministic and there will never 2 requests with the same nonce on the same batch.

The only potential problem I see is that we could remove a request that will be needed for a pending batch, and the node will need to fetch that request from other nodes while at the same time broadcasting the new request and trying to make other nodes overwrite the old (selected) request.

Unsure what is the best way to prevent this. I'll wait for @kape1395 input

packages/chain/mempool/typed_pool_by_nonce.go
```diff
@@ -0,0 +1,173 @@
...
+	if len(reqsByAccount) == 1 { // just remove the entire array for the account
+		p.reqsByAcountOrdered.Delete(account)
+		return
+	}
+	reqsByAccount[indexToDel] = nil // remove the pointer reference to allow GC of the entry object
+	reqsByAccount = slices.Delete(reqsByAccount, indexToDel, indexToDel+1)
+	p.reqsByAcountOrdered.Set(account, reqsByAccount)
+}
+
+func (p *TypedPoolByNonce[V]) Iterate(predicate func(account string, requests []*OrderedPoolEntry[V])) {
```
dessaya: Nit: technically it's not a predicate since it does not return bool

dessaya (approved):

packages/chain/mempool/typed_pool_by_nonce.go
kape1395: Good point here. Indeed there was an assumption that if F+1 node has proposed the request, it will be available at least on one correct node. So consensus nodes will wait for that request from other nodes if they have no such in their mempool. If a request is dropped by overriding, the consensus will be stuck waiting for such a request.

One way to resolve that is to mark a request as non-proposable instead of removing it. This way, a node could share the request if needed. This is only needed for the period of time when the request was proposed, but the consensus still needs to be finished.

kape1395 (approved):

PR #2784, opened 2023-08-08
Loading chain state from WAL optimised

Juliusan:
When writing block to file in WAL, at first block index is written. This allows to optimise algorithm, which loads chain state at the start of the node from WAL. Before the change, to sort WAL blocks by index the whole block had to be read. Now only first 4 bytes (`uint32`) are read. Feature requested in discussion: https://iotafoundation.slack.com/archives/G01F91K2CRX/p1691054725116629.

NOTE: after this PR is merged, WAL (and separate WAL files) from earlier releases will be incompatible.

### DIFF
--- a/packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_cache.go
+++ b/packages/chain/statemanager/sm_gpa/sm_gpa_utils/block_cache.go
@@ -89,7 +89,7 @@ func (bcT *blockCache) GetBlock(commitment *state.L1Commitment) state.Block {
 	if bcT.wal.Contains(commitment.BlockHash()) {
 		block, err := bcT.wal.Read(commitment.BlockHash())
 		if err != nil {
-			bcT.log.Errorf("Error reading block %s from WAL: %w", commitment, err)
+			bcT.log.Errorf("Error reading block index %v %s from WAL: %w", block.StateIndex(), commitment, err)
 			return nil
 		}
 		bcT.addBlockToCache(block)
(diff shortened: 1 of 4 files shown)


dessaya (approved):

shufps:
regarding `NOTE: after this PR is merged, WAL (and separate WAL files) from earlier releases will be incompatible.`

Do we have some tool to convert the WAL files into the new format?

And adding a version byte as the first byte in the file is generally a good idea.

This way backwards compatibility might be retained if there are other changes in future.

jorgemmsilva (approved):
Agree with @shufps , adding a version byte might be a good idea so its easier to make changes to the format in the future

Juliusan:
Added backward compatibility and tests for that. As old WAL format didn't have version byte, there is a tiny possibility (1/256^4, if we assert that trie root hash is completely random) of old WAL format file being identified as new version 1 file. To lower that possibility I used 4 bytes to store version. This won't affect blocks being loaded from WAL, because loading old format file as version 1 format file should fail and wasp will try to loaded as old format file. However unexpected error might occur when restoring state from WAL on node startup. If that happens, sorry to say, but the file will have to be edited by hand. I think, it is very unlikely so I haven't written a tool to do that. But if that appears to be a problem, I will develop a tool.

shufps:
Thx a lot for backwards compatibilty :raised_hands: Fallback to read the file as v1 for the unlikely event that the random 4 first bytes of a file are read as v2 header with version 1, is good I think.

But a conversion tool also would have been okay (even if crammed into a fork directly into the from-WAL-syncing logic) and just add the version for future changes because we are probably the only one with v0 files :blush:

PR #2787, opened 2023-08-08
state: prevent parallel writes to the state DB

dessaya:


packages/state/db.go
```diff
@@ -179,7 +187,5 @@ func (db *storeDB) restoreSnapshot(root trie.Hash, r io.Reader) error {
 	if err != nil {
 		return err
 	}
-
-	db.setLatestTrieRoot(root)
```
dessaya, line 183: @Juliusan previously the "latest trie root" was automatically set after restoring the snapshot. I figured we don't want that if we are restoring snapshots for other reasons.

luke-thorne (approved):

PR #2790, opened 2023-08-09
refactor: Concise writeReceiptToBlockLog log message

howjmay:
Remove the `err: ` part in log message when there is no error returned


jorgemmsilva (changes requested):
more lines of code for the same effect :-1: 

I would not make this change

PR #2792, opened 2023-08-09
CmtLog redesigned.

kape1395:


packages/chain/cmt_log/WaspChainRecovery_V2 L1.tla
```diff
@@ -0,0 +1,225 @@
+-------------------------- MODULE WaspChainRecovery_V2 -------------------------
+(*
+This specification is concerned with a Chain's committee recovery in the case,
+when we have more than F nodes failed by crashing and probably loosing the
```
jorgemmsilva:
```suggestion
when we have more than F nodes failed by crashing and probably losing the
```

jorgemmsilva (approved):
there are some TODOs in the .tla files. I'm not knowledgeable to review those.

`var_log_index.go` has a lot of "development notes" and cases to be handled (some of those appear to have been solved already), we should probably clean that up in later changes.

the go code LGTM :+1:

kape1395:
> there are some TODOs in the .tla files. I'm not knowledgeable to review those.
> 
> `var_log_index.go` has a lot of "development notes" and cases to be handled (some of those appear to have been solved already), we should probably clean that up in later changes.
> 
> the go code LGTM 👍

I will work on the documentation for this stuff next, then I will be able to clean up the TLA files as well as the comments.

PR #2798, opened 2023-08-12
feat: Allow free request by GasPerToken is 0:0

### LINKED ISSUE #2119: Allow "0 fee" execution
Right now setting the gasPerToken to 0:0 will mean nothing can be payed for. 
We can add an edge case to make running requests "free" (0 fees).

Should be documented that its not something we recommend and doesn't make sense on public networks. However it might be potentially useful for "private" chains where access is restricted.

howjmay:
closes https://github.com/iotaledger/wasp/issues/2119

### DIFF
--- a/packages/util/ratio.go
+++ b/packages/util/ratio.go
@@ -103,11 +103,21 @@ func (ratio Ratio32) HasZeroComponent() bool {
 	return ratio.A == 0 || ratio.B == 0
 }
 
+var ZeroGasFee = Ratio32{}
+
+func (ratio Ratio32) IsRatioValid() bool {
+	return ratio == ZeroGasFee || !ratio.HasZeroComponent()
+}
+
+func (ratio Ratio32) IsGasFeeCharged() bool {
+	return ratio != ZeroGasFee
+}
+
 func (ratio *Ratio32) Read(r io.Reader) error {
 	rr := rwutil.NewReader(r)
 	ratio.A = rr.ReadUint32()
 	ratio.B = rr.ReadUint32()
-	if rr.Err == nil && ratio.HasZeroComponent() {
+	if rr.Err == nil && ratio.IsRatioValid() {
 		rr.Err = errors.New("ratio has zero component")
 	}
 	return rr.Err
--- a/packages/util/ratio.go
+++ b/packages/util/ratio.go
@@ -103,21 +103,20 @@ func (ratio Ratio32) HasZeroComponent() bool {
 	return ratio.A == 0 || ratio.B == 0
 }
 
-var ZeroGasFee = Ratio32{}
-
-func (ratio Ratio32) IsRatioValid() bool {
-	return ratio == ZeroGasFee || !ratio.HasZeroComponent()
+func (ratio Ratio32) IsValid() bool {
+	return ratio.IsZero() || !ratio.HasZeroComponent()
 }
 
-func (ratio Ratio32) IsGasFeeCharged() bool {
-	return ratio != ZeroGasFee
+func (ratio Ratio32) IsZero() bool {
+	ZeroGasFee := Ratio32{}
+	return ratio == ZeroGasFee
 }
 
 func (ratio *Ratio32) Read(r io.Reader) error {
 	rr := rwutil.NewReader(r)
 	ratio.A = rr.ReadUint32()
 	ratio.B = rr.ReadUint32()
-	if rr.Err == nil && ratio.IsRatioValid() {
+	if rr.Err == nil && !ratio.IsValid() {
 		rr.Err = errors.New("ratio has zero component")
 	}
 	return rr.Err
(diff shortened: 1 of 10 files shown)


packages/util/ratio.go
```diff
@@ -99,15 +99,15 @@ func (ratio Ratio32) Type() string {
 	return "Ratio32"
 }
 
-func (ratio Ratio32) HasZeroComponent() bool {
-	return ratio.A == 0 || ratio.B == 0
+func (ratio Ratio32) HasInvalidZeroComponent() bool {
+	return (ratio.A == 0 || ratio.B == 0) && !(ratio.A == 0 && ratio.B == 0)
 }
```
dessaya: This funtion IMO now looks too specific for the gas ratio. The `Ratio32` type is meant to represent any ratio, not just for gas fee (although right now the only use is for the gas fee).

I sugest moving this logic to the `gas` package:

```
package gas

var ZeroGasFee = Ratio32{}

func IsRatioValid(r Ratio32) bool {
    return r == ZeroGasFee || !r.HasZeroComponent()
}

func IsGasFeeCharged(r Ratio32) bool {
    return r != ZeroGasFee
}
```

packages/vm/vmimpl/runreq.go
```diff
@@ -140,6 +140,17 @@ func (reqctx *requestContext) checkAllowance() {
 }
 
 func (reqctx *requestContext) shouldChargeGasFee() bool {
+	// freeGasPerToken checks whether we charge token per gas
+	// If it is free, then we will still burn the gas, but it doesn't charge tokens
+	// NOT FOR PUBLIC NETWORK
```
dessaya, line 145: End users will not see this warning. Also there is no way for them to guess that they have to set 0:0 to disable the gas fee. IMO there should be a dedicated command in `wasp-cli` (like `wasp-cli chain disable-gas-policy` or whatever), or at least some example in the wiki.

packages/util/ratio.go
howjmay: I have use `IsRatioValid() ` to replace the current usage of `HasZeroComponent() `. Is it ok?

packages/vm/vmimpl/runreq.go
howjmay: I have added the command and relating test

packages/util/ratio.go
```diff
@@ -103,11 +103,21 @@ func (ratio Ratio32) HasZeroComponent() bool {
...
+var ZeroGasFee = Ratio32{}
+
+func (ratio Ratio32) IsRatioValid() bool {
+	return ratio == ZeroGasFee || !ratio.HasZeroComponent()
+}
+
+func (ratio Ratio32) IsGasFeeCharged() bool {
+	return ratio != ZeroGasFee
+}
+
```
dessaya: Hey, that's better but still has a "separation of concerns" problem: `Ratio32` is a generic math type. It can represent a ratio between any two integer values, for example bytes/network packets, cars/people, hamburgers/McDonald's stores...

So it should not know anything about gas, or anything ISC related. IMO these functions should go in the `gas` package.

dessaya (approved):

packages/util/ratio.go
howjmay: It is good idea.

packages/util/ratio.go
```diff
@@ -107,7 +107,7 @@ func (ratio *Ratio32) Read(r io.Reader) error {
 	rr := rwutil.NewReader(r)
 	ratio.A = rr.ReadUint32()
 	ratio.B = rr.ReadUint32()
-	if rr.Err == nil && ratio.HasZeroComponent() {
+	if rr.Err == nil && !(ratio == &Ratio32{} || !ratio.HasZeroComponent()) {
```
howjmay: @dessaya here it has a cyclic importing problem. And we need to support `0:0` here too. Therefore, I have to do in this way. Is it ok?

dessaya: Well... I don't know. Again, this validation is specific to gas. From the point of view of `Ratio32`, 0:0 is invalid (since mathematically 0/0 is undefined). And both 0:x or x:0 are "dangerous" because they might panic withzero division error.

But OK, maybe the simplest solution is to add the concept of a "Zero" Ratio32, so the invariant of Ratio32 can be:

> Either A != 0 and B != 0, or both A == 0 and B == 0.

Then:

```
func (ratio Ratio32) IsZero() bool {
	return ratio.A == 0 && ratio.B == 0
}

func (ratio Ratio32) IsValid() bool {
	return ratio.IsZero() || !ratio.HasZeroComponent()
}
```

Then:

```suggestion
	if rr.Err == nil && ratio.IsValid() {
```

Or just ignore all my comments and do what you thik is best. This is just me nitpicking... :stuck_out_tongue:

howjmay: Let me do the above way. That looks fine for me

PR #2801, opened 2023-08-15
chore: Synchronize contract interface naming

howjmay:


howjmay:
What is the rule of the naming of the keys of state variables, parameters, prefixes? Maybe we need one?

jorgemmsilva (changes requested):
>What is the rule of the naming of the keys of state variables, parameters, prefixes? Maybe we need one?

we don't have anything strict, basically "don't repeat vars" and "single char as most as possible to save space". To be honest I don't like how we do it tho, using an enum or similar would probably be nicer so we cannot have human error creating collision between keys in the same state. But it is what it is, it works in the current form and we should be changing this much in the future anyway

---

This is a positive change, its good to keep the code with some organization/structure.
however, I've noticed these changes result in the `testdbhash` being updated, which it shouldn't IMO

howjmay:
> > What is the rule of the naming of the keys of state variables, parameters, prefixes? Maybe we need one?
> 
> we don't have anything strict, basically "don't repeat vars" and "single char as most as possible to save space". To be honest I don't like how we do it tho, using an enum or similar would probably be nicer so we cannot have human error creating collision between keys in the same state. But it is what it is, it works in the current form and we should be changing this much in the future anyway
> 
> This is a positive change, its good to keep the code with some organization/structure. however, I've noticed these changes result in the `testdbhash` being updated, which it shouldn't IMO



@jorgemmsilva 
But I changed the naming of variables in this PR, which would cause the binary is not the same. I think the db hash should changed, since we save the binary to kvstore. 
Not sure whether I made some mistake for the understanding

jorgemmsilva:
@howjmay 
> But I changed the naming of variables in this PR, which would cause the binary is not the same. I think the db hash should changed, since we save the binary to kvstore.

exactly. But that should not happen (we shouldn't change the keys as it will cause issues on running networks)

howjmay:
@jorgemmsilva
So change like this I should discard in this PR?
https://github.com/iotaledger/wasp/pull/2801/files#diff-d006fa22dfaf47a4bd1d3856d50e1b301921108506448973993232694853c3c1R62

I think this place could be one of the reason that I need to update the hash

jorgemmsilva (approved):

PR #2810, opened 2023-08-21
fix: Conduct proper aud, iss, and sub in jwt

howjmay:
The issuer has been changed to the public key of the nodeID

### DIFF
--- a/packages/authentication/jwt_auth.go
+++ b/packages/authentication/jwt_auth.go
@@ -10,6 +10,7 @@ import (
 	"github.com/labstack/echo/v4"
 
 	"github.com/iotaledger/wasp/packages/authentication/shared/permissions"
+	"github.com/iotaledger/wasp/packages/cryptolib"
 )
 
 // Errors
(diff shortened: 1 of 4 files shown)


lmoe (approved):
Looks good. 
I hope we have at some point a proper auth provider than can span over multiple nodes.

howjmay:
> Looks good. I hope we have at some point a proper auth provider than can span over multiple nodes.

@lmoe I don't understand this. Would you mind providing more information then I can add them as TODO in codebase

lmoe:
@howjmay Don't worry about it for now, it should when rather be a proper task.

The problem is that each node has a different private key and all JWT tokens are signed with a private key. 

When you put multiple Wasp nodes behind a load balancer, the JWT token validation will fail on all nodes except the one which has signed it.

One option to solve this, is a separate auth service node that signs/validates tokens and handles the logins. 
All Wasp nodes would connect to this service and ask for token validation when a request comes in.

So we would basically need to move all auth related parts to a separate node, write an API around it and deploy it with the Wasp nodes - or even better, use a full fledged oidc service that most likely exists already anyway. 

But I would not spend time on it right now. :)

howjmay:
> 

wow I won't either Then I am going to merge this one

PR #2813, opened 2023-08-22
chore: Remove unnecessary todo

howjmay:
The feature that the TODO mentioned will make the codebase too complex and not a really high demanded requirement

### DIFF
--- a/packages/vm/gas/table.go
+++ b/packages/vm/gas/table.go
@@ -66,7 +66,7 @@ var burnTable = BurnTable{
 	BurnCodeUtilsBLSValidSignature:     {"bls valid", constValue(2000)},
 	BurnCodeUtilsBLSAddrFromPubKey:     {"bls addr", constValue(50)},
 	BurnCodeUtilsBLSAggregateBLS1P:     {"bls aggregate", linear(CoefBLSAggregate)},
-	BurnCodeMinimumGasPerRequest1P:     {"minimum gas per request", minBurn(10000)}, // TODO maybe make it configurable (gov contract?)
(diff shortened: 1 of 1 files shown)


howjmay:
Maybe we should make this into part of `setFeePolicy`?

packages/vm/core/governance/governanceimpl/gas.go
```diff
@@ -19,10 +19,11 @@ func setFeePolicy(ctx isc.Sandbox) dict.Dict {
 	ctx.RequireCallerIsChainOwner()
 
 	data := ctx.Params().MustGetBytes(governance.ParamFeePolicyBytes)
-	_, err := gas.FeePolicyFromBytes(data)
+	fp, err := gas.FeePolicyFromBytes(data)
 	ctx.RequireNoError(err)
 
 	ctx.State().Set(governance.VarGasFeePolicyBytes, data)
+	gas.MinimumGasPerRequest = fp.MinGasPerRequest
```
dessaya: :thinking: This is a global variable? What if there are multiple chains with different configurations?

howjmay: oh shit.... 
At this moment all the chains share the same gas table. 
I think if we want the configarable feature, it means we need to have gas table for each chain respectively.

howjmay: I will save the gas table in the `ChainInfo` struct

howjmay: I think this will bring another big feature that we support configurable gas for each instructions for different chains.

cwarnerdev (approved):

PR #2823, opened 2023-08-24
add additional random helpers

### LINKED ISSUE #2747: Solidity helper function for randomness
When using our magic smart contract entropy functionality you get a single deterministic random hash per block (only available after consensus has been reached, so unpredictable beforehand). If you wish to use this entropy for multiple random numbers you need to generate these random numbers based on this entropy (similar to https://www.geeksforgeeks.org/random-number-generator-in-solidity-using-keccak256/ but using the entropy from the magic contract instead of some other entropy source). It would be nice if we can offer some helper functionality here for some common use-cases to more easily generate random numbers/hashes without having to implement the functionality for converting entropy  + index to an actual list of random numbers/hashes. 
```[tasklist]
### Tasks
- [x] Helper function for generating a random uint256
- [x] Helper function for generating a random number within a certain range (min, max)
- [x] Helper function for generating a random hash (bytes32)
- [x] Documentation / examples in Wiki
```

luke-thorne:
# Description of change

Devs appreciate more ways to get random values, this adds a few more. 

Resolves #2747

## Type of change

### DIFF
--- a/tools/evm/iscutils/README.md
+++ b/tools/evm/iscutils/README.md
@@ -36,15 +36,19 @@ contract MyEVMContract {
     using PRNG for PRNG.PRNGState;
 
     event PseudoRNG(uint256 value);
+    event PseudoRNGHash(bytes32 value);
     
     PRNG.PRNGState private prngState;
 
     function emitValue() public {
         bytes32 e = ISC.sandbox.getEntropy();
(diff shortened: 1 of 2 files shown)


cwarnerdev (approved):
These are great additions. Nice find on that open issue as well!

PR #2837, opened 2023-08-30
Snapshot refactoring

Juliusan:
Refactored snapshots loading to avoid loading them on non empty store. On node start, if store is empty, newest snapshot is found and loaded. If this is not acceptable, node can be configured to load specific snapshot for each chain on node start. This simplified state manager as now it doesn't look for a snapshot to load. However, it added some constrains mainly to limit the depth of search for blocks. If a block, which is older than current snapshot, or which is older than last pruned block, is needed by state manager, the node simply panics. To lessen the probability of such situations, snapshot production is delayed: you can configure, how much newer blocks must be committed before snapshot is produced. I also updated the documentation, which is not visible in this PR.


packages/isc/chainid.go
```diff
@@ -19,7 +19,10 @@ var emptyChainID = ChainID{}
 
 // ChainID represents the global identifier of the chain
 // It is wrapped AliasAddress, an address without a private key behind
-type ChainID iotago.AliasID
+type (
+	ChainID    iotago.AliasID
+	ChainIDKey string
```
dessaya, line 24: Is this really necessary? The `AliasID` type is already an array, so it can be used as a map key. Can't we just remove the `Key()` method?

dessaya (approved):

packages/isc/chainid.go
Juliusan: Didn't think of that. @jorgemmsilva  added `Key` method in ChainID with commit `5b2a16c98cea1e051d0da9122be6c59491192503`. I am not sure why. I just wanted the key to have its own separate type. I could either remove `Key` method (and type) altogether or change `ChainIDKey` type to be alias of `ChainID` if it is any possibility that iota might change `AliasID` type to be of something that cannot be a key.

dessaya: My vote would be for removing it, but either option is fine.

jorgemmsilva: I agree, :axe: it

PR #2857, opened 2023-09-07
fix(evm): properly revert emitted logs after revert

### LINKED ISSUE #2847: Events emitted but evm state not updated.
**Describe the bug**
On rare occasions, events are emitted from a function call, but the state changes of the function did not happen. This means that the events are reporting a wrong evm state and any systems (such as subgraph) relying on these events are not representing the actual evm state. 

**To Reproduce**
This is not easily reproducible, I don't know what has to happen for this issue to occur. I will explain how I observe the bug though.

This is a function in our (IOTA Heroes) adventure contract:
image
I know quite a few things are happening in the function. I don't know if calling some external contracts is required to reproduce this error. 
"tokens" is an ERC1155 contract, its being used to send the IOTA Heroes game items to the contract while the hero is on an adventure.
adventureCounter is a simple Counter from Openzeppelin to keep track of the next adventureId.

Notice the adventureCounter being incremented in the bottom of the function as well as the Event (AdventureInitiated) being emitted in the same contract.

The bug is observed when we see multiple AdventureInitiated events (occuring over 100 blocks) having the same adventureId. (meaning the adventureCounter was not updated).

Here is a query from our subgraph showing a bunch of the buggy events:
image
As you can see the adventureId is 4033 in all of these events, you can also see the blockNumber increasing. This should not be possible.

To eliminate the issue having anything to do with subgraph here is a screenshot from a script inspecting two of the events in question directly from the rpc:
image

**Expected behavior**
When calling any function I'd expect that all state changes and events happen as expected, or that everything is correctly reverted (and no event emitted).

**Screenshots**
See above

**Network and versioning**
ShimmerEVM testnet

**Ideas **
I don't know what the issue is here, it happens in multiple other contracts, this is just an example that is used a lot in the game, so there are many examples from this one. In other examples we observed that some of the state changes did indeed happen, but not all of them. For example we have a contract where the user can purchase a new hero (erc721) using a mix of erc20 tokens and erc1155 tokens. In some cases only the erc1155 tokens was substracted but not the erc20 tokens. This is just some additional anecdotal info that might help pinpoint the issue.

dessaya:
Also fixes calculation of logIndex value in jsonrpc.

closes #2847


jorgemmsilva (approved):
nice!

Crelde:
Yeeees! You go guys!

PR #2863, opened 2023-09-11
fix: Kill inx and hornet after cluster tests on macOS

howjmay:


tools/cluster/tests/cluster.go
```diff
@@ -64,7 +64,14 @@ func newCluster(t *testing.T, opt ...waspClusterOpts) *cluster.Cluster {
...
+		stopCluster(clu)
+	})
 
 	return clu
 }
+
+func stopCluster(clu *cluster.Cluster) {
+	clu.Stop()
+	l1.Stop()
+}
```
jorgemmsilva, line 77: this is a bad idea, the L1 is kept alive though a "cluster test" run. (many test cases are executed on the same network)

.github/workflows/heavy-tests.yml
```diff
@@ -247,6 +249,7 @@ jobs:
     name: Test
     runs-on:
       - self-hosted
+      - macos-latest
```
jorgemmsilva, line 252: what's the point of adding this? this yml file is only for CI

tools/cluster/tests/cluster.go
howjmay: umm or I should add the stop function in the end of all tests?

.github/workflows/heavy-tests.yml
howjmay: This is for the test. I want to enable cluster test in this PR to check whether anything wrong here. I will remove this once the PR is ready

tools/cluster/tests/cluster.go
jorgemmsilva: lol no. please re-read my comment.

PR #2870, opened 2023-09-13
feat: allow multi-chain interactions with EVM AgentID (BREAKING)

jorgemmsilva:
Changes the on-chain request structure to allow to specify EVM addresses as the "sender contract"


packages/isc/agentid_eth.go
```diff
@@ -12,24 +13,31 @@ import (
...
-	eth common.Address
+	chainID ChainID
+	eth     common.Address
 }
 
 var _ AgentID = &EthereumAddressAgentID{}
 
-func NewEthereumAddressAgentID(eth common.Address) *EthereumAddressAgentID {
-	return &EthereumAddressAgentID{eth: eth}
+func NewEthereumAddressAgentID(eth common.Address, chainID ChainID) *EthereumAddressAgentID {
```
dessaya: Nit: to make it consistent with `NewContractAgentID`:

```suggestion
func NewEthereumAddressAgentID(chainID ChainID, eth common.Address) *EthereumAddressAgentID {
```

packages/isc/contract_identity.go
```diff
@@ -0,0 +1,99 @@
...
+	"io"
+
+	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/iotaledger/wasp/packages/util/rwutil"
+)
+
+type ContractIdentity struct {
+	// can either be an Hname or a solidity contract
+	kind rwutil.Kind
```
dessaya: Could it be something like the following, to make it more explicit / type-safe?

```suggestion
	kind ContractIdentityKind
```

where:

```
type ContractIdentityKind byte // or rwutil.Kind

const (
	contractIdentityKindEmpty ContractIdentityKind = iota
	contractIdentityKindHname
	contractIdentityKindEthereum
)
```

packages/isc/contract_identity.go
```diff
@@ -0,0 +1,99 @@
...
+
+func EmptyContractIdentity() ContractIdentity {
+	return ContractIdentity{kind: contractIdentityKindEmpty}
+}
+
+func ContractIdentityFromHname(hn Hname) ContractIdentity {
+	return ContractIdentity{hname: hn, kind: contractIdentityKindHname}
+}
+
+func ContractIdentityFromEvmAddress(addr common.Address) ContractIdentity {
```
dessaya:
```suggestion
func ContractIdentityFromEVMAddress(addr common.Address) ContractIdentity {
```

packages/isc/contract_identity.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (c *ContractIdentity) Hname() (Hname, error) {
+	if c.kind == contractIdentityKindHname {
+		return c.hname, nil
+	}
+	return 0, fmt.Errorf("not an Hname contract")
+}
+
+func (c *ContractIdentity) EvmAddress() (common.Address, error) {
```
dessaya:
```suggestion
func (c *ContractIdentity) EVMAddress() (common.Address, error) {
```

packages/isc/contract_identity.go
```diff
@@ -0,0 +1,99 @@
...
+	if c.kind == contractIdentityKindHname {
+		return c.evmAddr, nil
+	}
+	return common.Address{}, fmt.Errorf("not an Evm contract")
```
dessaya:
```suggestion
	return common.Address{}, fmt.Errorf("not an EVM contract")
```

packages/solo/ledgerl1l2.go
```diff
@@ -27,6 +28,24 @@ func (ch *Chain) L2Accounts() []isc.AgentID {
...
+				aid = isc.NewEthereumAddressAgentID(ethAddr, ch.ChainID)
+			default:
+				panic("bad key length")
+			}
+			ret = append(ret, aid)
+			continue
+		}
 		aid, err := codec.DecodeAgentID([]byte(key))
 		require.NoError(ch.Env.T, err)
 		ret = append(ret, aid)
```
dessaya: Maybe this should be a utility function somewhere? Like `accounts.AgentIDFromKey()`

dessaya (approved):

packages/isc/contract_identity.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (c *ContractIdentity) Hname() (Hname, error) {
+	if c.kind == contractIdentityKindHname {
+		return c.hname, nil
+	}
+	return 0, fmt.Errorf("not an Hname contract")
+}
+
+func (c *ContractIdentity) EVMAddress() (common.Address, error) {
+	if c.kind == contractIdentityKindHname {
```
guhyeonjeong, line 93: is this right?

jorgemmsilva: it is not!
unused, so it doesn't cause a bug, but it for sure would in the future, nice catch. thanks :+1:

PR #2879, opened 2023-09-15
fix: Fix burnlog not presented

howjmay:
burnlog is not presented in the solo tests which is caused by burnlog is not serializable.

### DIFF
--- a/packages/vm/core/evm/evmtest/utils_test.go
+++ b/packages/vm/core/evm/evmtest/utils_test.go
@@ -90,6 +90,7 @@ func initEVM(t testing.TB, nativeContracts ...*coreutil.ContractProcessor) *solo
 		AutoAdjustStorageDeposit: true,
 		Debug:                    true,
 		PrintStackTrace:          true,
+		GasBurnLogEnabled:        false,
 	})
 	for _, c := range nativeContracts {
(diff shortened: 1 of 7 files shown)


packages/testutil/testdbhash/TestStorageContract.hex
```diff
@@ -1 +1 @@
-0xa400ec6dfc8ab5d3a36a37f72ffc5536b7c250e4a243bdcd043769fba228d18c
```
jorgemmsilva: can we disable the gasburnlog on this test so the hash doesn't change? (I say this because it asserts that this change won't modify the state on live environments)

packages/vm/core/evm/evmtest/utils_test.go
```diff
@@ -90,6 +90,7 @@ func initEVM(t testing.TB, nativeContracts ...*coreutil.ContractProcessor) *solo
 		AutoAdjustStorageDeposit: true,
 		Debug:                    true,
 		PrintStackTrace:          true,
+		DisableBurnLog:           true,
```
jorgemmsilva: nitpick: I find it easier to grok having a "GasBurnLogEnabled" flag that defaults to true

jorgemmsilva (approved):

PR #2881, opened 2023-09-18
fix: Fix blobHash collision

howjmay:


howjmay:
The python code to generate the hash value to compare in this PR
```python
import hashlib
import codecs

# byte string to be converted
b_string = b'\x04\x00\x00\x00'

# decoding the byte string to unicode string
u_string = codecs.decode(b_string, 'utf-8')
text = u_string+"key0"+"val0"+u_string+"key1"+"val1"
print(u_string)
hashObject = hashlib.blake2b(text.encode('utf-8'), digest_size=32)
digest = hashObject.hexdigest()

print(digest)
```

packages/vm/core/blob/internal.go
```diff
@@ -25,11 +26,13 @@ func mustGetBlobHash(fields dict.Dict) (hashing.HashValue, []kv.Key, [][]byte) {
...
-	for _, k := range sorted {
-		v := fields.Get(k)
+	var prefix [4]byte
+
+	for _, key := range sorted {
+		v := fields.Get(key)
 		values = append(values, v)
-		all = append(all, v)
-		all = append(all, []byte(k))
+		binary.LittleEndian.PutUint32(prefix[:], uint32(len(key)))
```
jorgemmsilva, line 36: we should leave a comment explaining why we need this  :+1:

packages/vm/core/blob/internal_test.go
```diff
@@ -0,0 +1,44 @@
...
+		h, keys, values := mustGetBlobHash(fields)
+		for i, k := range keys {
+			require.Equal(t, fields[k], values[i])
+		}
+
+		resHash, err := hex.DecodeString("2e42c9185213b81ab77236086c13b7e45471c7104d058197a573a4f10b12ab35")
+		require.NoError(t, err)
+		require.Equal(t, resHash, h.Bytes())
+	})
+	t.Run("with escape separator", func(t *testing.T) {
```
jorgemmsilva: I guess this test doesn't make sense anymore

packages/vm/core/blob/internal_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"encoding/hex"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/iotaledger/wasp/packages/kv/dict"
+)
+
+func TestMustGetBlobHash(t *testing.T) {
+	t.Run("normal", func(t *testing.T) {
```
jorgemmsilva, line 13: we should add a test that asserts the collision doesn't happen anymore, right?
 so `{"ab", "123"}` yields a different hash than `{"ab1","23"}`

jorgemmsilva (approved):

PR #2885, opened 2023-09-21
fix(evm): allow estimateGas with no funds

dessaya:


packages/vm/vmimpl/runreq.go
```diff
@@ -327,6 +327,10 @@ func (reqctx *requestContext) calculateAffordableGasBudget() (budget, maxTokensT
 		gasBudget = reqctx.vm.chainInfo.GasLimits.MinGasPerRequest
 	}
 
+	if reqctx.vm.task.EstimateGasMode {
+		return gasBudget, math.MaxUint64
+	}
```
jorgemmsilva, line 332: won't this result in "false positive" estimations?

basically in situations where the estimation would result in `insufficient funds for gas * price + value`

dessaya: Yes, certainly. I don't think there is a solution that fits every use case. I'd imagine that it should be possible to estimate the gas before actually depositing the funds, but we should probably adopt the same behavior than other networks.

cc @fijter

fijter: Even if it's 'wrong' we should do the same thing other networks are doing here yeah :)

jorgemmsilva (approved):

PR #2888, opened 2023-09-23
fix: Make WriteBytes size smaller

howjmay:


jorgemmsilva (changes requested):
This is currently uint32 because that's required in some places iirc

kape1395:
Indeed.

howjmay:
@jorgemmsilva uploading blob needs uint32. I am checking how can I solve this issue in the audit in other way

PR #2892, opened 2023-09-27
test: add NFT deposit estimation cluster test

jorgemmsilva:
- enables gasburnlog on ISC estimategas calls
- NFTL1Mint functionality for cluster tests 
- new test

### DIFF
--- a/packages/chainutil/runvm.go
+++ b/packages/chainutil/runvm.go
@@ -31,7 +31,7 @@ func runISCTask(
 		TimeAssumption:       blockTime,
 		Entropy:              hashing.PseudoRandomHash(nil),
 		ValidatorFeeTarget:   accounts.CommonAccount(),
-		EnableGasBurnLogging: false,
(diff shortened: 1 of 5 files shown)


packages/chainutil/runvm.go
```diff
@@ -31,7 +31,7 @@ func runISCTask(
 		TimeAssumption:       blockTime,
 		Entropy:              hashing.PseudoRandomHash(nil),
 		ValidatorFeeTarget:   accounts.CommonAccount(),
-		EnableGasBurnLogging: false,
+		EnableGasBurnLogging: estimateGasMode,
```
dessaya, line 34: This will use extra memory in production, right? But the burn log is never logged or returned to the caller...

dessaya (approved):

packages/chainutil/runvm.go
jorgemmsilva: very little extra memory. it will on the chain/estimate-gas endpoint. will DM an example

PR #2903, opened 2023-10-03
feat(jsonrpc): configurable eth_getLogs limits

dessaya:
Adds:

```
      --webapi.limits.jsonrpc.maxBlocksInLogsFilterRange int      maximum amount of blocks in eth_getLogs filter range (default 1000)
      --webapi.limits.jsonrpc.maxLogsInResult int                 maximum amount of logs in eth_getLogs result (default 10000)
```


components/jsonrpc/component.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/iotaledger/hive.go/app"
+	"github.com/iotaledger/wasp/packages/evm/jsonrpc"
+)
+
+var Component = &app.Component{
+	Name: "JSONRPC",
+	Params: &app.ComponentParams{
+		Params: map[string]any{
+			"jsonrpc": jsonrpc.Params,
```
jorgemmsilva: I mean, this works - but it breaks the pattern of how we use components/packages. Also it is mutating a package variable, which feels off.

Usually we receive the parameters in some constructor and pass it down as required. In this case these parameters could be part of the webapi configuration

jorgemmsilva (approved):

PR #2904, opened 2023-10-04
fix(jsonrpc): require tx.GasPrice == chain gas price

dessaya:


packages/evm/jsonrpc/evmchain.go
```diff
@@ -185,19 +185,22 @@ func (e *EVMChain) SendTransaction(tx *types.Transaction) error {
 		return fmt.Errorf("invalid transaction nonce: got %d, want %d", tx.Nonce(), expectedNonce)
 	}
 
-	if err := e.checkEnoughL2FundsForGasBudget(sender, tx.Gas()); err != nil {
+	gasFeePolicy := e.GasFeePolicy()
+	if err := e.checkEnoughL2FundsForGasBudget(sender, tx.Gas(), gasFeePolicy); err != nil {
+		return err
+	}
+	if err := evmutil.CheckGasPrice(tx, gasFeePolicy); err != nil {
```
jorgemmsilva, line 192: we should probably add a test to assert this check is happening.

jorgemmsilva (approved):

packages/evm/jsonrpc/evmchain.go
dessaya: It's `TestEVMGasPriceMismatch`

PR #2918, opened 2023-10-09
feat: (webapi) add endpoint for mempool visualization

jorgemmsilva:


packages/webapi/controllers/chain/mempool.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/iotaledger/wasp/packages/webapi/controllers/controllerutils"
+)
+
+func (c *Controller) getMempoolContents(e echo.Context) error {
+	controllerutils.SetOperation(e, "get_mempool_contents")
+	ch, _, err := controllerutils.ChainFromParams(e, c.chainService)
+	if err != nil {
+		return err
+	}
+	return e.Stream(http.StatusOK, "application/octet-stream", ch.GetMempoolContents())
```
dessaya, line 17: Why binary instead of JSON?

dessaya (approved):

packages/webapi/controllers/chain/mempool.go
jorgemmsilva: I wanted to do the minimum work on the server side as possible. Also to avoid having a large amount of memory used (essentially the entire mempool) it made sense to me to use this approach.
I guess each individual request could be json encoded so it's easier to parse on the client, but it should be a stream to avoid big memory allocation

packages/chain/mempool/mempool.go
```diff
@@ -303,6 +306,18 @@ func (mpi *mempoolImpl) AccessNodesUpdated(committeePubKeys, accessNodePubKeys [
 	}
 }
 
+func (mpi *mempoolImpl) writeContentAndClose(pw *io.PipeWriter) {
+	mpi.onLedgerPool.WriteContent(pw)
+	mpi.offLedgerPool.WriteContent(pw)
+	pw.Close()
```
howjmay: should we use defer to ensure the close is execute?

jorgemmsilva: good point

PR #2933, opened 2023-10-11
feat: Add DelPrefix and refactor array.Erase()

howjmay:


packages/kv/kv.go
```diff
@@ -39,7 +39,7 @@ type KVWriter interface {
 	Set(key Key, value []byte)
 	Del(key Key)
 
-	// TODO add DelPrefix(prefix []byte)
+	DelPrefix(prefix []byte)
```
dessaya: I'm not sure about this one... the intention of the TODO was to add a `DelPrefix()` function that can do the job in O(1), or at least more efficiently than a simple iteration, assuming that the underlying store can do that.

This implementation can just be rewritten as a free function, without of modifying the interfaces:

```
func KeysWithPrefix(kv KVIterator, prefix []byte) []Key {
  var ks []Key
  kv.IterateKeys(prefix, func(key Key, value []byte) bool {
    ks = append(ks, key)
    return true    
  })
  return ks
}

func DelPrefix(kv KVStore, prefix []byte) {
  for _, k := range KeysWithPrefix(kv, prefix) {
    kv.Del(k)
  }
}
```

packages/kv/dict/dict.go
```diff
@@ -83,6 +83,14 @@ func (d Dict) Del(key kv.Key) {
 	delete(d, key)
 }
 
+func (d Dict) DelPrefix(prefix []byte) {
+	for k := range d {
+		if bytes.HasPrefix([]byte(k), prefix) {
+			delete(d, k)
```
dessaya: Modifying a data structure inside the iteration loop might be unsafe.

https://go.dev/ref/spec#For_statements

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

packages/kv/kv.go
howjmay: ummm or I should close this PR first. It may take a lot of time to figure out the O(1) implementation? Especially, there are multiple data structure to implement

dessaya: I think it's just not possible to do it in O(1) with hive's `kvsore` as the backend. I'd forget about this TODO, it's not important.

howjmay: I will modify this PR to remove the TODO only

dessaya (approved):

PR #2950, opened 2023-10-16
Fix how access nodes considers a loosing branch.

kape1395:


packages/chain/chainmanager/var_access_nodes_state.go
```diff
@@ -115,6 +127,21 @@ func (vas *varAccessNodeStateImpl) BlockConfirmed(confirmed *isc.AliasOutputWith
...
 		})
+		for si := stateIndex + 1; ; si++ {
+			es, ok := vas.pending.Get(si)
+			if !ok {
+				break
+			}
+			es = lo.Filter(es, func(e *varAccessNodeStateEntry, _ int) bool {
+				if lo.ContainsBy(loosing, func(o *isc.AliasOutputWithID) bool { return o.OutputID() == e.consumed }) {
+					vas.log.Debugf("⊳ Removing[loosing_fork,consumes=%v] %v", e.consumed, e.output)
+					loosing = append(loosing, e.output)
```
jorgemmsilva: is this line correct? why do we append the output again to the "losing" collection?

packages/chain/chainmanager/var_access_nodes_state.go
```diff
@@ -106,7 +106,19 @@ func (vas *varAccessNodeStateImpl) BlockConfirmed(confirmed *isc.AliasOutputWith
 	stateIndex := confirmed.GetStateIndex()
 	vas.confirmed = confirmed
 	if vas.isAliasOutputPending(confirmed) {
+		// Clean all the outputs that are older (by StateIndex) than the new confirmed output.
+		// Also, clean all the blocks that have the higher index, but are know to be based
+		// on a block from a loosing branch.
+		loosing := []*isc.AliasOutputWithID{}
```
jorgemmsilva: typo: "loosing" -> "losing"

kape1395: Thanks. Fixed.

packages/chain/chainmanager/var_access_nodes_state.go
kape1395: Yes. First, I find the base (oldest AO) on the losing branch. Then, collect all the AOs that are based on that losing AO. If I find an AO that consumes a losing AO, I drop it and add it to the buffer of losing AOs to clear outputs that consume them.

jorgemmsilva (approved):

PR #2951, opened 2023-10-16
Feat: mempool TTL

jorgemmsilva:


packages/chain/node.go
```diff
@@ -895,6 +896,20 @@ func (cni *chainNodeImpl) ensureConsensusInst(ctx context.Context, needConsensus
 	}
 
 	consensusInstance, _ := consensusInstances.Get(logIndex)
+
+	// TODO might not be the best place.. where to do this exactly?
```
jorgemmsilva: @kape1395

kape1395: The place is OK. I would intuitively send that message on a timer, but that will work, too.

kape1395 (approved):

PR #2962, opened 2023-10-18
Feat/iscutils tests

cwarnerdev:
Adds solo tests for iscutils package
Notable change: `vm/core/evm/evmtest` boilerplate for spinning up solo to test contracts has been modified to be exported from the package. It was previously all private. This allows use of it in `tools/evm/iscutils` to easily write tests.


packages/vm/core/evm/evmtest/contractInstance.go
```diff
@@ -0,0 +1,166 @@
...
+type EVMContractInstance struct {
+	chain         *SoloChainEnv
+	defaultSender *ecdsa.PrivateKey
+	address       common.Address
+	abi           abi.ABI
+}
+
+type CallFnResult struct {
+	tx         *types.Transaction
+	EvmReceipt *types.Receipt
```
dessaya: Nit: following Go naming conventions [*](https://github.com/golang/go/wiki/CodeReviewComments#initialisms):

```suggestion
	EVMReceipt *types.Receipt
```

Also nit: kinda arbitrary that `EVMReceipt` is public but `iscReceipt` is not. Same for the other structs made public.

dessaya (approved):

packages/vm/core/evm/evmtest/contractInstance.go
cwarnerdev: Good point 👍

PR #2969, opened 2023-10-22
feat: Set local lib as dependency for developing

howjmay:


BugFreeSoftware:
This feature was already available

PR #2970, opened 2023-10-22
fix: Fix wasmclient

howjmay:


BugFreeSoftware:
What exactly is being fixed here?

packages/wasmvm/wasmclient/src/isc/keypair.rs
```diff
@@ -19,7 +19,7 @@ impl KeyPair {
         if seed_bytes.len() != 0 {
             seed.copy_from_slice(seed_bytes);
         }
-        let key = ed25519::SecretKey::from_bytes(seed);
+        let key = ed25519::SecretKey::from_bytes(&seed);
```
howjmay, line 22: Reference is used here

packages/wasmvm/wasmclient/src/isc/keypair.rs
```diff
@@ -63,7 +64,7 @@ impl KeyPair {
 impl Clone for KeyPair {
     fn clone(&self) -> Self {
         return KeyPair {
-            private_key: ed25519::SecretKey::from_bytes(self.private_key.to_bytes()),
+            private_key: ed25519::SecretKey::from_bytes(&self.private_key.to_bytes()),
```
howjmay, line 67: and here

howjmay:
But I am thinking what cause my formatting so different from yours. I may uninstalled my plugin and reinstall again

packages/wasmvm/wasmclient/src/isc/keypair.rs
BugFreeSoftware: Why? It works already and the from_bytes does not expect a reference

BugFreeSoftware: In fact, seed IS a reference, but from_bytes expects a copy

howjmay: In my local side, I need to do these changes otherwise I can't even pass compilation. Maybe something I missed. Let me check it again

BugFreeSoftware: What Rust version are you using? I'm on the latest (1.73.0). Although it worked with all previous versions as well. And looking at the function and its parameters there should be no need for an &

howjmay: I am using `rustc 1.71.0 (8ede3aae2 2023-07-12)`
But still feel strange the function depends on iota_crypto, so different rust version shouldn't make difference tho

BugFreeSoftware: Maybe delete all your Cargo.lock files. It may be that it pulls in a different version of the crypto lib. I will do the same and see if that is the problem.

howjmay: here is the wasp contract/client I am using
https://github.com/howjmay/wasp-contract-example

howjmay: But I think one problem is even though I did the change here, the tests still passed. It means the current tests in wasp don't well cover this part. Otherwise it is so strange that both passed

BugFreeSoftware: Okay it was on my end. They changed the signature of the from_bytes function. So now I get the same error.

BugFreeSoftware (approved):

PR #2971, opened 2023-10-23
feat: show deposits from L1 to EVM

jorgemmsilva:


packages/vm/core/evm/emulator/emulator.go
```diff
@@ -366,3 +368,25 @@ func (c *chainContext) Engine() consensus.Engine {
 func (c *chainContext) GetHeader(common.Hash, uint64) *types.Header {
 	panic("not implemented")
 }
+
+// create a fake tx from isc magic to a target address so that L1 deposits can be tracked on the EVM
+func (e *EVMEmulator) CreateL1Deposit(
```
dessaya: This feels like it should be responsibility of evmimpl, not EVMEmulator?

packages/vm/core/evm/emulator/emulator.go
```diff
@@ -366,3 +368,25 @@ func (c *chainContext) Engine() consensus.Engine {
...
+) (*types.Transaction, *types.Receipt) {
+	// create a fake tx from the isc magic contract
+	value := util.BaseTokensDecimalsToEthereumDecimals(assets.BaseTokens, e.ctx.BaseTokensDecimals())
+	nonce := uint64(0)
+	tx := types.NewTransaction(nonce, addr, value, 0, util.Big0, assets.Bytes())
+
+	// create a fake receipt
+	receipt := &types.Receipt{
+		Type:              types.LegacyTxType,
+		CumulativeGasUsed: e.BlockchainDB().getPendingCumulativeGasUsed(),
```
dessaya: This made me realize that the CumulativeGas calculation may be buggy, because getPendingCumulativeGasUsed() returns the CumulativeGas from the receipt array, which is empty until AddTransaction() is called.

Not trivial to reproduce, since we would need several txs in the same block.

If it is indeed a bug, I think it may be fixed by moving the calculation of CumulativeGasUsed to AddTransaction().

dessaya (approved):

packages/vm/core/evm/emulator/emulator.go
jorgemmsilva: I placed this here because `getPendingCumulativeGasUsed` is not available outside (on evmimpl)

PR #2990, opened 2023-10-24
fix: do no crash if tx for a receipt does not have a sender

jorgemmsilva:


packages/evm/jsonrpc/types.go
```diff
@@ -164,12 +165,17 @@ func RPCMarshalReceipt(r *types.Receipt, tx *types.Transaction) map[string]inter
 		r.Logs = []*types.Log{}
 		r.Bloom = types.CreateBloom(types.Receipts{r})
 	}
+	var sender common.Address
+	if v, r, s := tx.RawSignatureValues(); !(util.IsZeroBigInt(v) && util.IsZeroBigInt(r) && util.IsZeroBigInt(s)) {
+		// leave sender empty for unsigned txs
+		sender = evmutil.MustGetSender(tx)
+	}
```
dessaya: Maybe move this to `evmutil` next to `MustGetSender()`? Like `GetSenderIfTxSigned()` or similar.

dessaya (approved):

PR #3000, opened 2023-10-25
feat: Simplify ts wasm contract compilation

howjmay:
If wasp path is not set, then alter will jump out to notify the possible error. 
If wasp path is set, then schema tool will automatically run npm install for dependencies.


BugFreeSoftware:
There are a number of problems with this approach. We need npm packages instead. That will solve it all.

PR #3002, opened 2023-10-25
fix: evm l1 deposit receipt

jorgemmsilva:


packages/vm/core/evm/evmimpl/impl.go
```diff
@@ -442,25 +442,32 @@ func newL1Deposit(ctx isc.Sandbox) dict.Dict {
...
-		GasUsed:           0,
-		Logs:              make([]*types.Log, 0),
+		Type:   types.LegacyTxType,
+		Logs:   make([]*types.Log, 0),
+		Status: types.ReceiptStatusSuccessful,
 	}
 	receipt.Bloom = types.CreateBloom(types.Receipts{receipt})
 
 	ctx.Privileged().OnWriteReceipt(func(evmPartition kv.KVStore) {
+		receipt.GasUsed = gas.ISCGasBurnedToEVM(ctx.Gas().Burned(), &chainInfo.GasFeePolicy.EVMGasRatio)
```
dessaya: Why do the calculation of the gas inside the callback? Using `ctx` here is a smell IMO, since the `ctx` is a captured variable and may not be usable anymore here. Better to receive any needed data as parameter in the callback.

Also: the call to `newL1Deposit` is typically done as part of an ISC request, right? And it may involve many other calls to other ISC contracts. Is this `GasUsed` the used gas by the whole ISC request, or just the `newL1Deposit` call?

jorgemmsilva: I agree but was the only way I found of getting the final gas usage. That function in particular is still usable at the time of writing the receipt.
Alternative would be to pass extra info in the callback function. I like that idea

dessaya: Also, please consider my second question: Can `newL1Deposit` be called twice in the same ISC request? What would happen with the `GasUsed` values then? Does `ctx.Gas().Burned()` return the gas burned by the ISC request or just the current call?

jorgemmsilva: Yes, good point. It's possible. The gas in each produced receipt will be equal to the final gas charged by the request. Not ideal, but seems okay

dessaya: Also: can an EVM tx call an ISC contract which calls `newL1Deposit`? :exploding_head:

jorgemmsilva: I added a guard to avoid potencialy weird scenarios:
https://github.com/iotaledger/wasp/pull/3002/files#diff-33438304417133af8f0877850a64c3ed3b17bde9ef869d751f5aa12008cf779cR79

`newL1Deposit` can only be called by `accounts.TransferAllowanceTo`

dessaya (approved):

Issue #3027, opened 2023-11-03
Missing values in Receipt


MoonLabsDev:
As stated in the official documentation:
https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactionreceipt

currently your receipt doesnt expose the values "type" and "effectiveGasPrice". We use effectiveGasPrice in one of our bots to multiply it with "gasUsed" to get the actually spend SMR amount (and then convert it to $, to show the $ cost of a transaction).
To be fully complient it would be nice of you add those 2 values to your list. In your case they are currently both static.

- effectiveGasPrice : QUANTITY - The sum of the base fee and tip paid per unit of gas.
- type: QUANTITY - integer of the transaction type, 0x0 for legacy transactions, 0x1 for access list types, 0x2 for dynamic fees.

So in your case effectiveGasPrice would be 1000 GWEI (so 1000 * 10^9) as hex string, and type would be "0x0"

As at some point more protocols and dApps from other chains will come to shimmer you should try to be as complient as possible with the EVM standard, so that they can migrate without having to make alot of adjustments "just for Shimmer".

trungtt198x:
@MoonLabsDev 
You mentioned on Discord that you'd wanna make an PR.
Below can be the relevant code to add `effectiveGasPrice`
https://github.com/iotaledger/wasp/blob/develop/packages/evm/jsonrpc/types.go#L175

MoonLabsDev:
https://github.com/iotaledger/wasp/pull/3028

MoonLabsDev:
another solution might be

r.EffectiveGasPrice
and r.Type instead of the static values, if you fill these value correctly, but i am not a GO developer so it is hard for me to read that code (first time I see GO)

PR #3028, opened 2023-11-03
Fixed Issue #3027

MoonLabsDev:
Fix for issue https://github.com/iotaledger/wasp/issues/3027 as documented in https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactionreceipt


jorgemmsilva (changes requested):
this needs to be dynamic and change with the chain price policy, hardcoded values are not correct.

PR #3050, opened 2023-11-07
Rebroadcasting requests is enabled and configurable

Juliusan:
Rebroadcasting of off-ledger requests is now once again enabled. To avoid it causing problems, the time period for it can be configured via `chains.broadcastInterval` parameter. It is set to 1 hour by default.

### DIFF
--- a/components/chains/params.go
+++ b/components/chains/params.go
@@ -8,7 +8,7 @@ import (
 
 type ParametersChains struct {
 	BroadcastUpToNPeers              int           `default:"2" usage:"number of peers an offledger request is broadcasted to"`
-	BroadcastInterval                time.Duration `default:"5s" usage:"time between re-broadcast of offledger requests"`
+	BroadcastInterval                time.Duration `default:"1h" usage:"time between re-broadcast of offledger requests"`
 	APICacheTTL                      time.Duration `default:"300s" usage:"time to keep processed offledger requests in api cache"`
 	PullMissingRequestsFromCommittee bool          `default:"true" usage:"whether or not to pull missing requests from other committee members"`
 	DeriveAliasOutputByQuorum        bool          `default:"true" usage:"false means we propose own AliasOutput, true - by majority vote."`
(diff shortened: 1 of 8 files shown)


tools/cluster/templates/waspconfig.go
```diff
@@ -87,7 +87,7 @@ var WaspConfig = `
   },
   "chains": {
     "broadcastUpToNPeers": 2,
-    "broadcastInterval": "5s",
+    "broadcastInterval": "1s",
```
jorgemmsilva: won't this make the spammy tests even more exhausting to run?

components/chains/params.go
```diff
@@ -8,7 +8,7 @@ import (
 
 type ParametersChains struct {
 	BroadcastUpToNPeers              int           `default:"2" usage:"number of peers an offledger request is broadcasted to"`
-	BroadcastInterval                time.Duration `default:"5s" usage:"time between re-broadcast of offledger requests"`
+	BroadcastInterval                time.Duration `default:"1h" usage:"time between re-broadcast of offledger requests"`
```
jorgemmsilva: should be possible to disable this feature altogether IMO, maybe accepting 0 as disabled

jorgemmsilva (approved):

components/chains/params.go
Juliusan: Karolis offered it. I went with easier solution :-) I'll implement the zero value option then. I have to check how `0` durations compare to each other :-)

tools/cluster/templates/waspconfig.go
Juliusan: Probably. I was happy regular cluster tests ran without any problems. I'll check the spam tests than.

kape1395 (approved):
I agree with the other comments.

tools/cluster/templates/waspconfig.go
jorgemmsilva: just set it to a high value, no? we don't even test this stuff in the cluster tests afaik.

Issue #3072, opened 2023-11-13
Value transactions calling ISCMagic can utilize wrong account


lmoe:
If the `ISCMagic.Send` is not called as documented, the sent value can get stuck in the wrong account.

PR #3073, opened 2023-11-13
fix: move value transactions to the correct account 

lmoe:
# Description of change

Currently all value transactions calling ISCMagic deposit into the 0x1074 account instead of the common account. 

This PR does a few things:

* It will block and revert all value transactions calling contract functions that are not marked as `payable`.
* It will extend `ISCMagic.Send` to enable `native` value transactions. 
  * (This adds Metamask token display support) 
* `ISCMagic.Send` still supports the other way of sending base tokens (Assets metadata).

fixes issue #3071 
fixes issue #3072 

## Type of change

Two tests were added to prove correct behavior:
`TestSendPayableValueTX` 
`TestSendUnpayableValueTX`


packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
```diff
@@ -68,21 +130,17 @@ func (h *magicContractHandler) Send(
...
 	}
-	// 	id := nftID.Unwrap()
-	h.adjustStorageDeposit(req)
 
-	// make sure that allowance <= sent tokens, so that the target contract does not
-	// spend from the common account
-	if !req.Assets.Spend(req.Metadata.Allowance) {
-		panic(errInvalidAllowance)
-	}
+	evmCallData := h.ctx.Request().EVMCallData()
```
dessaya: I think here we should use the `value` received from the EVM call, which is not necessarily the same `value` sent in the tx.

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
```diff
@@ -53,6 +55,66 @@ func (h *magicContractHandler) TakeAllowedFunds(addr common.Address, allowance i
...
+	req.Assets = isc.NewAssets(adjustedTxValue.Uint64(), req.Assets.NativeTokens, req.Assets.NFTs...)
+
+	h.adjustStorageDeposit(req)
+
+	if !req.Assets.Spend(req.Metadata.Allowance) {
+		panic(errInvalidAllowance)
+	}
+
+	// This should be the 0x107400[...] address
+	evmAddr := isc.NewEthereumAddressAgentID(h.ctx.ChainID(), *evmCallData.To)
```
dessaya: Same here: instead of `evmCallData.Value`, just use the `value` received from the EVM. And instead of `evmCallData.To`, just use the `iscmagic.Address`

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
```diff
@@ -53,6 +55,66 @@ func (h *magicContractHandler) TakeAllowedFunds(addr common.Address, allowance i
...
+// This will ignore the base token inside `req.Assets.BaseToken` and will use the proper TxValue of the Tx.
+// By design, all transactions reaching the EVM will cause a value transfer from the sender to the recipient.
+// In our case, this address is 0x1074000[...]. Which means that before *this* function is reached, the value has already
+// been moved from the sender to the 0x1074000[...] address.
+// The native assets and the NFT still remains inside the senders account as this is Wasp specific.
+// Therefore, this function first moves the sent base token from the 0x1074[...] address to the common account
+// Then moves the native assets and NFT from the **sender account** to the common account.
+// Then moves all three assets to the actual target address on behalf of the sender.
+func (h *magicContractHandler) sendViaTxValue(req isc.RequestParameters, evmCallData *isc.EVMCallData) {
+	adjustedTxValue := new(big.Int).Div(evmCallData.Value, big.NewInt(1e12))
```
jorgemmsilva: we have util functions for this, check `EthereumDecimalsToBaseTokenDecimals` and `BaseTokensDecimalsToEthereumDecimals`

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
```diff
@@ -68,21 +130,17 @@ func (h *magicContractHandler) Send(
...
+		evmCallData.Value.BitLen() == 0 {
+		h.sendViaMetadata(req)
+		return
+	}
 
-	h.ctx.Privileged().SendOnBehalfOf(
-		isc.ContractIdentityFromEVMAddress(h.caller.Address()),
-		req,
-	)
+	h.sendViaTxValue(req, evmCallData)
```
jorgemmsilva: Why do we need `sendViaMetadata` and `sendViaTxValue` ? 
Couldn't we handle the msg.value separately and sum everything up in a single `Allowance` object?

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
lmoe: Just an oversight. :D

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
lmoe: I guess we could. Found it cleaner and easier to read by separating the behavior, that's all.

packages/vm/core/evm/evmimpl/iscmagic_handler.go
```diff
@@ -21,16 +22,18 @@ import (
 // For example, if ISC.getL2NFTs() is called from solidity, this will
 // correspond to a call to [GetL2NFTs].
 type magicContractHandler struct {
-	ctx    isc.Sandbox
-	caller vm.ContractRef
+	ctx     isc.Sandbox
+	caller  vm.ContractRef
+	txValue *big.Int
```
dessaya: Nit: I'd call it `callValue` or just `value`. Because again, it's not necessarily the same value set in the tx.

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
```diff
@@ -53,6 +58,69 @@ func (h *magicContractHandler) TakeAllowedFunds(addr common.Address, allowance i
...
+	h.moveAssetsToCommonAccount(req.Assets)
+
+	h.ctx.Privileged().SendOnBehalfOf(
+		isc.ContractIdentityFromEVMAddress(h.caller.Address()),
+		req,
+	)
+}
+
+// sendViaTxValue enables the transfer of assets by sending a transaction with value attached.
+// This will ignore the base token inside `req.Assets.BaseToken` and will use the proper TxValue of the Tx.
```
dessaya: Ignoring the value in the `req.Assets` seems like surprising behavior. Why not add them? Or raise an error if both `req.Assets.BaseToken` and `value` are set.

packages/vm/core/evm/evmtest/evm_test.go
```diff
@@ -590,6 +592,88 @@ func TestISCGetSenderAccount(t *testing.T) {
...
+	require.Nil(t, res.ISCReceipt)
+
+	// L2 balance of evm core contract is 0
+	require.Zero(t, env.Chain.L2BaseTokens(isc.NewContractAgentID(env.Chain.ChainID, evm.Contract.Hname())))
+	// L2 balance of ISC magic contract (0x1074...) is 0
+	require.Zero(t, env.Chain.L2BaseTokens(isc.NewEthereumAddressAgentID(env.Chain.ChainID, iscmagic.Address)))
+	// L2 balance of common account is: 0
+	require.Zero(t, env.Chain.L2BaseTokens(isc.NewContractAgentID(env.Chain.ChainID, 0)))
+	// L2 balance of sender is: initial - value sent in tx - gas fee
+	require.EqualValues(t, senderInitialBalance, env.Chain.L2BaseTokens(isc.NewEthereumAddressAgentID(env.Chain.ChainID, ethAddress)))
```
dessaya: What happened with the gas fee? :thinking:

dessaya (approved):

PR #3093, opened 2023-11-16
chore: refactor Iota2.0 consensus ChainOutputs

lmoe:


packages/chain/cons/cons.go
```diff
@@ -553,12 +554,13 @@ func (c *consImpl) uponVMInputsReceived(aggregatedProposals *bp.AggregatedBatchP
...
-		AnchorOutput:         decidedBaseAnchorOutput.GetAnchorOutput(),
-		AnchorOutputID:       decidedBaseAnchorOutput.OutputID(),
-		Store:                c.chainStore,
-		Requests:             aggregatedProposals.OrderedRequests(requests, *randomness),
-		TimeAssumption:       aggregatedProposals.AggregatedTime(),
+		Processors: c.processorCache,
+		Inputs:     decidedBaseAnchorOutput,
+		Store:      c.chainStore,
+		Requests:   aggregatedProposals.OrderedRequests(requests, *randomness),
+		// TODO: <lmoe> Is TimeAssumption->Timestamp a 1:1 change?
```
jorgemmsilva, line 561: think so, ye

packages/chain/cons/cons.go
```diff
@@ -611,34 +614,44 @@ func (c *consImpl) uponVMOutputReceived(vmResult *vm.VMTaskResult) gpa.OutMessag
...
-		Essence: resultTxEssence,
-		Unlocks: transaction.MakeSignatureAndAliasUnlockFeatures(len(resultTxEssence.Inputs), signatureForUnlock),
+
+	resultInputs, err := resultTx.Inputs()
+	if err != nil {
+		panic(fmt.Errorf("cannot get inputs from result TX: %w", err))
 	}
+
+	// TODO: <lmoe> This is most likely just trash :D
+	tx := &iotago.SignedTransaction{
```
jorgemmsilva, line 632: why do we create a `SignedTransaction` here?

lmoe: Late night uneducated guess, based on the model. It seemed to be the only way to set the Unlocks. 
I added the comment to take a later look again.

jorgemmsilva: yeah, I think you're right :+1: , this seems correct

jorgemmsilva (approved):

packages/chain/cons/cons.go
dessaya: Yes, just a rename

PR #3179, opened 2023-12-13
fix(evm): prevent nil pointer access

dessaya:


packages/vm/core/evm/evmtest/setup.go
```diff
@@ -263,7 +259,7 @@ func (e *SoloChainEnv) maxGasLimit() uint64 {
 }
 
 func (e *SoloChainEnv) DeployContract(creator *ecdsa.PrivateKey, abiJSON string, bytecode []byte, args ...interface{}) *EVMContractInstance {
```
jorgemmsilva, line 261: makes sense to add the value as a parameter (might be used by some contracts)

packages/vm/core/evm/evmimpl/impl.go
```diff
@@ -424,7 +430,9 @@ func callContract(ctx isc.Sandbox) dict.Dict {
 		ctx.Privileged().GasBurnEnable(true)
 		gasErr := panicutil.CatchPanic(
 			func() {
-				ctx.Gas().Burn(gas.BurnCodeEVM1P, gas.EVMGasToISC(res.UsedGas, &gasRatio))
+				if res != nil {
+					ctx.Gas().Burn(gas.BurnCodeEVM1P, gas.EVMGasToISC(res.UsedGas, &gasRatio))
```
jorgemmsilva, line 434: when can `res` be `nil`? do we not charge gas for this tx?

packages/vm/core/evm/evmtest/setup.go
dessaya: Yes, I needed this for Chris' contract that had a `constructor() payable`.

packages/vm/core/evm/evmimpl/impl.go
dessaya: It happened to me when sending a tx with a very small gasLimit, less than the intrinsic gas.

I guess we already charge the minimum gas set in ISC? I'll add a test.

jorgemmsilva: we already handle this in the jsonrpc side:
https://github.com/iotaledger/wasp/blob/develop/packages/evm/jsonrpc/jsonrpctest/env.go#L363-L373
https://github.com/iotaledger/wasp/blob/develop/packages/evm/jsonrpc/waspevmbackend.go#L59-L65

not sure if we should add this `res != nil` check here. could potentially be exploited by some edge case that makes res be empty? :shrug:

dessaya: I just added the test, without the `res != nil` I get nil pointer. Probably because I'm calling from solo instead of jsonrpc.

I don't see how this could add a vulnerability that wasn't there in the first place? The behavior is still the same, the only thing that changes is the error reported ("intrinsic gas" instead of "nil pointer").

jorgemmsilva: You're right. Disregard my last comment :+1:

jorgemmsilva (approved):

PR #3221, opened 2024-01-09
refactor: allow 18 decimals in the EVM

jorgemmsilva:


lmoe (approved):

packages/solo/run.go
```diff
@@ -68,7 +68,6 @@ func (ch *Chain) runTaskNoLock(reqs []isc.Request, estimateGas bool) *vm.VMTaskR
 		// state baseline is always valid in Solo
 		EnableGasBurnLogging: ch.Env.enableGasBurnLogging,
 		EstimateGasMode:      estimateGas,
-		MigrationsOverride:   ch.migrationScheme,
```
dessaya, line 71: Why this change?

packages/util/misc.go
```diff
@@ -34,6 +34,10 @@ func IsZeroBigInt(bi *big.Int) bool {
 	return len(bi.Bits()) == 0
 }
 
+func IsPositiveBigInt(n *big.Int) bool {
+	return n.Cmp(Big0) == 1
```
dessaya: `n.Sign() > 0`?

packages/vm/core/accounts/fungibletokens.go
```diff
@@ -47,6 +47,23 @@ func creditToAccount(state kv.KVStore, accountKey kv.Key, assets *isc.Assets) {
 	}
 }
 
+// CreditToAccount brings new funds to the on chain ledger
+// NOTE: this function does not take NFTs into account
```
dessaya: The `NOTE` is probably not needed since the parameter is already `big.Int`, not `Assets`

packages/solo/solotest/snapshot.db
dessaya, line 1: Is this intended to be committed?

dessaya (approved):
So this change allows accounts to own fractions of a glow, but still on L1 the ledger must be balanced, right? I mean, it must still hold that the sum of all L2 accounts is equal to the L1 assets owned by the chain.

IIRC there are functions in `solo` and `vmimpl` that assert that invariant. Maybe those need to be adapted to include the extra decimals in the sum?

See `CheckLedger` in `packages/vm/core/accounts/checkledger.go` and `MustBalanced` in `packages/vm/vmtxbuilder/totals.go`

packages/solo/run.go
jorgemmsilva: it's not much of an "override" if it always has a value, right? :stuck_out_tongue: 
it was causing issues when testing the migration

packages/solo/solotest/snapshot.db
jorgemmsilva: no, thank you for pointing that out :facepalm:

jorgemmsilva:
@dessaya thanks for the review, good points.
`packages/vm/vmtxbuilder/totals.go` shouldn't need to be updated, as it only uses information from the transaction (inputs/outputs)

`packages/vm/core/accounts/checkledger.go` does make sense to update, working on it :+1:

PR #3236, opened 2024-01-19
fix: allow pre-decimals-migration state to be read

jorgemmsilva:


packages/solo/solotest/snapshot.db
dessaya, line 1: This committed intentionally?

dessaya (approved):

packages/solo/solotest/snapshot.db
jorgemmsilva: no, thanks for noticing

PR #3255, opened 2024-02-06
improve DBHash in preparation for iota2.0

dessaya:
There are now 3 "db hash" variants to choose for a test:

* `VerifyDBHash` works as before: calculates a hash of the low-level DB contents (including the whole trie, all states, blocks, etc).
* `VerifyStateHash` calculates a hash of the whole chain state at the latest state index.
* `VerifyContractStateHash` calculates a hash of the state of a single core contract, at the latest state index. An optional prefix can be supplied to verify just a subset of the keys in the state.

There is also a new `DB_DUMP` environment variable that can be used to store a hex dump of the DB contents, and later compute a diff between two different branches.


packages/solo/solo.go
```diff
@@ -221,8 +223,44 @@ func (env *Solo) batchLoop() {
...
+func (env *Solo) IterateChainTrieDBs(
+	f func(chainID *isc.ChainID, k []byte, v []byte),
+) {
+	env.chainsMutex.Lock()
+	defer env.chainsMutex.Unlock()
+
+	chainIDs := lo.Keys(env.chains)
+	slices.SortFunc(chainIDs, func(a, b isc.ChainID) int { return bytes.Compare(a.Bytes(), b.Bytes()) })
+	for _, chID := range chainIDs {
+		chID := chID // prevent loop variable aliasing
```
jorgemmsilva, line 235: when go 1.22? :pray:

jorgemmsilva (approved):

PR #3288, opened 2024-02-21
refactor: accounts state access

dessaya:


packages/vm/core/accounts/state.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+			return isc.NewContractAgentID(chainID, hn), nil
+		case common.AddressLength:
+			var ethAddr common.Address
+			copy(ethAddr[:], []byte(key))
+			return isc.NewEthereumAddressAgentID(chainID, ethAddr), nil
+		default:
+			panic("bad key length")
+		}
+	}
```
jorgemmsilva, line 122: isn't this missing L1addresses?

jorgemmsilva (approved):

packages/vm/core/accounts/state.go
dessaya: Idk, I didn't write this, I just moved it.

Seems like it is missing? But in that case the bug is also on develop :thinking:

dessaya: Unlike ContractAgentID and EthereumAgentID, AddressAgentID does not include a ChainID, that's why.

jorgemmsilva: that gets handled by `codec.AgentID.Decode`, nevermind :+1:

PR #3289, opened 2024-02-22
Add JWT token import, handle viper config read error

lmoe:
# Description of change

Adds a simple import function for JWT keys that are still remaining inside the config. 

Additionally does some error handling when reading the config file to notify the user of structural issues with an existing config file.

### DIFF
--- a/tools/wasp-cli/authentication/import.go
+++ b/tools/wasp-cli/authentication/import.go
@@ -1,0 +1,38 @@
+package authentication
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/iotaledger/wasp/tools/wasp-cli/cli/config"
+	"github.com/iotaledger/wasp/tools/wasp-cli/log"
+)
+
+func initImportCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "import",
+		Short: "Imports all JWT tokens from the config into the OS Keychain",
+		// Args:  cobra.ArbitraryArgs,
+		Run: func(cmd *cobra.Command, args []string) {
+
+			tokens := config.GetAuthTokenForImport()
+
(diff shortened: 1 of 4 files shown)


tools/wasp-cli/authentication/import.go
```diff
@@ -0,0 +1,35 @@
...
+		Short: "Imports all JWT tokens from the config into the OS Keychain",
+		Run: func(cmd *cobra.Command, args []string) {
+			tokens := config.GetAuthTokenForImport()
+
+			fmt.Println("Importing JWT tokens from the config into the OS Keychain.")
+
+			kc := config.GetKeyChain()
+			for k, v := range tokens {
+				if v == "" {
+					fmt.Printf("Could not import JWT token for node '%v'\n", k)
```
dessaya: Tip: use `%q` instead of `'%v'` (handles non-printable characters better)

tools/wasp-cli/cli/keychain/keychain.go
```diff
@@ -39,7 +39,7 @@ type KeyChain interface {
 }
 
 func jwtTokenKey(node string) string {
-	return fmt.Sprintf("%s.%s", jwtTokenKeyPrefix, node)
+	return fmt.Sprintf("%s.%v", jwtTokenKeyPrefix, node)
```
dessaya, line 42: Just curious: why this change? AFAIK `%v` defaults to `%s` for strings :thinking:

dessaya (approved):

jorgemmsilva (approved):

tools/wasp-cli/cli/keychain/keychain.go
lmoe: Just turned out to be my default as %v can also directly print other types more or less dynamically. While with %s you would get a runtime error when the variable type is different. Taken out of this context, sure a %s would fit "better" in this case. :D

PR #3336, opened 2024-03-14
Add auto token register

lmoe:
# Description of change

Adds an automatic ERC20 and IRC30 registration for native tokens. IRC27 will follow.

removed:
* foundryModify
* foundryDestroy 

is outdated but remains:
* foundryCreate

added:
* nativeTokenCreate
* nativeTokenModify
* nativeTokenDestroy

`nativeTokenCreate` now combines the foundryCreation and the ERC20 token registration, therefore requires `token name`, `token symbol` and `token decimals` as additional arguments.

Furthermore, it stores IRC30 compatible metadata into the foundry output.

* Wasp-clis `create-foundry` was replaced with `create-native-token` and requires additional arguments like above 
  * erc20-register functions are still there for older foundries
* ISCMagic has one additional function `createNativeTokenFoundry` (nothing was removed) 
  * `foundryCreateNew` and `registerERC20NativeToken` remains for compatibility   
* Account core contract was regenerated
* Tests were adjusted and new test cases were added
* Some test structures were renamed from xyzFoundry to xyzNativeToken 
* Fixed a bug in `foundryOutputRec`, which results in a wrong input commitment if metadata is defined (was not serialized correctly)  

It is mostly a breaking change in terms of the `account` core contract as foundry funcs were removed and the wasp-cli as `create-foundry` was replaced.
 
The ISCMagic has only one additional function more.

## Type of change


packages/isc/request_offledger.go
```diff
@@ -42,27 +42,20 @@ var (
 )
 
 type ImpersonatedOffLedgerRequestData struct {
-	req     *OffLedgerRequestData
```
jorgemmsilva, line 44: this is a good change, but it should be it's own PR tbh :/

contracts/wasm/corecontracts/test/core_accounts_test.go
```diff
@@ -195,21 +195,21 @@ func TestFoundryNew(t *testing.T) {
 	require.Equal(t, uint32(1), fnew.Results.FoundrySN().Value())
 }
 
-func TestFoundryModifySupply(t *testing.T) {
+func NativeTokenndryModifySupply(t *testing.T) {
```
jorgemmsilva: typo

and I'm guessing this should start with `TestNative....`

packages/isc/request_offledger.go
```diff
@@ -42,27 +42,20 @@ var (
...
+	OffLedgerRequestData
```
jorgemmsilva, line 44: good change, but this should be its own PR :wrench:

packages/vm/core/accounts/interface.go
```diff
@@ -8,10 +8,15 @@ var Contract = coreutil.NewContract(coreutil.CoreContractAccounts)
...
 var (
 	// Funcs
-	FuncDeposit                = coreutil.Func("deposit")
-	FuncFoundryCreateNew       = coreutil.Func("foundryCreateNew")
-	FuncFoundryDestroy         = coreutil.Func("foundryDestroy")
-	FuncFoundryModifySupply    = coreutil.Func("foundryModifySupply")
+	FuncDeposit = coreutil.Func("deposit")
+
+	// Kept for compatibility reasons
+	FuncFoundryCreateNew = coreutil.Func("foundryCreateNew")
```
jorgemmsilva, line 14: I don't get it.
if this is "kept for compatibility", shouldn't  `foundryDestroy` and `foundryModifySupply` also be? :thinking:

packages/vm/core/evm/evmimpl/restricted.go
```diff
@@ -24,3 +24,26 @@ func restricted(handler coreutil.Handler) coreutil.Handler {
...
+
+		contractCall := caller.(*isc.ContractAgentID)
+		if contractCall.Hname() != contractHname {
+			panic(vm.ErrIllegalCall)
+		}
+
+		fmt.Printf("%v %v\n", caller, contractHname)
+		return handler(ctx)
+	}
+}*/
```
jorgemmsilva: remove?

jorgemmsilva (commented):
nice! :rocket:

packages/vm/core/accounts/interface.go
lmoe: Main reason for that was, that wasp-cli still has a foundry-create command and I wasn't sure where it is still in use. For example in some scripts or something. I didn't want to break things. 

It is also still available in the magic contract, where I was very sure I could break things for 3rd party developers if I just remove it. :D

jorgemmsilva (approved):

PR #3340, opened 2024-03-19
Statemanager fixes

Juliusan:
Some minor fixes to State manager:
* Pipes are closed once state manager is no longer needed
* `Block.Equals` method used instead of custom implementation in State manager tests
* `State.Equals` method moved to `state` packager from state manager; it might be doing too much, but that is what I used in state manager tests.

### DIFF
--- a/packages/state/state.go
+++ b/packages/state/state.go
@@ -1,6 +1,7 @@
 package state
 
 import (
+	"bytes"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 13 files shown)


jorgemmsilva (approved):

packages/state/state.go
```diff
@@ -80,3 +81,42 @@ func (s *state) SchemaVersion() isc.SchemaVersion {
...
+	return true
+}
+
+type commonEntry struct {
+	value1 []byte
+	value2 []byte
+}
+
+func getCommonState(state1, state2 State) map[kv.Key]*commonEntry {
+	result := make(map[kv.Key]*commonEntry)
```
dessaya, line 107: If I understand correctly the size of this map will be O(#keys in the state). Is this intended to run in production? It might take a lot of RAM...

Juliusan: It was running in tests only. But `Equals` method should be able to run in production too. I suppose, it would be enough to test the equality of `TrieRoot` in `Equals`. Will change that in another PR.

PR #3358, opened 2024-03-28
Postpone recovery from rejections for several L1 milestones.

kape1395:


packages/chain/cmt_log/var_output.go
```diff
@@ -7,35 +7,43 @@ import (
 	"github.com/iotaledger/wasp/packages/isc"
 )
 
+const (
+	postponeRecoveryMilestones = 3
```
jorgemmsilva: this should be a parameter so we can re-configure without having to build/re-deploy

jorgemmsilva (approved):

PR #3376, opened 2024-04-23
Feat: mempool gas auction

jorgemmsilva:
adds configurable limits for the mempool of: offledger, on-ledger and timelocked requests

mempool now proposes the requests which have the highest gas price


packages/chain/mempool/mempool.go
```diff
@@ -584,30 +594,43 @@ func (mpi *mempoolImpl) handleConsensusProposal(recv *reqConsensusProposal) {
...
+//nolint:gocyclo
 func (mpi *mempoolImpl) refsToPropose(consensusID consGR.ConsensusID) []*isc.RequestRef {
 	//
 	// The case for matching ChainHeadAO and request BaseAO
 	reqRefs := []*isc.RequestRef{}
 	if !mpi.tangleTime.IsZero() { // Wait for tangle-time to process the on ledger requests.
-		mpi.onLedgerPool.Filter(func(request isc.OnLedgerRequest, _ time.Time) bool {
-			if isc.RequestIsExpired(request, mpi.tangleTime) {
-				return false // Drop it from the mempool
+		mpi.onLedgerPool.Iterate(func(e *typedPoolEntry[isc.OnLedgerRequest]) bool {
```
kape1395, line 603: We have to iterate according to some order, e.g., gas, to keep the intersections not too small.

kape1395 (approved):

packages/chain/mempool/mempool.go
jorgemmsilva: that's the case, `Iterate` uses `olp.ordered` :+1:

PR #3410, opened 2024-05-23
evm: emit Transfer events on L1 deposits

dessaya:


packages/vm/core/evm/evmimpl/external.go
```diff
@@ -17,14 +17,14 @@ func Nonce(evmPartition kv.KVStoreReader, addr common.Address) uint64 {
...
-func RegisterERC721NFTCollectionByNFTId(store kv.KVStore, nft *isc.NFT) {
+func RegisterERC721NFTCollectionByNFTId(evmState kv.KVStore, nft *isc.NFT) {
 	metadata, err := isc.IRC27NFTMetadataFromBytes(nft.Metadata)
 	if err != nil {
 		panic(errEVMCanNotDecodeERC27Metadata)
 	}
 
 	addr := iscmagic.ERC721NFTCollectionAddress(nft.ID)
-	state := emulator.NewStateDBFromKVStore(emulator.StateDBSubrealm(store))
+	state := emulator.NewStateDBFromKVStore(evm.EmulatorStateSubrealm(evmState))
```
dessaya, line 27: @lmoe @jorgemmsilva 

I think this was a bug because it was calling emulator.StateDBSubrealm() twice (once here and again inside NewStateDBFromKVStore.

I'm not sure about the implications of the bug, I think the erc721 contract was being stored under some bogus key. I also don't know how this affects the production instances.

packages/vm/core/evm/evmimpl/impl.go
```diff
@@ -465,10 +468,38 @@ func newL1Deposit(ctx isc.Sandbox) dict.Dict {
...
+		logs = append(logs, makeTransferEvent(iscmagic.ERC721NFTsAddress, addr, iscmagic.WrapNFTID(nftID).TokenID()))
+
+		// if the NFT belongs to a collection, emit a Transfer event from the corresponding ERC721NFTCollection contract
+		if nft := ctx.GetNFTData(nftID); nft != nil {
+			if collectionNFTAddress, ok := nft.Issuer.(*iotago.NFTAddress); ok {
+				collectionID := collectionNFTAddress.NFTID()
+				erc721CollectionContractAddress := iscmagic.ERC721NFTCollectionAddress(collectionID)
+				stateDB := emulator.NewStateDB(newEmulatorContext(ctx))
+				if stateDB.Exist(erc721CollectionContractAddress) {
+					logs = append(logs, makeTransferEvent(erc721CollectionContractAddress, addr, iscmagic.WrapNFTID(nftID).TokenID()))
```
jorgemmsilva, line 491: won't this produce 2 events if the NFT belongs to a collection? I think it should only produce 1 event (from the collection contract)

dessaya: Yes, it produces 2 events. I'm not sure what is the expected behavior; I guessed it was like this but I can change it if it's wrong.

jorgemmsilva: only 1 event should be produced IMO

jorgemmsilva (approved):

PR #3448, opened 2024-06-24
Docs/move get balance.sol from wiki

salaheldinsoliman:
# Description of change
This is the first PR addressing moving Wiki contracts to the wasp repo.
A `GetBalance.sol` contract was added, alongside a `solo` test for it.

fixes issue: https://github.com/iotaledger/iota-wiki/issues/1611 

A `Solo` test.

## Change checklist

- [x] I have followed the [contribution guidelines](/CONTRIBUTING.md) for this project


packages/vm/core/evm/evmtest/evm_test.go
```diff
@@ -2729,3 +2730,75 @@ func TestDisableMagicWrap(t *testing.T) {
...
+	decimals := env.Chain.EVM().BaseToken().Decimals
+	var value uint64
+	instance.CallFnExpectMultipleEvents(nil, "GotBaseBalance", &value, "getBalance", nativeTokenIDBytes)
+	realBalance := util.BaseTokensDecimalsToEthereumDecimals(value, decimals)
+	assert.Equal(t, balance, realBalance)
+
+	// test 2
+	// get the agnetId of the contract deployer
+	// and compare it with the agentId returned by the contract
+	var adgntID []byte
```
jorgemmsilva: typo

packages/vm/core/evm/evmtest/contractInstance.go
```diff
@@ -152,6 +153,21 @@ func (e *EVMContractInstance) CallFnExpectEvent(opts []ethCallOptions, eventName
...
+	require.Equal(e.chain.t, types.ReceiptStatusSuccessful, res.EVMReceipt.Status)
+	topic := e.abi.Events[eventName].ID
+	for _, log := range res.EVMReceipt.Logs {
+		if slices.Contains(log.Topics, topic) {
+			err = e.abi.UnpackIntoInterface(v, eventName, log.Data)
+		}
+		require.NoError(e.chain.t, err)
+	}
+	return res
+}
```
jorgemmsilva: the name of this function is misleading, `CallFnExpectMultipleEvents` implies it will assert that multiple events were emitted, which is not the case. IMO changing changing the existing `CallFnExpectEvent` function to have this behavior seems okay.

btw, some issues with the implementation of this function:
 - it doesn't error if the event is not found.
 - since we keep iterating, it will unpack the value of the `last` found event, rather than the `fist`. It should look like this instead:
 ```
 if slices.Contains(log.Topics, topic) {
			err = e.abi.UnpackIntoInterface(v, eventName, log.Data)
		        require.NoError(e.chain.t, err)
		        return ret
		}
 ```

jorgemmsilva: but I don't really think this function is necessary, because the test contract functionality should be split into separate function IMO

packages/evm/evmtest/GetBalance.sol
```diff
@@ -0,0 +1,27 @@
...
+        
+        uint64 baseBalance = ISC.accounts.getL2BalanceBaseTokens(agentID);
+        emit GotBaseBalance(baseBalance);
+
+        NativeTokenID memory id = NativeTokenID({ data: nativeTokenID});
+        uint256 nativeTokens = ISC.accounts.getL2BalanceNativeTokens(id, agentID);
+        emit GotNativeTokenBalance(nativeTokens);
+
+        uint256 nfts = ISC.accounts.getL2NFTAmount(agentID);
+        emit GotNFTIDs(nfts);
```
jorgemmsilva: would make more sense to split the functionality to be tested in different functions, rather than calling `getBalance` multiple times and assert different types of events.

packages/vm/core/evm/evmtest/contractInstance.go
salaheldinsoliman: This does make sense yes, Initially I didn't want to touch the `.sol` file from the Wiki, but splitting it to be one event per function seems better 👍

Dr-Electron:
Just a small comment. Let's move this into its own subfolder. Something like how-tos. This will make it more clear that the team has to be careful when changing this docs as we import them in the wiki. Maybe even add a note about that somewhere in the files? Because we might import subparts of the code and if someone changes a line or two it will break what we show in the wiki

salaheldinsoliman:
> Just a small comment. Let's move this into its own subfolder. Something like how-tos. This will make it more clear that the team has to be careful when changing this docs as we import them in the wiki. Maybe even add a note about that somewhere in the files? Because we might import subparts of the code and if someone changes a line or two it will break what we show in the wiki

@Dr-Electron I initially wanted to implement it in `/wasp/documentation/tutorial-examples/test`, but we would need to modify the CI to build solidity contracts in that dir too. Do you mean a subfolder under the same `evmtest` dir?

Dr-Electron:
> > Just a small comment. Let's move this into its own subfolder. Something like how-tos. This will make it more clear that the team has to be careful when changing this docs as we import them in the wiki. Maybe even add a note about that somewhere in the files? Because we might import subparts of the code and if someone changes a line or two it will break what we show in the wiki
> 
> @Dr-Electron I initially wanted to implement it in `/wasp/documentation/tutorial-examples/test`, but we would need to modify the CI to build solidity contracts in that dir too. Do you mean a subfolder under the same `evmtest` dir?

Yeah lets keep it in evmtest so the CI need no or just minor changes. But have a subfolder there. Maybe some day we can revisit that whole importing logic 👍

jorgemmsilva:
you can absolutely create a new directory like `/wasp/documentation/tutorial-examples/test`. just add another line here:
https://github.com/iotaledger/wasp/blob/develop/Makefile#L39
after that `make compile-solidity` should compile stuff inside the new dir

salaheldinsoliman:
@jorgemmsilva @Dr-Electron   How about a subdir `wiki-how-tos` under `evmtest`, to keep all Solidity there? and also since it is not really a tutorial

jorgemmsilva:
@salaheldinsoliman sounds good to me :+1:

Dr-Electron:
I personally would just call it [how_tos](https://github.com/iotaledger/iota-sdk/tree/develop/sdk/examples/how_tos) so it is the same as in the SDK. But yeah that sounds good. Go for it

jorgemmsilva (approved):

PR #3462, opened 2024-07-08
Feat: issue evm event on L1 nft mint

jorgemmsilva:
In the following block after minting a L1 NFT form the chain (the first block where said NFTID is known), The committee will now produce an EVM event, so that EVM client applications (indexers/explorers/wallets) can account for the new NFT.

misc:
 - fixes an issue where NFT contracts were being created unecessarily
 - fixes the limit of requests that can fit in an ISC block to maxUint16


lmoe (approved):

packages/vm/vmimpl/runtask.go
```diff
@@ -215,6 +240,11 @@ func (vmctx *vmContext) runRequests(
 				allReqs = append(allReqs, retry)
 			}
 		}
+
+		// abort if num of requests is above max_uint16.
+		if reqIndex+1 == math.MaxUint16 {
+			break
```
dessaya, line 247: Maybe add a log or similar? Becasue "errors should never pass silently".

dessaya (approved):

PR #3474, opened 2024-07-17
fix: L1 NFT attributes

jorgemmsilva:
`getTokenURI` currently does not correctly return the attributes from L1 NFTs metadata.

due to an array of mappings to arbitrary types (JSON) being very hard to represent in Solidity, I opted to remove the `attributes` field from the `getIRC27NFTData` function.


packages/isc/irc27nft.go
```diff
@@ -14,10 +14,10 @@ type IRC27NFTMetadata struct {
 	Royalties      map[string]float32 `json:"royalties,omitempty"`
 	IssuerName     string             `json:"issuerName,omitempty"`
 	Description    string             `json:"description,omitempty"`
-	Attributes     string             `json:"attributes,omitempty"`
+	Attributes     []interface{}      `json:"attributes,omitempty"`
```
dessaya, line 17: Nit: `[]any` is shorter :)

Also: `json.RawMessage` does not work for this case?

dessaya (approved):

packages/isc/irc27nft.go
jorgemmsilva: ah, I had auto-merge enabled :/ . Good points tho, thanks :+1: 

`json.RawMessage` unfortunately makes [this test](https://github.com/iotaledger/wasp/blob/develop/packages/isc/irc27nft_test.go#L17) fail because it strips the whitespace:
```
        	            	- Attributes: (json.RawMessage) (len=37) {
        	            	-  00000000  7b 22 74 72 61 69 74 5f  74 79 70 65 22 3a 20 22  |{"trait_type": "|
        	            	-  00000010  46 6f 6f 22 2c 20 22 76  61 6c 75 65 22 3a 20 22  |Foo", "value": "|
        	            	-  00000020  42 61 72 22 7d                                    |Bar"}|
        	            	+ Attributes: (json.RawMessage) (len=34) {
        	            	+  00000000  7b 22 74 72 61 69 74 5f  74 79 70 65 22 3a 22 46  |{"trait_type":"F|
        	            	+  00000010  6f 6f 22 2c 22 76 61 6c  75 65 22 3a 22 42 61 72  |oo","value":"Bar|
        	            	+  00000020  22 7d                                             |"}|
        	            	  
```

PR #3481, opened 2024-07-25
Fix chain not found error

vitaliy-io:
Fix api error message if chainID not found
Fix api error message if network prefix is incorrect


packages/webapi/controllers/chain/chain.go
```diff
@@ -65,7 +65,7 @@ func (c *Controller) getChainInfo(e echo.Context) error {
 	}
 
 	chainInfo, err := c.chainService.GetChainInfoByChainID(chainID, e.QueryParam(params.ParamBlockIndexOrTrieRoot))
-	if errors.Is(err, interfaces.ErrChainNotFound) {
+	if errors.As(err, &interfaces.ChainNotFoundError{}) {
```
nnikolash: @jorgemmsilva please note style of usage of ChainNotFoundError.
`errors.Is` replaced with `errors.As`. There was another option - to add separate wrapper in addition to ErrChainNotFound, which will return it through `Unwrap()`. That would allow to keep `errors.Is` as it is now. But that would also split error into two separate parts, which we thought is not a nice thing to have.

jorgemmsilva (approved):

packages/isc/chainid.go
```diff
@@ -121,6 +127,10 @@ func (id ChainID) ShortString() string {
 
 // String human-readable form (bech32)
 func (id ChainID) String() string {
+	if id.Empty() {
+		return ""
```
dessaya: This probably breaks the invariant:

```
ChainIDFromString(chainID.String()) == chainID
```

Maybe it's better to panic if the chainID is "empty"?

dessaya: Or just return the bech32 as before. Why is the check needed?

nnikolash: Oh, thanks, actually I reverted it. It was accidentally pushed

dessaya (approved):

jorgemmsilva (approved):

PR #3515, opened 2024-11-21
Implement trace_block EP and conversion logic

vitaliy-io:


packages/evm/jsonrpc/trace_block.go
```diff
@@ -0,0 +1,205 @@
...
+type TraceBlock struct {
+	Jsonrpc string   `json:"jsonrpc"`
+	Result  []*Trace `json:"result"`
+	ID      int      `json:"id"`
+}
+
+type Trace struct {
+	Action      interface{}  `json:"action"`
+	BlockHash   *common.Hash `json:"blockHash,omitempty"`
+	BlockNumber uint64       `json:"blockNumber,omitempty"`
```
dessaya, line 18: Maybe this should be `hexutil.Uint64`? Same for `TransactionPosition`.

dessaya (approved):

packages/evm/jsonrpc/trace_block.go
vitaliy-io: It's in decimal format in https://docs.chainstack.com/reference/ethereum-trace_block.
In erigon it's also just uint64

PR #3517, opened 2024-12-05
chore: Fix client output map

lmoe:
# Description of change

Adds a timelock filter to the OutputMap query to ignore timelocks that are still in the future.

## Type of change

### DIFF
--- a/packages/l1connection/l1connection.go
+++ b/packages/l1connection/l1connection.go
@@ -14,6 +14,7 @@ import (
 	iotago "github.com/iotaledger/iota.go/v3"
 	"github.com/iotaledger/iota.go/v3/builder"
 	"github.com/iotaledger/iota.go/v3/nodeclient"
+
 	"github.com/iotaledger/wasp/packages/cryptolib"
 	"github.com/iotaledger/wasp/packages/isc"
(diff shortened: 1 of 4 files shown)


packages/l1connection/l1connection.go
```diff
@@ -92,8 +93,20 @@ func (c *l1client) OutputMap(myAddress iotago.Address, timeout ...time.Duration)
...
 
 	bech32Addr := myAddress.Bech32(parameters.L1().Protocol.Bech32HRP)
+	falseParam := false
+	trueParam := true
+
 	queries := []nodeclient.IndexerQuery{
-		&nodeclient.BasicOutputsQuery{AddressBech32: bech32Addr},
+		&nodeclient.BasicOutputsQuery{
+			AddressBech32: bech32Addr,
+			IndexerTimelockParas: nodeclient.IndexerTimelockParas{
```
alexsporn, line 102: you need to actually do 2 queries, because this would only give you ones with a timelock that passed, not ones without a timelock

alexsporn: So you need to add a second one where `HasTimelock:      &falseParam`

alexsporn (approved):
lgtm!

PR #3521, opened 2025-01-14
Task/improve evm tracer performance

vitaliy-io:
Fix N^2 behavior of block tracers

### DIFF
--- a/packages/evm/jsonrpc/evmchain.go
+++ b/packages/evm/jsonrpc/evmchain.go
@@ -699,7 +699,7 @@ func (e *EVMChain) traceTransaction(
 		BlockNumber: new(big.Int).SetUint64(blockNumber),
 		TxIndex:     int(txIndex),
 		TxHash:      tx.Hash(),
-	}, config.TracerConfig, false)
+	}, config.TracerConfig, false, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -736,17 +736,21 @@ func (e *EVMChain) debugTraceBlock(config *tracers.TraceConfig, block *types.Blo
 
 	blockNumber := uint64(iscBlock.BlockIndex())
 
+	blockTxs := block.Transactions()
+	fakeTxs := make([]*types.Transaction, 0, len(blockTxs))
+	for _, tx := range blockTxs {
+		if e.isFakeTransaction(tx) {
+			fakeTxs = append(fakeTxs, tx)
+		}
+	}
+
 	tracer, err := newTracer(tracerType, &tracers.Context{
 		BlockHash:   block.Hash(),
 		BlockNumber: new(big.Int).SetUint64(blockNumber),
-	}, config.TracerConfig, true)
+	}, config.TracerConfig, true, types.Transactions(fakeTxs))
 	if err != nil {
 		return nil, err
 	}
-
-	// if e.isFakeTransaction(tx) {
-	// 	return tracer.TraceFakeTx(tx)
-	// }
 
 	err = e.backend.EVMTrace(
 		iscBlock.PreviousAliasOutput,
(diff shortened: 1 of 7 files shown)


packages/evm/jsonrpc/tracer_call.go
```diff
@@ -119,12 +120,15 @@ type TxTraceResult struct {
 }
 
 type callTracer struct {
-	callstack []CallFrame
-	config    callTracerConfig
-	gasLimit  uint64
-	depth     int
-	interrupt atomic.Bool // Atomic flag to signal execution interruption
-	reason    error       // Textual reason for the interruption
+	txToStack  map[common.Hash][]CallFrame
```
dessaya, line 123: Isn't it possible to avoid changing this struct and also `prestateTracer`?
I'm asking because these files are almost a 1 to 1 copy of the code in go-ethereum, and so they are very easy to update whenever there are changes upstream. These changes will make it more difficult to keep the code updated.

Also, I see that there is some duplicated logic in the call and prestate tracers, surrounding the `map[common.Hash]`, `currentTxHash`, `traceBlock` and  `fakeTxs`, so maybe there is a different solution? How does go-ethereum solve this problem?

vitaliy-io: Geth does not have N^2 problem because their VM layer is different, the bottom line is that geth's VM allows transactions to be executed individually (meaning that each tx has it's own tracer) but reusing the same state (thus allowing txs to see the changes made by previous txs in the same block). Basically they loop through the block transactions and for each one they run VM with tracing. 
Out VM however accepts a list of transactions to run and only then it gets the state to base on. I'm sure that we could change our VM to work like geth, but for me it's hard to understand how difficult that would be (maybe it's easy).
That's why I decided to alter the tracers itself. But I agree that it would make it difficult to copy fixes from geth

dessaya: Ok, I see. Anyway, do you think it's possible to fix or at least decrease the amount of duplicated code?

Maybe moving that logic to our own `Tracer` struct:

```
type Tracer struct {
    txTracers  map[common.Hash]*tracers.Tracer
    currentTx  common.Hash
	traceBlock bool
	fakeTxs    []*types.Transaction
}
```

packages/evm/jsonrpc/evmchain.go
```diff
@@ -729,32 +717,42 @@ func (e *EVMChain) debugTraceBlock(config *tracers.TraceConfig, block *types.Blo
...
-		if err == nil {
-			results = append(results, TxTraceResult{
-				TxHash: tx.Hash(),
-				Result: result,
-			})
+	blockTxs := block.Transactions()
+	fakeTxs := make([]*types.Transaction, 0, len(blockTxs))
+	for _, tx := range blockTxs {
+		if evmutil.IsFakeTransaction(tx) {
+			fakeTxs = append(fakeTxs, tx)
```
dessaya: So `fakeTxs` contains the list of fake txs in the block. But it doesn't record their tx index. I don't remember, is it possible that a block contains regular and fake txs interleaved? Shouldn't the trace ouput respect the transaction indices?

```
block:
  tx #0: regular tx
  tx #1: fake tx
  tx #2: regular tx
  tx #3: fake tx
```

packages/evm/jsonrpc/tracer_call.go
vitaliy-io: will try to fix that

packages/evm/jsonrpc/evmchain.go
vitaliy-io: I think it's possible for a block to contain both, why not. I could not test it via solo though 
I thought the order was irrelevant because it's a map, but probably you are right, chatGpt says that order should be respected.
Will fix that

packages/evm/jsonrpc/tracer_call.go
vitaliy-io: Refactored GetResult method of tracers and extracted similar logic
Could not reuse `Tracer struct ` for common fields because they would not be accessible from tracers, not sure how exactly you proposed to extract them

packages/evm/jsonrpc/evmchain.go
vitaliy-io: fixed

dessaya (approved):

dessaya:
Pls apply this change in isc-private once it's merged :pray:

PR #3522, opened 2025-01-15
Fix wasp-cli request sending

lmoe:
# Description of change

Updates the CLI to ignore timelocked outputs only when sending (requests)

### DIFF
--- a/clients/chainclient/chainclient.go
+++ b/clients/chainclient/chainclient.go
@@ -5,6 +5,7 @@ import (
 	"math"
 
 	iotago "github.com/iotaledger/iota.go/v3"
+
 	"github.com/iotaledger/wasp/clients/apiclient"
 	"github.com/iotaledger/wasp/clients/apiextensions"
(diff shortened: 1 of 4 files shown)


clients/chainclient/chainclient.go
```diff
@@ -47,6 +48,7 @@ type PostRequestParams struct {
 	Allowance                *isc.Assets
 	gasBudget                uint64
 	AutoAdjustStorageDeposit bool
+	OnlyUnlockedOutputs      bool
```
dessaya, line 51: Honestly if you are sending a request it doesn't make much sense to use locked outputs anyway... is the flag really needed?

dessaya (approved):

clients/chainclient/chainclient.go
lmoe: Yes because its needed by PostRequest functions. The OutputMap function is so interwoven that we need a flag before that. Its also used to pull balances and other information, where timelocked objects might be relevant.

PR #3523, opened 2025-01-21
refactor(evm): simplify tracer code

dessaya:


packages/vm/core/evm/evmimpl/impl.go
```diff
@@ -510,9 +511,33 @@ func AddDummyTxWithTransferEvents(
 	}
 	receipt.Bloom = types.CreateBloom(types.Receipts{receipt})
 
-	if !reuseCurrentTxContext {
+	callTracerHooks := func() {
```
vitaliy-io, line 514: are you sure that all cases covered by lagacy IsFakeTransaction() method will be covered by this code in AddDummyTxWithTransferEvents().
In other words do all fake transactions go through AddDummyTxWithTransferEvents function?

packages/vm/core/evm/evmimpl/internal.go
```diff
@@ -36,18 +36,7 @@ func getTracer(ctx isc.Sandbox) *tracing.Hooks {
...
-		return tracer.Tracer.Hooks
-	}
-
-	// if tx index is set, we're tracing a specific transaction
-	if tracer.TxIndex != nil && *tracer.TxIndex == uint64(ctx.RequestIndex()) {
-		return tracer.Tracer.Hooks
-	}
-
-	return nil
+	return tracer.Hooks
```
vitaliy-io, line 39: So we are going to execute tracing for all transactions in a block even if we only need to trace a single transaction
Is it ok performance-wise?

packages/vm/core/evm/evmimpl/impl.go
dessaya: Yes, I'm fairly sure. I searched for all usages of BlockchainDB.AddTransaction and this was the only one.

Also I think that if I missed some case, the tests would not be passing.

packages/vm/core/evm/evmimpl/internal.go
dessaya: Yes, exactly. It simplifies a lot of code, and since most blocks have only a few transactions I think it should be ok. The only performance problem we had was with that block with hundreds of txs, and they were tracing the whole block anyway.

vitaliy-io (approved):

